PR #41, opened 2013-01-17
Add TeVCat

cdeil:
TeVCat ( http://tevcat.uchicago.edu/ ) is an online catalog of TeV gamma-ray sources that is continuously updated and I would like to add it to astroquery.

I already have code to make astropy tables from their JSON data.
At the moment I do `View -> Developer -> JavaScript Console` in the Chrome browser and type `jsonData`.
The one thing I don't know how to do is how to get the data from Python.
@keflavich Do you know how to do that? Wout `TeVCat` be acceptable for `astroquery` or do you think it's too niche?


keflavich:
@cdeil this would be a *great* sprint topic

cdeil:
I'm closing this PR.
No TeVCat for Astroquery.

The main reason is that I've been in email contact with the TeVCat people and I know they don't want people to just grab TeVCat like this. This is also the reason they don't offer it for download as a little FITS file or whatever. They want people to access the data via the TeVCat webpage.

Thank you @keflavich for all the time you've spent on this PR to give me feedback!

cdeil:
Just in case someone finds this old closed issue.

* The TeVCat terms of use state that bulk download is forbidden (they want people to go to their website or contact them privately)
* We've made a better open TeV source catalog: https://github.com/gammapy/gamma-cat
* There is support for gamma-cat in `gammapy.catalog` (using Gammapy classes to represent spectra, lightcurves, ...), so no need to add to astroquery.

PR #458, opened 2014-11-02
CosmoSim remote tests

agroener:
Creating CosmoSim remote rests. First commit also switches out old <code>logging.warning</code> or <code>print</code> statements for <code>LoginError</code> .


keflavich:
@agroener sorry this has been sitting idle for so long.  I'll merge it if it works.  I got a failure, though:
http://pastebin.com/R66D1E66

agroener:
Yea, the tests I was trying to come up with kept failing, I believe because I'm still unable to re-run queries which have been run in the past due to caching. If I've run a query, it's cached by default (which is awesome). If I then delete it from the server and try to run it again, I'm unable to because it thinks it has already been run. So now that particular query has been 'used up' so to speak. I think I've submitted an issue on this, but there was no conclusion as to what I should do.

keflavich:
Right... hrm.  I think in an older version of the code did not pass the `cache=False` option through to the `_request`, but as far as I can tell your code now does.  Is it that `cache=False` just isn't working?  I couldn't find your old issue; maybe it was this one (https://github.com/astropy/astroquery/pull/418) which has now been merged?

agroener:
Let me snoop around a bit and see where exactly the issues are. It's been a little while since I've worked on this.

agroener:
@keflavich 

Okay, so the issue I filled out on this topic is #455 . Essentially, to reproduce the error I'm talking about, you would have to:

1) Run the query successfully with cache=True
2) Use the delete function to delete the query from the server (this is done using functionality in the cosmosim code I've written; users can manage their queries on the server itself)
3) Since it was the case that cache=True, trying to re-run the same sql query on the server will not work, because it still exists locally.

It is a problem of it existing locally, but not existing on the server. Does this make more sense?

These are things I could look into:
1) In this situation, I could check to see if it exists in cache and doesn't exist on the server. I could try changing it such that cache=False in this case, and perhaps that will work.
2) We could 'open up cache', delete that query, and close cache (I'm imagining it as a pickle) - but this would require offering some functionality for doing this, and I don't know how caching is done to be honest.
3) The cache could be wiped, and rebuilt from existing queries on the server, and this would clean out old queries which exist in cache, but no longer exist on the server. Again - this would require knowing how caching works.

Of these options, I think the second would be the better, because this would ensure that queries do not build up locally, even though they're being managed remotely. Let me know what you think about these ideas.

keflavich:
I think we want to have the `delete` function remove the locally cached version in addition to the remote version.  That seems to be what your (2) is doing?

The caching system creates a hash (a hopefully-unique alphanumeric string) given some set of input parameters (html POST data).  This hash can easily be regenerated from any set of query data.  So you just need to find the appropriate cache file and delete it.  Look at the [`hash` function in `query.py`](https://github.com/astropy/astroquery/blob/master/astroquery/query.py#L59) to see how the hash (and filename) are generated.

agroener:
Okay - awesome, that's exactly what I was looking for all along. I'll look into it and see how I can delete a specific cached query using the hash.

agroener:
Question: 
If I were able to re-create the exact hash (assuming the hash would be unique to the content of the query; similar to how Git uses SHA1 hashes), are you proposing that I create a delete function, or are you saying that there already exists a way to delete it from the cache? 

I guess I'm a bit confused as to how I am able to call the hash function, since the CosmoSim class does not inherit this method.

keflavich:
You would need to create a delete method.

Cosmosim doesn't inherit the `hash` method, but it inherits the [`_request`](https://github.com/astropy/astroquery/blob/master/astroquery/query.py#L116) method, which creates an [`AstroQuery`](https://github.com/astropy/astroquery/blob/master/astroquery/query.py#L29) object, which has the [`hash`](https://github.com/astropy/astroquery/blob/master/astroquery/query.py#L59) method.  So you could either create the `AstroQuery` on your own, or you could have `_request` create it and then grab the hash.  Maybe it makes sense to split out the `AstroQuery` generation part of the `_request` method, perhaps into a `_new_query` method?

keflavich:
This PR has lingered for long enough.  Closing.  Please reopen if it's finished.

PR #470, opened 2015-01-12
Docs are not building with "cannot import sphinx extension" errors

keflavich:
The RTD docs builds are failing:
https://readthedocs.org/builds/astroquery/2084778/

```
Extension error:
Could not import extension astropy.sphinx.ext.edit_on_github (exception: No module named sphinx.ext.edit_on_github)
```

@astrofrog - if you have a quick solution or can point me to someone who does, that would be helpful.  If this isn't a solved problem I'll just leave it open for now.

PR #482, opened 2015-02-05
Package template merge

keflavich:
I squashed all of package-template into one commit and merged it.  There's one open question, though.


setup.py
```diff
@@ -78,6 +78,14 @@
 package_info['package_data'].setdefault(PACKAGENAME, [])
 package_info['package_data'][PACKAGENAME].append('data/*')
 
+# Adam's note: I don't know what to do with this.
+# # Define entry points for command-line scripts
+# entry_points = {}
+# entry_points['console_scripts'] = [
+#     'astropy-package-template-example = packagename.example_mod:main',
+# ]
```
keflavich: @astrofrog: can you tell me what ought to be done with this?  Should `packagename` be changed to `astroquery` here?  Should `example_mod` also be astroquery?  I'm unfamiliar with this.  Maybe this is only for command line scripts, which astroquery currently does not include?

keflavich: @cdeil since you were helping with package-template-merging stuff, do you have any idea what I should be doing with this segment?

keflavich:
This probably replaces #470

setup.py
astrofrog: You can just remove all that because there are no scripts in astroquery.

keflavich: OK, thanks - I just wanted to confirm that this was indeed for scripts.  I'll change the comment.

PR #485, opened 2015-02-05
Don't run 2to3

cdeil:
@keflavich - A small fix after #482 ... I think in the merge you changed to `use_2to3=True` in `setup.py` by accident ... this changes it back.
(and removes two files from the package-template that are irrelevant for astroquery)


keflavich:
Why should we not use 2to3?  I changed it because it was changed in the template, but I have no real clue either way.  For the other changes, I agree and will merge.

cdeil:
> Why should we not use 2to3?

Because `2to3` changes the code, so line numbers and code in tracebacks from Python 2 / 3 users e.g. in error reports are different. Even more so, it's inconvenient for `astroquery` developers that use Python 3 (and shouldn't we all these days ... :-)) because developing in-place with `pip install --user --editable .` or the older invocation `python setup.py develop` will not work properly (I think).

In the case of `astroquery` the codebase already simultaneously works with Python 2 and 3 since #320 and turning `2to3` off makes sure it stays that way (at least for code that is covered by tests).

keflavich:
Sold!

Issue #494, opened 2015-03-01
Remote Failures, March 2015


keflavich:
https://gist.github.com/keflavich/d0bca4e16f4eefcdd357

The failures are in ALMA and NRAO.  The ALMA failure may be related to a server change.

keflavich:
The ALMA ones have been fixed.  The NRAO errors stopped, so that was a one-of error.

PR #495, opened 2015-03-01
Fix ALMA page parsing for staging requests

keflavich:
ALMA staging requests have been failing recently because of a change in the data the server returns.  The change can be observed in these two URLs:

https://almascience.eso.org/rh/requests/anonymous/519752156
https://almascience.eso.org/rh/requests/anonymous/786572566

These are identical requests (for the same data), but the former includes size information and links.

The "fix" is to download the downloadRequest shell script and parse the download URL from that script.  This is not really any more or less of a hack, but it has the unfortunate failing of not including the file size data.

Also, this addresses part of #494


keflavich:
Remote tests pass locally.  ALMA is back in action!

PR #501, opened 2015-03-25
Update to v0.4 / v1.0 astropy configuration system

keflavich:
@mdboom migrated this from #498 since there's some hope that that PR and other old ones can be merged against astropy with reversion.

This is still a WIP as there are some unclear conflicts remaining.

PR #502, opened 2015-03-26
Include config file in source repo

mdboom:
It turns out most of #501 is unnecessary -- astroquery was already on the "new" config API and basically all we were seeing was the result of a recent breakage in astropy (that only affected affiliated packages).

This is basically #501 boiled down to only the necessary bit, which is to include a copy of the config file in the source tree since it is no longer generated in the new config scheme.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,6 @@ __pycache__
 # Other generated files
 */version.py
 */cython_version.py
-astroquery/astroquery.cfg
 
 htmlcov
 .coverage
--- a/astroquery/astroquery.cfg
+++ b/astroquery/astroquery.cfg
@@ -1,0 +1,197 @@
+[besancon]
+
+# Besancon download URL.  Changed to modele2003 in 2013.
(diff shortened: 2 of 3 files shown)


keflavich:
Right, thanks!  Will this still fail until astropy/3627 is merged?

mdboom:
Yes.  It requires astropy/astropy#3627 to work.  I hope to merge that as soon as Travis passes over there.

keflavich:
Got it.  Ping me when that's done and I'll restart the tests here (I don't think you have the rights to do that on astroquery?  maybe you do?).  Not sure if you saw my previous question because I edited it - will we have to do something about testing against astropy-1.0, since that won't have astropy/#3627 included?

mdboom:
astropy 1.0 is fine.  The bug is only on astropy 1.0.1.  If we do release a 1.0.2, it should be fixed there.  What I plan to look into next is whether there's a workaround in order to support astropy 1.0.1.  Not sure what, short of monkey-patching.

mdboom:
The astropy side is merged.

keflavich:
@mdboom could you rebase or merge master?  I merged #503; I'd like travis to go green before merging this guy even though I'm sure it is fine.

PR #503, opened 2015-03-26
Monkey patch astropy so that configuration system works

mdboom:
Ugh.  Embarrassing.

@embray: I guess I should put the equivalent in the package template?


embray:
I don't know.  I'd just as soon release v1.0.2 with the fix and say forget it :)

mdboom:
As it turns out, the bug only affects astroquery, so the urgency on 1.0.2 than it otherwise might be.

Issue #477, opened 2015-01-28
Vizier.find_catalogs str TypeError on Python 3


cdeil:
I think `Vizier.find_catalogs` is broken on Python 3 ... e.g. for the example from the [astroquery.vizier docs page](https://astroquery.readthedocs.org/en/latest/vizier/vizier.html):
```
>>> from astroquery.vizier import Vizier
>>> Vizier.find_catalogs('Kang W51')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/deil/Library/Python/3.4/lib/python/site-packages/astroquery-0.3.dev2704-py3.4.egg/astroquery/vizier/core.py", line 174, in find_catalogs
    if 'STOP, Max. number of RESOURCE reached' in response.content:
TypeError: Type str doesn't support the buffer API
```

It works on Python 2 with a warning:
```
>>> Vizier.find_catalogs('Kang W51')
/Users/deil/Library/Python/2.7/lib/python/site-packages/astroquery-0.3.dev2704-py2.7.egg/astroquery/vizier/core.py:564: UserWarning: VOTABLE parsing raised exception: 
  warnings.warn("VOTABLE parsing raised exception: {0}".format(ex))
{u'J/ApJ/706/83': <astropy.io.votable.tree.Resource object at 0x10721cf50>, u'J/ApJS/191/232': <astropy.io.votable.tree.Resource object at 0x108917610>}
```

Looks like there's no way to make this work on Python 3?
```
>>> Vizier.find_catalogs(b'Kang W51')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/deil/Library/Python/3.4/lib/python/site-packages/astroquery-0.3.dev2704-py3.4.egg/astroquery/vizier/core.py", line 174, in find_catalogs
    if 'STOP, Max. number of RESOURCE reached' in response.content:
TypeError: Type str doesn't support the buffer API
>>> Vizier.find_catalogs(u'Kang W51')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/deil/Library/Python/3.4/lib/python/site-packages/astroquery-0.3.dev2704-py3.4.egg/astroquery/vizier/core.py", line 174, in find_catalogs
    if 'STOP, Max. number of RESOURCE reached' in response.content:
TypeError: Type str doesn't support the buffer API
```

PR #509, opened 2015-04-02
Python 3 fix in Vizier.find_catalogs

barentsen:
`Vizier.find_catalogs` is currently broken in Python 3.

This PR makes the necessary compatibility fixes. It enables `test_vizier_remote.py` to pass in both Python 2.7 and 3.4 on my machine (whereas it is failing in 3.4 at present).

If this gets merged, #477 can be closed.

### DIFF
--- a/astroquery/vizier/core.py
+++ b/astroquery/vizier/core.py
@@ -171,15 +171,15 @@ class VizierClass(BaseQuery):
                                  url=self._server_to_url(),
                                  data=data_payload,
                                  timeout=self.TIMEOUT)
-        if 'STOP, Max. number of RESOURCE reached' in response.content:
+        if 'STOP, Max. number of RESOURCE reached' in response.text:
             raise ValueError("Maximum number of catalogs exceeded.  Try setting max_catalogs "
                              "to a large number and try again")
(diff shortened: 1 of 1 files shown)


astroquery/vizier/core.py
```diff
@@ -171,15 +171,15 @@ def find_catalogs(self, keywords, include_obsolete=False, verbose=False,
...
+        if 'STOP, Max. number of RESOURCE reached' in response.text:
             raise ValueError("Maximum number of catalogs exceeded.  Try setting max_catalogs "
                              "to a large number and try again")
         result = self._parse_result(response, verbose=verbose, get_catalog_names=True)
 
         # Filter out the obsolete catalogs, unless requested
         if include_obsolete is False:
-            for (key, resource) in result.items():
-                for info in resource.infos:
+            for key in list(result):
```
barentsen: Note: the call to `list()` is necessary to force Py 3 to return a list rather than an iterator. The use of an iterator will trigger a `RuntimeError` when `del result[key]` is called below.

keflavich:
Looks good, and it passes locally.  Will merge when travis says so.

PR #515, opened 2015-04-14
ALMA: Add one more doc link and run remote tests against beta server

keflavich:


keflavich:
The coverage check is not relevant; we can't run remote tests on CI services anyway.

PR #521, opened 2015-04-24
Add Astrometry.net support

keflavich:
Working on this with Victor Terron


astroquery/astrometry_net/core.py
```diff
@@ -0,0 +1,475 @@
...
+# export all the public classes and methods
+__all__ = ['Astrometry', 'AstrometryClass']
+
+# declare global variables and constants if any
+
+# Now begin your main class
+# should be decorated with the async_to_sync imported previously
+
+@async_to_sync
+class AstrometryClass(BaseQuery):
```
keflavich: Could we call this `AstrometryNetClass` instead?

astroquery/astrometry_net/core.py
```diff
@@ -0,0 +1,475 @@
...
+    def _store_API_key(self, value):
+        """ Cache the Astrometry.net API key on disk. """
+        pass
+
+    def _get_stored_API_key(self):
+        """ Return the API key, raise KeyError if not cached on disk. """
+        raise KeyError
+
+    @property
+    def key(self):
```
keflavich: probably better to call this `api_key` to avoid confusion with dict keys

astroquery/astrometry_net/core.py
vterron: Sure, I'll rename it.

astroquery/astrometry_net/core.py
vterron: +1

vterron:
I have talked to the Astropy developers and it seems that automatically writing the API in the configuration file would not be a good idea. There are several reasons for this — the most convincing of them probably being that this could be very confusing for users, who may found that another library has modified their configuration file by calling our code. 

Would using `getpass` be an option? The Astronometry.net API key does not have an associated username, so maybe we could use something like `astrometry-net-user` as a placeholder.

fred3m:
Ok, I think this should be good to go. Try testing it and let me know if it works. The only thing that will need to be changed is that they use a print function in a way I've never seen to build their data packet, and it won't be compatible with Python 3 (I had to comment out the __future__ import).

astroquery/astrometry_net/core.py
```diff
@@ -0,0 +1,618 @@
...
+            def _write_headers(self, msg):
+                # We don't want to write the top-level headers;
+                # they go into Request(headers) instead.
+                if self.root:
+                    return                        
+                # We need to use \r\n line-terminator, but Generator
+                # doesn't provide the flexibility to override, so we
+                # have to copy-n-paste-n-modify.
+                for h, v in msg.items():
+                    print >> self._fp, ('%s: %s\r\n' % (h,v)),
```
keflavich: I think you can use `self._fp.write` instead of `print`

sargas:
Does this code still work? Is there any remaining issues that prevent it from being merged?

keflavich:
Actually, it probably can be merged... it would be nice to have some remote tests in place so we can be sure, but really I just lost track of it.

astroquery/astrometry_net/__init__.py
```diff
@@ -0,0 +1,32 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+"""
+<Put Your Tool Name Here>
+-------------------------
+
+:author: <your name> (<your email>)
```
keflavich: this header needs to be filled out

astroquery/astrometry_net/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+# for docs and examples on how to do this
+# Below is a common use case
+
+from astropy.config import ConfigurationItem
+
+# Set the server mirrors to query
+SERVER = ConfigurationItem('server',
+                           ['http://dummy_server_mirror_1',
+                            'http://dummy_server_mirror_2',
+                            'http://dummy_server_mirror_n'],
```
keflavich: we need real URLs here, or these need to be removed

astroquery/astrometry_net/core.py
```diff
@@ -0,0 +1,618 @@
...
+        'scale_type': 'ul', 
+        'scale_lower': 0.20,
+        'scale_upper': 0.30,
+        'center_ra': 8.711675,
+        'center_dec': -78.9810555556,
+        'radius': 15,
+        'parity': 0
+    }
+
+    AstrometryNet.solve(sources, settings, 'X_IMAGE', 'Y_IMAGE', 'FWHM_IMAGE', 'FLAGS', 'FLUX_APER')
```
keflavich: Can this be turned into a RemoteTest?

astroquery/astrometry_net/core.py
```diff
@@ -0,0 +1,618 @@
...
+
+
+Settings
+========
+In order to speed up the astrometric solution it is possible to pass a dictionary of settings
+to `astroquery.AstrometryNet.build_pacakge` or `astroquery.AstrometryNet.build_pacakge`.
```
keflavich: package

astroquery/astrometry_net/core.py
```diff
@@ -0,0 +1,618 @@
...
+# should be decorated with the async_to_sync imported previously
+
+astrometry_net_url = 'http://nova.astrometry.net/'
+apiurl = 'http://nova.astrometry.net/api/'
+
+@async_to_sync
+class AstrometryNetClass(BaseQuery):
+
+    """
+    Not all the methods below are necessary but these cover most of the common cases, new methods may be added if necessary, follow the guidelines at <http://astroquery.readthedocs.org/en/latest/api.html>
```
keflavich: wrap this to 80 chars please

astroquery/astrometry_net/core.py
```diff
@@ -0,0 +1,618 @@
...
+    """
+    # use the Configuration Items imported from __init__.py to set the URL, TIMEOUT, etc.
+    URL = SERVER()
+    TIMEOUT = TIMEOUT()
+
+    def _store_API_key(self, value):
+        """ Cache the Astrometry.net API key on disk. """
+        pass
```
keflavich: I guess this still needs to be implemented?

astroquery/astrometry_net/core.py
```diff
@@ -0,0 +1,618 @@
...
+        self._key = value
+
+    def __init__(self):
+        """ Load the cached API key, show a warning message if can't be found. """
+
+        try:
+            self._key = self._get_stored_API_key()
+        except KeyError:
+            log.warn("Astrometry.net API key not found")
+            log.warn("You need to register it with Astrometry.key = 'XXXXXXXX'")
```
keflavich: will astrometry.net work without an API key?  If not, then this should be an Exception, not a warning.

astroquery/astrometry_net/tests/data/dummy.dat
```diff
@@ -0,0 +1,2 @@
+this is a dummy data file.
+Similarly include xml, html, fits and other data files for tests
```
keflavich: Either remove or replace this file

astroquery/astrometry_net/tests/setup_package.py
```diff
@@ -0,0 +1,10 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+import os
+# setup paths to the test data
+# can specify a single file or a list of files
+def get_package_data():
+    paths = [os.path.join('data', '*.dat'),
+             os.path.join('data', '*.xml')] # etc, add other extensions
+    # you can also enlist files individually by names
+    # finally construct and return a dict for the sub module
+    return {'astroquery.template_module.tests': paths}
```
keflavich: this file needs to be updated; template_module should be replaced with astrometry_net

astroquery/astrometry_net/tests/test_module.py
```diff
@@ -0,0 +1,69 @@
...
+from astropy.table import Table
+import astropy.coordinates as coord
+import astropy.units as u
+
+from ...utils.testing_tools import MockResponse
+
+# finally import the module which is to be tested
+# and the various configuration items created
+from ... import template_module
+from ...template_module import (SERVER, TIMEOUT)
```
keflavich: template_module -> astrometry_net

astroquery/astrometry_net/tests/test_module.py
```diff
@@ -0,0 +1,69 @@
...
+# reading in data from some data file:
+def get_mockreturn(url, params=None, timeout=10, **kwargs):
+    filename = data_path('dummy.dat')
+    content = open(filename, "r").read()
+    return MockResponse(content, **kwargs)
+
+# finally test the methods using the mock HTTP response
+def test_query_object(patch_get):
+    result = template_module.core.DummyClass.query_object('m1')
+    assert isinstance(result, Table)
```
keflavich: need a real test here

astroquery/astrometry_net/tests/test_module_remote.py
```diff
@@ -0,0 +1,14 @@
...
+# the actual HTTP request rather than monkeypatching them.
+# should be disabled or enabled at will - use the
+# remote_data decorator from astropy:
+
+from astropy.tests.helper import remote_data, pytest
+
+@remote_data
+class TestDummyClass:
+    # now write tests for each method here
+    pass
```
keflavich: need a real test here too

astroquery/astrometry_net/core.py
keflavich: ...though we want a workaround in place so that offline tests can be run

fred3m:
Sorry, I let this slip through the cracks too. I haven't taken a look at this in a while so it may take me a few days to make the changes.

keflavich:
@fred3m @crawfordsm Just a ping reminding you that this is open and nearly done

fred3m:
Thanks for the reminder. I've been neck deep in job applications and my thesis lately but one weekend when I need a break I'll try and pound this out. Sorry that this is taking so long...

crawfordsm:
Thanks Fred and good luck with everything!

keflavich:
superceded by #1163

PR #537, opened 2015-04-25
Adding LCOGT archive support

zemogle:
This request only has basic image meta data and photometry querying. You can search by name and region. This is for the LCOGT archive which is hosted by IPAC and updated daily.


keflavich:
@zemogle I've made a PR onto this.  I'm happy to merge once you merge that.

keflavich:
Merging, but we need to add docs: `lcogt/lcogt.rst` and an entry in `index.rst`!

zemogle:
Thanks. Its always the docs that get forgotten. I'll get on to that this weekend.

PR #546, opened 2015-05-27
LAMDA: Add a utility for computing the critical density

keflavich:
The critical density for an n>2 level system requires full knowledge of the collision rate network.  It's not trivial to compute, so I've added a utility for it.  Also, a minor bugfix to raise a useful error when LAMDA fails.

The critical density tool needs tests and docs prior to merging.

### DIFF
--- a/astroquery/lamda/__init__.py
+++ b/astroquery/lamda/__init__.py
@@ -16,6 +16,3 @@ Note:
   data are encouraged.
 """
 from .core import Lamda
-
-import warnings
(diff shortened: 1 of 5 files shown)


keflavich:
A remaining 'TODO' item is to allow for a TIMEOUT

keflavich:
While this doesn't include much testing right now, there is a nice complete end-to-end test here:
https://gist.github.com/fdfd584195cd70b91ab0
that shows that everything works.  Another major TODO is integrating this test and testing for appropriate values from ncrit.

PR #554, opened 2015-06-17
Why payload.update always called 

keflavich:
In core.py query_async always calls payload.update. For example: this resets my request for result_view = project to result_view = raw. Why is this, or am I using the wrong method to get a list of projects instead of raw data? Thank you!


keflavich:
Ah, the mistake was mine, I was on an outdated version of astroquery.  It turns out this is easy, allowing `project` just works.

PR #557, opened 2015-06-26
Bugfix - payload keyword 'version' should be 'data_version'. 

vilhelmp:
The payload keyword ['version'] should be ['data_version'], don't know if Splatalogue changed this recently or if we had it wrong previously.
I also added the 'vall' option to the version selection (i.e. both v1.0 and v2.0).

To reproduce this in the unfixed version: 
```python
from astroquery import splatalogue
import astropy.units as u
results = splatalogue.Splatalogue.query_lines(
    min_frequency= 703*u.GHz,
    max_frequency=706*u.GHz,
    chemical_name='Acetaldehyde',
    version='v1.0'
    )
len(results)
```
This is equivalent to this (except the filter for vt=0) URL:
http://www.cv.nrao.edu/php/splat/c.php?submit=submit&sid[]=393&chemical_name=Acetaldehyde%20(CH3CHO%20vt=0)&el4=el4&ls1=ls1&displayRecomb=displayRecomb&displayLovas=displayLovas&displaySLAIM=displaySLAIM&displayJPL=displayJPL&displayCDMS=displayCDMS&displayToyaMA=displayToyaMA&displayOSU=displayOSU&displayLisa=displayLisa&displayRFI=displayRFI&data_version=v1.0&no_atmospheric=no_atmospheric&from=703&to=706&frequency_units=GHz&show_unres_qn=show_unres_qn&show_upper_degeneracy=show_upper_degeneracy&submit=1

with the fix the length is just 1, without it, it is 144.

### DIFF
--- a/astroquery/splatalogue/core.py
+++ b/astroquery/splatalogue/core.py
@@ -27,7 +27,7 @@ class SplatalogueClass(BaseQuery):
     QUERY_URL = conf.query_url
     TIMEOUT = conf.timeout
     LINES_LIMIT = conf.lines_limit
-    versions = ('v1.0', 'v2.0')
+    versions = ('v1.0', 'v2.0', 'vall')
     # global constant, not user-configurable
     ALL_LINE_LISTS = ('Lovas', 'SLAIM', 'JPL', 'CDMS', 'ToyoMA', 'OSU',
                       'Recomb', 'Lisa', 'RFI')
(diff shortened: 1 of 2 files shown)


keflavich:
@vilhelmp good catch, I don't know whether it was an error or if splatalogue has changed their API (they do make frequent changes).  Could you add a regression test for this?  Only a remote test is needed

PR #563, opened 2015-07-31
Splatalogue minor fix

keflavich:
Strip out carriage returns from text b/c they are not intentional and break the parser


keflavich:
This was fixed upstream, so for now I'll say it's not needed.

PR #589, opened 2015-10-08
pass data by file

Jerry-Ma:
I don't know whether this is universal or not. When I tries to do cross_id for a larger list of coordinates (~400, which is not that large really), it fails and gives such message:

```
Your_browser_sent_an_invalid_request
------------------------------------
                      </body></html>
```

Without digging into the details, I suspect this is due to some limitation in requests library (?)

Instead of passing the data via `paste` field, now the data is passed as a file. At least this now works for me.

### DIFF
--- a/astroquery/sdss/core.py
+++ b/astroquery/sdss/core.py
@@ -149,7 +149,8 @@ class SDSSClass(BaseQuery):
                              for i in range(len(coordinates))])
 
         # firstcol is hardwired, as obj_names is always passed
-        request_payload = dict(uquery=sql_query, paste=data,
+        files = {'upload': ('astroquery', data)}
+        request_payload = dict(uquery=sql_query,
                                firstcol=1,
                                format='csv', photoScope='nearPrim',
                                radius=radius,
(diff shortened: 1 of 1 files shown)


keflavich:
I wonder if this is a limitation of POST requests in general?

It looks like `data` is a string in what you passed, but I think it needs to be a file object: http://docs.python-requests.org/en/latest/user/quickstart/#post-a-multipart-encoded-file

bsipocz:
I can check this over the weekend, but the function definitely worked for
me for thousands of objects, with the old infrastructure. Something might
have broken since.

bsipocz:
This works beautifully, and I expect the same logic may even fix #2211. I will open a follow-up PR instead using the commit from here rebased on the top of `main` as it seems that this PR proceeds the times when github allowed maintainers to push to PR branches.

bsipocz:
Superseeded by #2305 and therefore closing. Thanks @Jerry-Ma for the fix, and I'm sorry it took this long to get this finish off and close to merge.

Jerry-Ma:
@bsipocz I am really glad that this works out after these years. Time really flies by. Cheers!

bsipocz:
Yes, the sdss archive is amazingly stable with a reliable API, outstanding amongst all.

PR #596, opened 2015-10-21
Correct irsa_dust link in the README

evertrol:
Minor fix, but since the links lives in the README on the frontpage, it's probably nicer to have the correct link.


keflavich:
Good catch!

PR #597, opened 2015-10-21
Update IRSA dust documentation

evertrol:
Now that I'm at it, might as well update the docs on the IRSA dust extinction service.

### DIFF
--- a/docs/irsa/irsa_dust.rst
+++ b/docs/irsa/irsa_dust.rst
@@ -122,28 +122,39 @@ value:
     >>> C = coord.SkyCoord("22h57m57.5s +26d09m00.09s", frame='fk4')
     >>> table = IrsaDust.get_extinction_table(C)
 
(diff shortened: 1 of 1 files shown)


keflavich:
I gather the examples were out of date?  Are the tests still OK?  I'm a little confused by this; did something change upstream at IRSA?

evertrol:
@keflavich Yes, sorry, the PR comment isn't that clear. But see the (extended) commit message itself: the IRSA server has changed, and the returned data contains a bunch of extra columns and rows.

The Travis build & test seems to be happy (though I'm not sure if this example is doc-tested anyway, since the resulting URL is variable).

keflavich:
@evertrol no, we don't run the doctests, in part because they require running a remote query, and in part because there are many queries that are not expected to return consistent results.  Anyway, thanks, this looks good.

Issue #599, opened 2015-10-24
Setting up new tests for a module


agroener:
Hi, It's been quite a while since I've worked on the CosmoSim module. One of the biggest loose ends which I would like to address first (before moving on to new features), is creating a solid test structure of its foundation. An issue is already open for the creation of tests for CosmoSim (https://github.com/astropy/astroquery/issues/457), however, I'm having some issues with running the tests.

Firstly, how can I hook my current branch into this issue?

keflavich:
@agroener to hook a branch into an issue and make a pull request, you need write access to the repository - so if you want me to do that, just let me know the branch name and I will

bsipocz:
You can do this through the GitHub api. There's a tools repo and an
issuetopr script either under the astropy user or @eteq, or I also must
have a fork of it (sorry I can't check it ATM as I'm away from laptops).

Anyway with that script you can turn an issue to a PR using a branch of
yours without write access.

agroener:
@keflavich The branch is: cosmosim-latest-unittests

keflavich:
@agroener since #600 already exists, we can't attach a new PR.  We'll just close this issue when that one is merged.  Let's move discussion there.

bsipocz:
I'm closing this in favour of https://github.com/astropy/astroquery/issues/457. Also, there are more relevant info in linked, closed PRs.

PR #600, opened 2015-10-24
Cosmosim latest unittests

agroener:
Picking up where I left off regarding the creation of tests for the CosmoSim module. This PR is ideally going to replace a previous one (https://github.com/astropy/astroquery/pull/458) once it is completely up and running, ideally. This is also related to a current issue (https://github.com/astropy/astroquery/issues/599).

### DIFF
--- a/astroquery/cosmosim/core.py
+++ b/astroquery/cosmosim/core.py
@@ -26,8 +26,11 @@ conf.max_width = -1
 from ..query import QueryWithLogin
 from . import conf
 
-__all__ = ['CosmoSim']
+# temp imports
+import ipdb
 
+#__all__ = ['CosmoSim']
+__doctest_skip__ = ['CosmoSim.*']
 
 class CosmoSimClass(QueryWithLogin):
 
@@ -59,17 +62,17 @@ class CosmoSimClass(QueryWithLogin):
         if not password_from_keyring:
             logging.warning("No password was found in the keychain for the provided username.")
             if password:
-                self.password = password
+                self._password = password
             else:
-                self.password = getpass.getpass("{0}, enter your CosmoSim password:\n".format(self.username))
+                self._password = getpass.getpass("{0}, enter your CosmoSim password:\n".format(self.username))
         else:
             logging.warning("Using the password found in the keychain for the provided username.")
-            self.password = password_from_keyring
+            self._password = password_from_keyring
 
         # Authenticate
         warnings.warn("Authenticating {0} on www.cosmosim.org...".format(self.username))
         authenticated = self._request('POST', CosmoSim.QUERY_URL,
-                                      auth=(self.username, self.password),
+                                      auth=(self.username, self._password),
                                       cache=False)
         if authenticated.status_code == 200:
             warnings.warn("Authentication successful!")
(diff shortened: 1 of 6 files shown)


astroquery/cosmosim/core.py
```diff
@@ -379,8 +387,8 @@ def check_all_jobs(self, phase=None, regex=None, sortby=None):
             if not phase and not regex:
                 if not sortby:
                     t = Table()
-                    t['JobID'] = self.job_dict.keys()
-                    t['Phase'] = self.job_dict.values()
+                    t['JobID'] = [i for i in self.job_dict.keys()]
```
keflavich: could use `list(self.job_dict.keys())` for the same effect

astroquery/cosmosim/tests/test_cosmosim.py
```diff
@@ -0,0 +1,44 @@
...
+try:
+    from ...cosmosim import CosmoSim
+    COSMOSIM_IMPORTED = True
+except:
+    COSMOSIM_IMPORTED = False
+from ...exceptions import LoginError
+SKIP_TESTS = not(HAS_KEYRING and COSMOSIM_IMPORTED)
+
+@pytest.mark.skipif('SKIP_TESTS')
+@remote_data
```
keflavich: since this is remote, it should go into `the test_cosmosim_remote` file

keflavich:
@agroener your testing issues are arising because you `import ipdb`. It seems that ipdb and pytest are incompatible

astroquery/cosmosim/core.py
agroener: Ah, yes. I had this in another branch but forgot to add it here. Oddly enough, casting to a list fails. I had to actually do a list comprehension to fix this.

agroener:
Ok. I've written some tests for the login/logout procedure for cosmosim. I've pushed up my changes because at this point in time, keyring is now doing something unexpected (since the last time I used it). It now is asking me for:

    Please enter password for encrypted keyring:

To remind ourselves, cosmosim uses the <code>login</code> method from its base class <code>QueryWithLogin</code>. Either keyring has changed in some way, or the implementation of this method has. Any thoughts?

keflavich:
`keyring` requires you to enter your global password to get access to any hidden passwords.  On Mac OS, this is done with the OS Keychain app, and on linux it's done with something called 'secret service': https://pypi.python.org/pypi/keyring.  So I think that message is most likely coming from secret service and you're on a linux box?  Otherwise I don't know what it means exactly

agroener:
Okay, that makes a bit more sense now. I did most of my developing on OS X previously. Now, I'm on linux, which is why the expected keyring behavior is differing. I'll go back to OS X for now, but I need to normalize the behavior after I get all of this working. I don't want this module to be platform dependent in any way.

keflavich:
@agroener where do we stand with this and #458?

agroener:
Well, I had intended to complete both a few weeks ago, but once I began working, I realized there were some minor discrepancies with keyring between OS X and Linux which was causing some unexpected login behavior. I've since fixed the issue, and things seem to be normalized a bit. The tests which I spent time on this last round of coding were regarding login/logout, of which there are a number of possible combinations when storing/deleting credentials are considered. 

I'll see if I can get around to finishing off login/logout and query tests in the next couple of weeks, pending my work schedule.

keflavich:
This has lingered for a little more than a week.  I'm closing it.  Please reopen if you complete it.

PR #603, opened 2015-10-27
Make Vizier.find_catalogs and get_catalogs play nice together

eteq:
This fixes a rather confusing bit of behavior I noticed when using the `Vizier` query.  I would have thought that the resource objects you get back from `find_catalogs` would be valid things to pass into `get_catalogs`.  But in master, only the _keys_ are valid.  This makes a very straightforward change that uses duck-typing to accept the resource objects.

It also adds a test of exactly this use case (and the relevant response file and related machinery).

### DIFF
--- a/astroquery/vizier/tests/test_vizier.py
+++ b/astroquery/vizier/tests/test_vizier.py
@@ -182,9 +182,15 @@ def test_get_catalogs(patch_post):
 def test_find_resource_then_get(patch_post):
     reses = vizier.core.Vizier.find_catalogs('2009ApJ...706...83K')
     res = next(iter(reses.values()))
+
     result = vizier.core.Vizier.get_catalogs(res)
     assert isinstance(result, commons.TableList)
     assert next(iter(result.keys())).startswith(res.name)
+
+    resultl = vizier.core.Vizier.get_catalogs([res])
+    assert type(result) == type(resultl)
+    assert len(result) == len(resultl)
+    assert len(result[0]) == len(resultl[0])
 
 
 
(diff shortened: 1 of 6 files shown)


astroquery/vizier/tests/test_vizier.py
```diff
@@ -159,6 +170,14 @@ def test_get_catalogs(patch_post):
     result = vizier.core.Vizier.get_catalogs('J/ApJ/706/83')
     assert isinstance(result, commons.TableList)
 
+def test_find_resource_then_get(patch_post):
+    reses = vizier.core.Vizier.find_catalogs('2009ApJ...706...83K')
+    res = reses.values()[0]
+    result = vizier.core.Vizier.get_catalogs(res)
+    assert isinstance(result, commons.TableList)
+    assert result.keys()[0].startswith(res.name)
```
keflavich: this is a python2-only idiom.  `tuple(results.keys())[0]` would work.  Ditto line 175 above

keflavich:
Looks good besides the minor bug.  Consider adding an example to the vizier docs?

keflavich:
ping @eteq - want to finish this off?

eteq:
@keflavich at long last (2 years later...) I think this is ready to go (although the tests will say for sure...)

keflavich:
We're still dealing with the py2.7 - astropy-helpers bug; you may have passed the tests.  I'll merge once we work out our CI issues.

keflavich:
@eteq looks like we dealt with CI issues... but the PR has gone out of date.  Mind rebasing?

eteq:
Thanks for doing the rebase I failed to do @keflavich. Assuming tests pass I think this is good to go!

PR #615, opened 2015-11-16
Fix a python3 string handling error

keflavich:
I'm open to suggestions on how to better handle this kind of thing.  I would really like to not have to worry about my string types every time I use a string.


astroquery/alma/utils.py
```diff
@@ -84,7 +84,13 @@ def parse_frequency_support(frequency_support_str):
 
         '[86.26..88.14GHz,976.56kHz, XX YY] U [88.15..90.03GHz,976.56kHz, XX YY] U [98.19..100.07GHz,976.56kHz, XX YY] U [100.15..102.03GHz,976.56kHz, XX YY]'
     """
-    supports = frequency_support_str.split("U")
+    try:
+        supports = frequency_support_str.split("U")
+    except TypeError:
+        # python 3 problem
+        frequency_support_str = frequency_support_str.tostring().decode('utf-8')
+        supports = frequency_support_str.split("U")
```
astrofrog: Would ``str(frequency_support_str).split('U')`` work?

PR #641, opened 2016-01-15
Xmatch: allow astropy tables input iwht more than 2 columns

hamogu:
Before this patch astropy tables as input could only have exactly
    two columns that would be exactly in the order RA, DEC. That is a bad
    design for two reasons:
    - What is the order of the columns is different.
    - After cross-matching, how do I know which object is which?
      Two RA/DEC columns are fine if I want to answer the question
      "What are the counterparts in catalog 2?" but not if I want
      to retrieve a joint catalog (e.g. I have my own optical data
      in cat1 and cross-match with 2MASS, then I want all mags in the
      returned catalog).


keflavich:
Looks good.  The travis failure is just something minor in the docstring:
https://travis-ci.org/astropy/astroquery/jobs/102496021#L1162

Fix that, and we're good to go.  Thanks!

hamogu:
I'm a little puzzled here: https://travis-ci.org/astropy/astroquery/jobs/102581753#L1162
I fixed it in `query` and the `query` -related warning are gone. However, it's still there for `query_asnyc`. How can that be? `query_async` does not even have a docstring, it's copied over from `query` using the `prepend_docstr_noreturns` decorator.

keflavich:
I'm very puzzled.  The docstrings are identical but sphinx is complaining about one.  Very weird.  

I think this is an existing uncaught error that we didn't catch because `automodapi` wasn't being used before.

hamogu:
It turn out that the first line of the docstring has to be empty, because that is how the decorator determines the number of white spaces to strip, i.e.
```
def func():
    '''
    This is a function that does ...

    Returns
    -----------
    stuff : 
'''
```
instead of the usual numpydoc format.
```
def func():
    '''one line summary of function

    Longer text that describes function in detail.

    Returns
    -----------
    stuff : 
'''
```
I'll open an issue for that, but for now I've fixed it.

keflavich:
:+1:

PR #671, opened 2016-04-05
Use HARPS new query form

j-faria:
[This ESO form](http://archive.eso.org/wdb/wdb/eso/repro/form) now states that "HARPS echelle non-polarimetric science products have been migrated to the science data products query form(s), and are no longer available via this interface".
This means that the interface to HARPS data is broken: `eso.query_instrument('harps', help=True)` returns nothing. 
I updated the url to the new form [here](http://archive.eso.org/wdb/wdb/adp/phase3_spectral/form?phase3_collection=HARPS) but the html in this page is different. 
This PR fixes the queries of HARPS data and I think it doesn't mess up anything else. 
Feel free to comment on my parsing of the html, I'm not sure if it's the best implementation.


bsipocz:
Thanks @j-faria for the PR.
We are aware that ESO is transferring to have new query forms (happening this week and next) and this affects not just HARPS. We plan to upgrade astroquery accordingly in the coming days.

j-faria:
I wasn't aware that other query forms were changing too. Then it's better to wait and upgrade everything together. I mostly query the ESO archive for HARPS data but let me know if I can help.

astroquery/eso/core.py
```diff
@@ -640,43 +642,78 @@ def _print_help(self, url, instrument, cache=True):
         resp = self._request("GET", url, cache=cache)
         doc = BeautifulSoup(resp.content, 'html5lib')
         form = doc.select("html body form pre")[0]
-        for section in form.select("table"):
```
bsipocz: Taking out this part of the code and duplicating below is not the best coding practice. Maybe move the if statement above and define a variable there, and use it here to loop over?

astroquery/eso/core.py
```diff
@@ -640,43 +642,78 @@ def _print_help(self, url, instrument, cache=True):
...
-                if tag.name == u"table":
-                    break
-                elif tag.name == u"input":
-                    if tag.get(u'type') == u"checkbox":
-                        checkbox_name = tag['name']
-                        checkbox_value = u"[x]" if ('checked' in tag.attrs) else u"[ ]"
+
+        if instrument in ('feros', 'harps'):
+            for section in form.select("fieldset"):
+                section_title = "".join(section.stripped_strings.next())
```
bsipocz: This part of the code doesn't have test coverage that's been checked on travis, but this line should fail with python3.

astroquery/eso/core.py
```diff
@@ -640,43 +642,78 @@ def _print_help(self, url, instrument, cache=True):
...
+                print(section_title)
```
bsipocz: Please do remove these debug print statements.

keflavich:
Thanks @j-faria.  Your work here might be the right template for modifying the rest of the ESO forms.  We welcome the assistance in transitioning to the new forms!  You can try running `python setup.py test -P eso --remote-data` to check if the code works on the remote service with the update URLs etc.

bsipocz:
I've left a few comments for the code, however I can't guarantee how much refractoring we need to do on the code (due to the eso changes I mentioned before), so it's completely possible that all my comments become irrelevant.

Also if you have time and interest we welcome any help in this work. I'll ping @keflavich to set up a time for a sprint maybe?

j-faria:
Thank you @bsipocz and @keflavich for the comments. I will try to run the tests and maybe add a new one too. I'd be up for a sprint to get this going. Do we know which query forms are changing and when?

keflavich:
Is this PR still needed?  If so, can we work on it this or next week with the aim of including it in a new release?

bsipocz:
I'm away this week but can spend an afternoon/day on the eso module next
week.

bsipocz:
I believe this was fixed in the meantime by #840.

Thanks @j-faria for working on it though.

PR #680, opened 2016-04-24
Simbad flux qual fix

zefrawg:
Astroquery is not returning flux quality from SIMBAD because it is using a misnamed votable field. This change fixes the problem by renaming the field.

### DIFF
--- a/astroquery/simbad/data/votable_fields_dict.json
+++ b/astroquery/simbad/data/votable_fields_dict.json
@@ -27,7 +27,7 @@
   "flux_bibcode(filtername)": "bibcode", 
   "flux_error(filtername)": "error value of the flux measurement", 
   "flux_name(filtername)": "name of the filter", 
-  "flux_quality(filtername)": "quality (A:best, E:worst) of the flux", 
+  "flux_qual(filtername)": "quality (A:best, E:worst) of the flux", 
   "flux_unit(filtername)": "flux unit (mag, jy, ...)", 
   "fluxdata(filtername)": "all fields related with a particular filter.", 
   "gcrv": "General Catalogue of Radial Velocities", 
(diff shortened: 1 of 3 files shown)


keflavich:
This appears to be incorrect on the SIMBAD help page, then:
http://simbad.u-strasbg.fr/simbad/sim-help?Page=sim-fscript#VotableFields

@zefrawg, thanks for the fix.  Could you provide a test, or a specific example we can use to verify the change?

zefrawg:
Yeah, it seems wrong on their help page. Here's a test post-change. Not sure if that's what you're looking for? I'm a newb to contributing.
```python

def simbad_flux_qual_fix_test():
	request = Simbad()
	request.add_votable_fields("flux_qual(V)")
	response = request.query_object('algol')
	return "FLUX_QUAL_V" in response.keys()
```

keflavich:
@zefrawg that's great!  You can put that almost verbatim into [`simbad/tests/test_simbad_remote.py`](https://github.com/astropy/astroquery/blob/master/astroquery/simbad/tests/test_simbad_remote.py), but rename it `test_simbad_flux_qual` and perhaps add a docstring "Regression test for issue 680".

Then, instead of `return`, use `assert`, which will raise an exception if it fails.

I'll also report this to CDS.

zefrawg:
Added. Cheers and thanks.

keflavich:
I've made a few edits (see https://github.com/astropy/astroquery/commit/8cd708c740602ac98a29b6ebd2bdd9ed10d6c65b) and merged this.  Thanks!

Tips for the future:

 - make sure your editor doesn't use tabs for indentation
 - class methods (functions that are part of a class) need to have `self` as the first parameter

PR #700, opened 2016-06-06
Get all SIMBAD object IDs when using query_criteria

keflavich:
I wasn't sure if this is currently possible, and in playing around with the syntax of `query_criteria` I wasn't able to accomplish this, but I'd like to be able to return all object IDs for each row within a `query_criteria` call. Currently I have to pass the list one by one to `query_objectids`, which is horribly slow. It appears that the `ids` votable tag is not currently supported, but there is an `id(opt)` tag which I cannot get to return anything more than the main ID.

Is this possible with the current version of astroquery?


keflavich:
the attached PR now includes the updated votable fields, which include 'ids'.  However, when I did an `ids` query, the `ids` column was empty.  `id` had comma-separated identifiers.  Perhaps that's an upstream error.

guillochon:
I verify that `id` does return a comma-delimited list of names, so for me at least the PR isn't necessary, but I guess if names are allowed to have commas it makes sense to use `ids`, if it is working.

keflavich:
Thanks for verifying.  I'll merge this because updating based on the docs is the right thing to do anyway.

PR #738, opened 2016-09-06
ALMA: improve repeat staging behavior

keflavich:
Fixes an issue where repeated attempts to stage the same UID will give an HTTP
405 error.  The fix is to cache the results then return them (with an
appropriate warning message) if 405 is encountered.

### DIFF
--- a/astroquery/alma/core.py
+++ b/astroquery/alma/core.py
@@ -235,10 +235,10 @@ class AlmaClass(QueryWithLogin):
         log.debug("First response URL: {0}".format(response.url))
         if response.status_code == 405:
             if hasattr(self,'_last_successful_staging_log'):
-                print("Error 405 received.  If you have previously staged "
-                      "the same UIDs, the result returned is probably "
-                      "correct, otherwise you may need to create a fresh "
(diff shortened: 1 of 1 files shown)


astroquery/alma/core.py
```diff
@@ -234,9 +234,16 @@ def stage_data(self, uids):
         self._staging_log['initial_response'] = response
         log.debug("First response URL: {0}".format(response.url))
         if response.status_code == 405:
-            raise HTTPError("Received an error 405: this may indicate you "
-                            "have already staged the data.  Try downloading "
-                            "the file URLs directly with download_files.")
+            if hasattr(self,'_last_successful_staging_log'):
+                print("Error 405 received.  If you have previously staged "
```
MSeifert04: I think a `log.info` message would be more appropriate then `print` but I don't know how you handled this in the package.

keflavich: Agreed.  used `log.warn` instead

keflavich:
Neither of the "failures" are relevant - I can't readily add test code for this, and the appveyor failure is something upstream.

PR #740, opened 2016-09-08
decode row run2d to ensure there is no binary prefix included in sdss url

### LINKED ISSUE #739: SDSS DR12 query returns wrong url: HTTP Error 404: Not Found
This problem happens when I tried to get a spectrum with the DR12 Plate MJD FiberID parameters, it does not crash if I used the DR7 parameters.

This is easier to explain with an example:

```
s = SDSS.get_spectra(plate=7167, mjd=56604, fiberID=182) 
```

returns `HTTPError: HTTP Error 404: Not Found`

This is because it is querying the wrong url:

```
s2 = SDSS.get_spectra_async(plate=7167, mjd=56604, fiberID=182)
[Downloaded object from URL http://data.sdss3.org/sas/dr12/boss/spectro/redux/b'v5_7_0'/spectra/7167/spec-7167-56604-0182.fits with ID 5558563336]
```

The problem is with the **b'v5_7_0'** part it should be just **v5_7_0** without the ' '

And the correct URL that takes you to download the spectrum is 

```
http://data.sdss3.org/sas/dr12/boss/spectro/redux/v5_7_0/spectra/7167/spec-7167-56604-0182.fits
```

I'm sure the problem is in this [part](https://github.com/astropy/astroquery/blob/master/astroquery/sdss/core.py#L582), especifically in the {run2d} argument.

This should be easy to fix but I could not find the root of the problem in the code

MSeifert04:
Fixes #739 by decoding `run2d` if it's a byte type.

However I'm not sure how one could test this with the test suite.

### DIFF
--- a/astroquery/sdss/core.py
+++ b/astroquery/sdss/core.py
@@ -575,11 +575,17 @@ class SDSSClass(BaseQuery):
         results = []
         for row in matches:
             linkstr = self.SPECTRA_URL_SUFFIX
-            # _parse_result returns bytes for instruments, requiring a decode
+            # _parse_result returns bytes (requiring a decode) for
+            # - instruments
+            # - run2d sometimes (#739)
+            if isinstance(row['run2d'], bytes):
+                run2d = row['run2d'].decode()
+            else:
+                run2d = row['run2d']
             link = linkstr.format(
                 base=conf.sas_baseurl, dr=data_release,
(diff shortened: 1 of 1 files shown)


MSeifert04:
should I rebase and push to trigger an AppVeyor build?

keflavich:
Yes, seems like a good idea.

MSeifert04:
:+1:

Issue #748, opened 2016-09-17
Build failure: qt4?


keflavich:
The latest builds fail with this after I merged the latest (green) PRs:

https://travis-ci.org/astropy/astroquery/jobs/160595430

I suspect this is something temporary related to conda?

MSeifert04:
Yes, probably related to the PyQT 5.6 release on conda (yesterday).

MSeifert04:
Just to link it here as well: The underlying issue is some patch with matplotlib rc-something, also reported here: https://github.com/ContinuumIO/anaconda-issues/issues/1068.

PR #749, opened 2016-09-17
Pin QT version in CI files

MSeifert04:
I don't know if this should be merged. But I'm too inexperienced for submitting something to `ci-helpers` and this is should be definitly fixed somewhere upstream.

This fixes `QT` and `PyQT` to version 4 to avoid the test failures mentioned in #748

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,11 +27,11 @@ env:
         - MAIN_CMD='python setup.py'
         - NUMPY_VERSION=stable
         - ASTROPY_VERSION=stable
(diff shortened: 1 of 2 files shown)


MSeifert04:
@keflavich Pinning the version makes the tests pass again - but I assume this should be fixed upstream. However I have no clue where.

cc @astrofrog @bsipocz

keflavich:
I'll merge this here - thanks @MSeifert04! - but as you said, it probably warrants an upstream change too.

PR #753, opened 2016-09-20
fix parsing of radius option in Simbad queries.

segasai:
Hi, 

Currently radius is incorrectly parsed when doing Simbad queries: 
Here is the illustration of the problem 

```
In [1]: import astropy.units as u
In [2]: import astroquery.simbad.core
In [3]: astroquery.simbad.core._parse_radius(1.92*u.degree)
Out[3]: '1d'
```
My patch fixes that. I also added a regression test for this issue.

```
In [1]: import astropy.units as u
In [2]: import astroquery.simbad.core
In [3]: astroquery.simbad.core._parse_radius(1.92*u.degree)
Out[3]: '1.92d'
```

Cheers,
      Sergey

### DIFF
--- a/astroquery/simbad/core.py
+++ b/astroquery/simbad/core.py
@@ -811,13 +811,11 @@ def _parse_radius(radius):
             index = 2  # use arcseconds when radius smaller than 1 arcsecond
         unit = ('d', 'm', 's')[index]
         if unit == 'd':
-            return str(int(angle.degree)) + unit
+            return str(angle.degree) + unit
         if unit == 'm':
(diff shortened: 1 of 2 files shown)


keflavich:
Thanks for this fix.  I think the original code was intended to parse radii like `5d10m25s`, which is usually intended to be used as a coordinate rather than a radius.  I'd like to review this once more before merging, just to make sure there aren't any hidden unintended consequences I'm missing.

segasai:
Thanks for merging!

PR #754, opened 2016-09-21
Update Splatalogue species name json table

keflavich:
Also updates the code used to generate it to suppress a warning and to make the script independently runnable.

I should run this regularly, but the .json table is currently a file we
serve... the better solution is to reproduce it on the fly, which requires
BeautifulSoup.  Probably just requiring that in the future is the better choice
than serving our own mirror of the .json file.


keflavich:
If you can't find the new OSU data using splatalogue, this is why (the species have different names in that catalog)

PR #757, opened 2016-09-27
Import parse_lamda_datafile into package level

migueldvb:
This function is useful to read locally modified versions of molecular datafiles in LAMDA format.


keflavich:
Thanks, this is a nice clarification.  The build failure has to do with a Sphinx-related build timeout, but since I can't see anything wrong I think it is mergeable.  I'll try restarting the docs build once more later today, then just merge.

PR #765, opened 2016-10-11
ESASKY: minor documentation fixes

keflavich:
attn @imbasimba

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,7 +2,7 @@
 ---------
 
 - Option to toggle the display of the download bar (#734)
-- ESASKY - added new module for querying the ESASKY archive (#758,#763)
+- ESASKY - added new module for querying the ESASKY archive (#758,#763,#765)
 
 0.3.2 (2016-06-10)
 ------------------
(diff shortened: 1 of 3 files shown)


keflavich:
@imbasimba some clarification questions from the docs:

"For catalogs, the query returns a maximum of 2000 sources per mission. To account for observation errors, this method will search for any sources within 5 arcsec from the object."
is this limitation imposed by the server or is it a controllable parameter within astroquery?  It seems that it is currently a non-user-modifiable string `sourceLimit` that is presumably interpreted by the server?  Do you want this to be user-modifiable?

imbasimba:
@keflavich The changes looks good to me.
I believe the source limit is used to ease the load on the server. But I'm not sure if this limitation makes sense in this context... I will ask the other guys on the project for their opinions and come back to you.

keflavich:
OK.  In other modules, we've set up a `ROW_LIMIT` configuration item that defaults to something small (to ease server loads by default), but can be made larger if needed.  If the server doesn't support this, we should keep the hard-coded limit in.  Otherwise, we can consider changing it.

imbasimba:
Yes, I too believe a parameter would be a good solution here, but I will await the opinion of the guy responsible for the server before making that change,

PR #771, opened 2016-10-18
Exoplanet catalog queries

bmorris3:
This PR is an API proposal in response to #764. Here I develop the `PlanetParams` object, which contains parameters of exoplanets, parsed from either exoplanets.org or NExScI's Exoplanet Archive.

Here's the docs page that I've drafted for the module, which demonstrates the API:
screen shot 2016-10-17 at 10 02 11 pm

Here are some potential details to discuss: 
1. The parameters returned by the two archives are often non-overlapping, and when the same parameter is available in both archives, they're stored in attributes with different names. I prefer keeping the attribute names the same as the table columns, though, since users of each table know which parameters to expect.
2. There's already an Open Exoplanet Catalog API within astroquery, which falls outside of the `exoplanets` module I'm proposing here. Should I find a way to incorporate that within this module?
3. I included two new JSON files in the [`astroquery/data` directory](https://github.com/bmorris3/astroquery/tree/exoplanets/astroquery/data) which contain the units for each unitful table column in the exoplanet tables. Do you have any objections to including these "data" in that directory?

cc @bsipocz @barentsen @lkreidberg @jlustigy


keflavich:
(1) seems fine
(2) Not necessarily, since they are independent entities.  Vizier and IRSA have some overlap (e.g., both contain the 2MASS and IRAS catalogs).  However, it would be helpful to note & describe the overlap.
(3) these should probably be in `astroquery/exoplanets/data` instead

bmorris3:
I've moved the data to `astroquery/exoplanets/data`. 

I think the astropy `lts` builds are failing because I use units in the exoplanets table that were recently introduced to astropy (astropy/astropy#4818). Which version is `lts`? 

How do you think I should I approach this – should we mark the exoplanets tests only for astropy > 1.2?

keflavich:
@bmorris3 yes, have the tests that rely on that check the astropy version.  Maybe also add a check in the module itself to raise a `NotImplementedError` if the astropy version is too low

bmorris3:
Update on the docs page to include the `PlanetParams.coords` docs:

screen shot 2016-10-18 at 4 04 08 pm

bmorris3:
The last remaining [build failure](https://travis-ci.org/astropy/astroquery/jobs/168772557#L1438) is a sphinx warning about references: 
```
<autosummary>:1: WARNING: citation not found: R1
<autosummary>:1: WARNING: citation not found: R3
<autosummary>:1: WARNING: citation not found: R15
<autosummary>:1: WARNING: citation not found: R16
```

Do you know if I am [using url references](https://github.com/bmorris3/astroquery/blob/f55000e621dfd54ea1fad333b264669630004237/astroquery/exoplanets/planet_params.py#L95-L96) incorrectly?

astroquery/exoplanets/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+from .exoplanet_archive import *
+from .exoplanets_org import *
+from .planet_params import *
```
keflavich: Could you use explicit imports here?  Especially since there is no `core.py`, it is helpful to be able to look at `__init__.py` to determine what's in the exoplanets namespace.

astroquery/exoplanets/exoplanet_archive.py
```diff
@@ -0,0 +1,82 @@
...
+TIME_ATTRS = {'rowupdate': 'iso'}
+BOOL_ATTRS = ('pl_kepflag', 'pl_ttvflag', 'pl_k2flag', 'st_massblend',
+              'st_optmagblend', 'st_radblend', 'st_teffblend')
+
+
+def exoplanet_archive_units():
+    """
+    Dictionary of exoplanet parameters and their associated units.
+    """
+    global PARAM_UNITS
```
keflavich: Seeing `global` here is scary.  I see why you're doing it, and I think it's safe enough.  Generally, the "better" approach is to make a `class` that will load & cache the data the first time it is called.

astroquery/exoplanets/exoplanet_archive.py
```diff
@@ -0,0 +1,82 @@
...
+        available). Default is `True`.
+
+    Returns
+    -------
+    table : `~astropy.table.Table`
+        Table of exoplanet properties.
+
+    References
+    ----------
+    .. [1] http://exoplanetarchive.ipac.caltech.edu/index.html
```
keflavich: This looks like it formats right when I build locally, so I think it's fine?

image

astroquery/exoplanets/tests/test_params.py
```diff
@@ -0,0 +1,114 @@
...
+    assert params.pl_hostname == 'HD 209458'
+    assert_quantity_allclose(params.pl_orbper, 3.52474859 * u.day,
+                             atol=1e-5 * u.day)
+
+    assert not params.pl_kepflag
+    assert not params.pl_ttvflag
+
+
+@remote_data
+@pytest.mark.skipif('APY_LT12')
```
keflavich: It would be good to have one tests that runs these non-astropy<1.2-valid queries and checks that the appropriate exception is raised.

astroquery/exoplanets/exoplanet_archive.py
keflavich: Actually, maybe it should be more like this:

```
Download (and optionally cache) the `exoplanets.org planets table`_.

References
----------
.. _exoplanets.org planets table: http://www.exoplanets.org
```

astroquery/exoplanets/exoplanet_archive.py
bmorris3: I agree that it's scary – it's a bad habit. I'll reimplement in a class format.

keflavich: Yes, that's about right.  Something like:
```
class ExoplanetTable(object)
    @property
    def PARAM_UNITS(self):
        if hasattr(self, '_param_units'):
              return self._param_units
        else: 
              # populate self._param_units
```
is a common pattern I use.  I don't like the all-caps function name in this specific case, but something along these lines should work...

bmorris3:
I think the errors remaining at this point are raised by another part of astroquery due to the dev version of astropy.

keflavich:
Looks good.  Are you ready for it to be merged?

I've seen those failures on other builds too.  I think @mhvk mentioned this on astropy-dev.  [This thread](https://github.com/astropy/ci-helpers/issues/137) is supposed to answer appveyor fails, but we're seeing travis fails, not appveyor...

bmorris3:
> Looks good. Are you ready for it to be merged?

I think so, but I'd love your comments on the docs. Are there sufficient examples? I wanted to show some more outputs, but there are tons of columns in these tables.

mhvk:
The failures look real, though would seem `astropy.wcs` related, with the views passed in not converted to int when turned into array indices (which has become required in numpy relatively recently). Would be good to raise this issue on astropy itself (ideally with a minimal example, of course...)

keflavich:
Ahh, sorry, the errors are different than last time I looked; previously it was an install error affecting these two versions.  I suspect it's related to this: https://github.com/astropy/astropy/pull/5411 but I'll investigate more.

keflavich:
@bmorris3 if you rebase, it should fix the build.

More examples are always good, but I can't suggest what to add.

bsipocz:
@bmorris3 - would you mind rebasing again? It seems that quite a few unrelated commits are picked up during the last one.

bmorris3:
@bsipocz What could I have done wrong? I did: 
```
git fetch upstream
git rebase upstream/master
```

Now when I do that command a second time I get tons of merge conflicts which are just previous commits.

bsipocz:
@bmorris3 - I'm heading home, but will definitely look at it later tonight. 
My first guess is that you see more if you do an interactive rebase, at least easier to spot where it goes wrong.

keflavich:
`git rebase -i upstream/master` is what I do.  I don't know what happened to yours, and I'm not sure a different rebase approach will fix it.  Rebasing is a tricky art.

MSeifert04:
> Ahh, sorry, the errors are different than last time I looked; previously it was an install error affecting these two versions. I suspect it's related to this: astropy/astropy#5411 but I'll investigate more.

I haven't realized when making that PR that it imposes another restriction. Could you raise an issue in astropy and cc me?

barentsen:
Hi @bmorris3, thanks a lot for working on this!  :clap:  Simplifying access to these two exoplanet archives is a valuable thing to do. I believe this would be a hot :fire: feature that would get a lot of use.

I found time to play with `astroquery.exoplanets` in detail today.  I identified six issues which I hope you may want to consider. Apologies in advance if these have already been discussed elsewhere. (I also include a few trivial comments at the end.)

Major questions:

1. Is there a reason to have this code live under `astroquery.exoplanets` as opposed to adding the modules `astroquery.exoplanet_archive` and `astroquery.exoplanets_org`?  I say this because all existing modules in astroquery seem to represent a data provider rather than a field of science.  In fact there is already a module called `astroquery.open_exoplanet_catalogue` which will be confusing to have alongside `astroquery.exoplanets`.

2. I believe you forgot to list `astroquery.exoplanets` under "Available Services" on the front page of the docs.  It is listed further down under "Catalogs" in the by-type listing, which I find confusing because this is an interface to two archives and so I would have looked for it under "Archives".  Of course this relates to question 1.

3. Why is the user required to call `PlanetParams.from_exoplanet_archive()` rather than `ExoplanetArchive.get_planet()`.  Is there a rationale behind this design decision? The latter may be preferred because it reduces the number of different classes a user needs to interact with, and matches the concept of having users interact with specific services rather than 'exoplanets'. I certainly haven't thought carefully about this however.  (PS: Do other astroquery services ever return a 'star' or 'galaxy' object?)

4. The NASA Exoplanet Archive [serves 26 different tables](http://exoplanetarchive.ipac.caltech.edu/docs/program_interfaces.html#data) altogether, including e.g. planet candidates from different missions, but `ExoplanetArchive.get_table()` only returns one specific table (i.e. the table of confirmed planets).  We should consider renaming `get_table()` into e.g. `get_exoplanets_table()` to make this explicit, or alternatively we should accept the table name as a parameter (the confirmed planets, i.e. the table with name `exoplanets`, would be a natural default).

5. The exoplanet archive only returns a subset of columns from its tables by default. One needs to add `&select=*` to the API call to get all columns (or `&select=name1,name2,...` for a subset). It would be nice if `ExoplanetArchive.get_table()` had an option to specify the columns desired.

6. It would be helpful for the docs to contain a direct link to the page at the NASA Exoplanet Archive that documents the exact meaning of each column in the confirmed exoplanets table (URL: http://exoplanetarchive.ipac.caltech.edu/docs/API_exoplanet_columns.html ).  If we decide to accept the table name as an option, then we should also link to the page that explains the various tables they serve (URL: http://exoplanetarchive.ipac.caltech.edu/docs/program_interfaces.html#data ). 

Finally, some minor suggestions that are trivial:
* Please consider adding an `__all__` to `exoplanets/__init__.py`.  It would prevent the objects that are being imported from `__future__` from being shown in the exoplanets namespace, which makes exploring the functionality from the command line easier.
* Consider adding a short docstring to the constructors of the `ExoplanetArchive` and `ExoplanetsOrg` classes for the same reason as above, to aid command line use as well.
* The api is very intolerant of white spaces and dashes in the planet names. For example, `PlanetParams.from_exoplanet_archive('Tres-2 b')` works but `PlanetParams.from_exoplanet_archive('Tres-2b')` or  `PlanetParams.from_exoplanet_archive('Tres2b')` doesn't. Do we want to enforce a standard, or can we make it more lenient? 
* The error message returned by e.g. ``PlanetParams.from_exoplanet_archive('Tres2b')`  is `ValueError: Planet "{0}" not found in exoplanets.org catalog`.  This contains two bugs: `{0}` and `exoplanets.org`.

I hope this is useful feedback.  Thanks again for working on this.

keflavich:
@barentsen Good points, especially about splitting into two modules.  I think we could retain `exoplanets` as a host for the various objects (e.g., `PlanetParams`) but have separate modules for each of the individual archive services.  @bmorris3, how does that sound to you?  @barentsen, if @bmorris3 is onboard, would you want to contribute some of those changes?  (PR on his PR)

bsipocz:
@bmorris3 - Re to the rebase (sorry it took so long to get back to this):

For some reason the commits are doubled in this branch. I'm not sure what went wrong and where, but the fix is easy:
- Do an interactive rebase ``git rebase -i astroquery/master ``. This will bring up your favourite editor with these commits:
```
pick c78df27 Init exoplanets module commit
pick d3aba17 New exoplanet parameter query API
pick 47f6ace Moving exoplanets units data to exoplanets/data, removing doctests.
pick 41b2e3e Adding package data setup_package script
pick 8370867 Adding minversion requirement for R_jup tests, adding coordinates
pick 04a55d1 Adding docs for the coord property [ci skip]
pick 5b8064e change the links
pick 0826bfe Moving to OO-style table caching
pick a156b02 Adding astroquery-style query_planet method
pick bf331c8 Adding more docs for downloading/querying tables
pick 211f142 more robust handling of units for astropy<1.2
pick 4ff784e Renaming test file
pick 7258f08 Init exoplanets module commit
pick 298d591 New exoplanet parameter query API
pick 61b8e7b Moving exoplanets units data to exoplanets/data, removing doctests.
pick 5bc64a2 Adding package data setup_package script
pick abb834f Adding minversion requirement for R_jup tests, adding coordinates
pick f55000e Adding docs for the coord property [ci skip]
pick aba04a5 change the links
pick 7621949 Moving to OO-style table caching
pick 566be7b Adding astroquery-style query_planet method
pick 43c4c1e Adding more docs for downloading/querying tables
pick ff229e1 more robust handling of units for astropy<1.2
pick afba8d7 Renaming test file
```

Simply keep only the first 12, delete the rest as they are pure repeats. I've checked, the commits are the same as the ones above them. Once doing that, there will be no conflicts, etc, and you will only need to force push it back here. If you want, you can check my copy of this branch, rebased here: https://github.com/bsipocz/astroquery/tree/bmorris3_exoplanets

bsipocz:
@bmorris3 - any chance you find time to finish this up? I would love to use it :)

bmorris3:
@bsipocz It's probably going to be a while – I'm slammed until after AAS 😢

bsipocz:
Thanks @bmorris3 to pick this up again. I have a few minor comments now, but will do a full review once you're ready.

 * please rebase, there shouldn't be any merge commits or changes to astropy_helpers
 * I wonder whether we could come up with a less informal name for ``exoplanets_org``, something that better reflects the underlying data of "Exoplanet Orbit Database"

bmorris3:
Hey @barentsen and @keflavich, I'm finally following up on Geert's awesome code review. I'm going to write up some notes on what I've done: 

1. I've split up the `exoplanets` module into three modules. There's an `exoplanets` module which houses the `PlanetParams` object, and could be the future home to any database-independent planet operations. 

2. I've added exoplanets under available services. 

3. The rationale behind `PlanetParams.from_exoplanet_archive()` rather than `ExoplanetArchive.get_planet()` gets a bit philosophical. Typically when you create a new instance of an object, you already know what attributes will be created on the new object. However, `PlanetParams` populates its attributes depending on what data source you give it. Since you can only know the attributes of a given `PlanetParams` object if you know where it came from, I wanted to make it explicit from what conditions you created each `PlanetParams` object. By making the constructor a class-method on `PlanetParams`, it ensures that the code that creates a new `PlanetParams` object always points back to its source. 

4. I see how that could be confusing. I've changed the name from `get_table` to `get_confirmed_planets_table` to make it clearer. 

5. I will welcome your PR making this change in the future! 

6. A direct link to the Exoplanet Archive column description has been added.

*** 

@barentsen Do you think you'll have a chance to review the changes sometime soon? I'd like to make a major update to astroplan at some point in the near future which will depend on this update to astroquery. Thanks again!

barentsen:
Awesome! I will try to review asap.

bmorris3:
@bsipocz I can't believe I didn't realize there was a better name than "exoplanets.org," that's all I've ever heard it pronounced out loud. Thanks for catching that! I've changed that name to "Exoplanet Orbit Database" everywhere it appears, and I've rebased.

astroquery/exoplanet_archive/exoplanet_archive.py
```diff
@@ -0,0 +1,104 @@
...
+__all__ = ['ExoplanetArchive']
+
+EXOPLANETS_CSV_URL = ('http://exoplanetarchive.ipac.caltech.edu/cgi-bin/'
+                      'nstedAPI/nph-nstedAPI?table=exoplanets')
+TIME_ATTRS = {'rowupdate': 'iso'}
+BOOL_ATTRS = ('pl_kepflag', 'pl_ttvflag', 'pl_k2flag', 'st_massblend',
+              'st_optmagblend', 'st_radblend', 'st_teffblend')
+
+
+class ExoplanetArchiveClass(object):
```
barentsen: Let's use `NasaExoplanetArchive`, which is the official full name of the archive and will avoid confusion with other archives?

astroquery/exoplanet_archive/exoplanet_archive.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Exoplanet Archive querying object. Use the ``get_table`` or ``get_planet``
+    methods to get information about exoplanets via the Exoplanet Archive.
+    """
```
barentsen: Also in the constructor comments, for clarity, let's call it the `NASA Exoplanet Archive` and mention its URL.

astroquery/exoplanet_archive/exoplanet_archive.py
```diff
@@ -0,0 +1,104 @@
...
+        table : `~astropy.table.Table`
+            Table of exoplanet properties.
+        """
+        if self._table is None:
+            table_path = download_file(EXOPLANETS_CSV_URL, cache=cache,
+                                       show_progress=show_progress, timeout=120)
+            exoplanets_table = ascii.read(table_path)
+
+            # Group by the host name
+            exoplanets_table.group_by('pl_hostname')
```
barentsen: Does this line serve a purpose?  `group_by` returns a new table but it is never used.

astroquery/exoplanet_archive/exoplanet_archive.py
```diff
@@ -0,0 +1,104 @@
...
+
+            # Store column of lowercase names for indexing:
+            lowercase_names = [' '.join([host_name.lower(), letter])
```
barentsen: The space between the host name and the planet letter is often dropped, e.g. people often say "Trappist-1b" instead of "Trappist-1 b".  Let's drop the space here, which will make it easier to support both "Trappist-1b" and "Trappist-1 b" down in `query_planet`.

astroquery/exoplanet_archive/exoplanet_archive.py
```diff
@@ -0,0 +1,104 @@
...
+            Name of planet
+
+        Return
+        ------
+        table : `~astropy.table.Table`
+            Table of one exoplanet's properties.
+        """
+
+        exoplanet_table = self.get_confirmed_planets_table()
+        return exoplanet_table.loc[planet_name.strip().lower()]
```
barentsen: Add `.replace(" ", "")` here to get rid of whitespaces, so that you can both support `query_planet("Trappist-1 b")` and `query_planet("Trappist-1b")`?  Right now the latter fails.

astroquery/exoplanets/planet_params.py
```diff
@@ -0,0 +1,153 @@
...
+
+import astropy.units as u
+from astropy.time import Time
+from astropy.coordinates import SkyCoord
+
+__all__ = ['PlanetParams']
+
+
+class PlanetParams(object):
+    """
```
barentsen: I struggle to understand the raison d'être of the `PlanetParams` class.  The only benefit I can see of calling `PlanetParams.from_exoplanet_archive("Trappist-1 b").pl_radj` instead of `ExoplanetArchive.query_planet("trappist-1 b")['pl_radj']` is that the former returns a Quantity and the latter a float, but this could be resolved by using a `QTable` instead of a `Table` in the implementation of `ExoplanetArchive`.

I don't see `StarParams` or `GalaxyParams` classes in astroquery, so I wonder why there should be a `PlanetParams`?  I would consider this carefully because adding this class means more code to maintain whilst risking confusing the user by providing two ways to do the same.

astroquery/exoplanet_archive/exoplanet_archive.py
bmorris3: Good idea!

astroquery/exoplanet_archive/exoplanet_archive.py
bmorris3: It used to. Removing it now.

astroquery/exoplanet_archive/exoplanet_archive.py
bmorris3: 👍

astroquery/exoplanets/planet_params.py
bmorris3: I've come around, I see it your way now. `PlanetParams` only offers cosmetic differences in the API, and it will probably more trouble than it's worth to maintain. I'm looking into returning a QTable now.

astroquery/nasa_exoplanet_archive/nasa_exoplanet_archive.py
```diff
@@ -0,0 +1,108 @@
...
+        table : `~astropy.table.QTable`
+            Table of exoplanet properties.
+        """
+        if self._table is None:
+            table_path = download_file(EXOPLANETS_CSV_URL, cache=cache,
+                                       show_progress=show_progress, timeout=120)
+            exoplanets_table = ascii.read(table_path)
+
+            # Store column of lowercase names for indexing:
+            lowercase_names = [''.join([host_name.lower().replace(' ', ''), letter])
```
barentsen: Nit-picking: you could simplify this to `host_name.lower().replace(' ', '') + letter`.

keflavich:
Thanks for the review @barentsen, and the major effort @bmorris3!  

We have a few outstanding issues:
https://travis-ci.org/astropy/astroquery/jobs/277702642#L1884 is a very weird one that I suspect is unrelated to this PR.  @bsipocz, any ideas?   This one might be the same: https://travis-ci.org/astropy/astroquery/jobs/277702641

https://travis-ci.org/astropy/astroquery/jobs/277702636 is a sphinx failure.  Again, I think it is an unrelated error.

astroquery/exoplanet_orbit_database/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+This packages contains affiliated package tests.
```
keflavich: Revise this to say "exoplanet orbit module tests"?

bmorris3:
@keflavich I've added offline tests with really small data files (one row of each table), and merged with master. Let me know if I still need to rebase, I'm not quite sure what's happening in the history of this branch.

bmorris3:
And that [pesky docs build warning](https://travis-ci.org/astropy/astroquery/jobs/282045664#L1625) is still showing up. Any tips @bsipocz?

bsipocz:
@bmorris3 - yes, please do rebase, and get rid of any modifications to astropy-helpers. I'm fairly certain that the doc build issue is due to the downgrades of the helpers.

bsipocz:
Something is still not quite right here, many unrelated commits are still brought in from master.

bmorris3:
@bsipocz Agreed. Trying to fix it now.

bsipocz:
Thanks. If you still run into issues, I'm happy to look at this in details, or have a pair rebasing on slack to figure this out, tomorrow (it's getting quite late here).

bmorris3:
@bsipocz Haha! I think I got it.

keflavich:
Looks good, though we still have 2 doc warnings:
https://travis-ci.org/astropy/astroquery/jobs/282057665#L2020

bmorris3:
@keflavich All fixed.

keflavich:
I added a tiny changelog entry for this in 69bf2c33aefbde9eb0f7807bd1956b778baa6893

bmorris3:
Thanks again for the extensive reviews, @barentsen!

PR #772, opened 2016-10-18
Corrected query_object_catalogs parameter order

imbasimba:


astroquery/esasky/core.py
```diff
@@ -161,7 +161,7 @@ def query_object_catalogs(self, position, catalogs=__ALL_STRING,
         query_object_catalogs("265.05, 69.0", ["Gaia DR1 TGA", "HSC"])
         """
         return self.query_region_catalogs(position, self.__ZERO_ARCMIN_STRING, 
-                                          row_limit, catalogs, 
+                                          catalogs, row_limit,
```
keflavich: Since this issue came up, why not pass these as keyword arguments?

imbasimba: Yup! I definitely see the benefit now :)

Btw, I am going to make a few more changes, so you don't need to merge yet.

imbasimba:
@keflavich That is the last known issue, so you can merge now. 

On an unrelated note... There are a few warnings that gets printed while I use the esasky module that I guess are part of astropy. This one for example:
WARNING: W35: None:6:2: W35: 'value' attribute required for INFO elements [astropy.io.votable.tree]

Is there anything we can do from the esasky module to not get these warnings or can they only be addressed from astropy code? Have other modules done anything to suppress warnings like this or do they appear all over?

keflavich:
Good question.  I think the warnings can be suppressed, but it might be worth investigating whether the votables can be corrected upstream.  It looks like there may be a real (but minor) bug in the votable.

keflavich:
@imbasimba Let me know if/when this is ready to merge.

imbasimba:
@keflavich It is ready.

PR #774, opened 2016-10-19
fixed broken heasarc query

nocturnalastro:
The queries where just returning the entire table because the field `object_name` should be `Entry`.

### DIFF
--- a/astroquery/heasarc/core.py
+++ b/astroquery/heasarc/core.py
@@ -2,6 +2,7 @@
 from __future__ import print_function
 from astropy.extern.six import BytesIO
 from astropy.table import Table
+from astropy.io import fits
 from ..query import BaseQuery
 from ..utils import commons
 from ..utils import async_to_sync
@@ -12,6 +13,7 @@ __all__ = ['Heasarc', 'HeasarcClass']
 
 @async_to_sync
 class HeasarcClass(BaseQuery):
+
     """HEASARC query class.
     """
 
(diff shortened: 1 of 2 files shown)


keflavich:
Could you add a remote test for this?  I'll merge it if it works, but it would be helpful to avoid future similar failures to have a test.

nocturnalastro:
Yeah sure I'll write one in the next couple of days.

nocturnalastro:
So looking at the remote test it should have failed perhaps it wasn't run?
So I checked on the web interface the correct value is 63 not 1000 for the default radius

I also noticed that when querying NuSTAR's and XMM's master tables there can sometimes be blank cells (as the obs have been accept but not done yet) this makes fits fail, as it tries to convert and empty string into  a float or an int. I tried setting TNULL but as it's just spaces it would give an empty string to fits when asked to return the value so I wrote a function to look for these empty cells and replace it with -1 where required.

keflavich:
Ah, ok, if it should have failed anyway, then that's just us not checking.  We don't run the remote tests all that often and they are not run automatically (to avoid crashing services and/or getting us marked as spammers).  

I'll check this & merge the fix when I get a chance.  Could you put your hack-to-fix-empty-cells code in a different PR?  We include tools like that fairly often because there are a lot of services that return non-compliant tables, especially VO tables, and I like astroquery to take on the burden of fixing those issues when possible.

nocturnalastro:
Yeah sure I'll sort it out this weekend :smile:

keflavich:
I've checked the fix locally and confirm that it looks right.  The HEASARC queries just were doing the wrong thing originally.  It would be great to get some local tests for this if you have the time @nocturnalastro.  In any case, thanks for this fix!

PR #779, opened 2016-10-26
Fixed bug in Integral query

imbasimba:


keflavich:
If you can add a regression test, it would be great, otherwise I'll just trust the change.  Let me know if/when it is ready to merge (after travis-ci tests pass)

imbasimba:
It is ready

PR #781, opened 2016-10-31
Heasarc: Fixed problem where lines where stripped to much

nocturnalastro:
I fixed a problem where the mini fits parser which adds -1 into blank number columns strips away if the last column is empty. This caused a problem with record not being the correct size (from what I can gather).


keflavich:
:+1:.  Again, it would be good to have some offline tests for this sort of issue.  Could you provide an example link (or example query) that resulted in this error?

nocturnalastro:
I just realized that this only happens using `query_id` (which queries based on the observation id) I wrote yesterday as `query_object` always has a filled end column. So I'll move fix into that branch and so it's introduced with that method. And I'll write some tests this weekend (hopefully) :)

PR #780, opened 2016-10-26
Bugfixes/enhancements for NRAO archive query tool

keflavich:
There were a few missing features and some incorrect parameter checks

PR #782, opened 2016-11-01
ALMA util fix: WCS indexing takes only integers

keflavich:
Fixes issues seen in #771 and #780 builds


keflavich:
@MSeifert04 this PR solved the issue in https://github.com/astropy/astroquery/pull/771#issuecomment-258023996.  I don't think anything is needed upstream - it makes sense to fail when non-integer indices are given.  That way, the user is forced to determine whether to round or truncate the floats to ints (in this case, I chose truncation).

MSeifert04:
@keflavich ok, thanks for the notification. 👍

PR #809, opened 2016-12-19
Minor splatalogue change

keflavich:
1. improved example
2. change parsing to be local again after finding out that the remote chemical parser is highly inflexible

### DIFF
--- a/astroquery/splatalogue/core.py
+++ b/astroquery/splatalogue/core.py
@@ -88,6 +88,40 @@ class SplatalogueClass(BaseQuery):
         reflags : int
             Flags to pass to `re`.
 
+        Examples
+        --------
+        >>> Splatalogue.get_species_ids('Formaldehyde')
+        {'03023 H2CO - Formaldehyde': '194',
+         '03106 H213CO - Formaldehyde': '324',
+         '03107 HDCO - Formaldehyde': '109',
+         '03108 H2C17O - Formaldehyde': '982',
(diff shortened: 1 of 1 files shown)


keflavich:
Unfortunately py2.7 is comparing unicode to unicode incorrectly.  I don't know how to fix this.  I'm inclined to just skip these tests, but if anyone else knows how to fix this bug, please let me know.

bsipocz:
@keflavich - I think it's fine to skip the tests for python 2.7, if this brakes in the future we will be able to identify it with python3.

keflavich:
Is it possible to disable a specific doctest for only py 2.7?

bsipocz:
@keflavich - I'm not sure whether the is a proper, more elegant way, but you should be able to use ``__doctest_requires__`` with specifying a python2.7 only thing to import.

keflavich:
....what the heck is going on with Windows?  It doesn't like `isclose` any more?  
https://ci.appveyor.com/project/Astropy/astroquery/build/1.0.266/job/yqbw4sk5dl57l41e#L948
I'm going to ignore these failures.

bsipocz:
The appveyor error is very weird, probably we can have a workaround by pinning some package version, I'll have more opportunity to check the details once back from the holidays. I also wonder whether other affiliated packages, or core astropy experience it?

Maybe @mwcraig saw this problem before?

Issue #839, opened 2017-02-08
feros missing in eso.list_instruments()


sousasag:
Hi,

I am trying to use astroquery to better explore the eso archive. I notice that feros is missing in the list_instruments().

From the code (astroquery.eso.core), it seems that feros is forseen, but I always get an empty table. I am wondering if it is because feros is not in the list.

I also noticed that harps if appended "manually" to the list. Should this be done also for feros?

keflavich:
@sousasag It might be necessary.  It looks like we had a fix partly implemented but did not complete it:
https://github.com/astropy/astroquery/pull/671
If you have any direct suggestions on what needs to be change, or would like to submit a pull request with the fix, we welcome the help!  Otherwise, we'll try to get to it, but I'm busy traveling at least for the next few days so I can't investigate immediately

sousasag:
Hi,
I am not really an expert with git, so it will take me some time to understand how to do a pull request, etc...

So far I got as far as:
create a new env with anaconda
clone the astroquery
and edit locally on the core.py

A solution that I found, a simple one, was to mimic what is coded for GROUND. 
I did append the feros to the list of instruments and then I added similar code for FEROS and HARPS like it was for GROUND.

```
            if instrument == 'grond':
                query_dict['instrument'] = "GROND"

            if instrument == 'harps':
                query_dict['instrument'] = "HARPS"

            if instrument == 'feros':
                query_dict['instrument'] = "FEROS"

```


And it seem it works. and of course, I also used the same url as for ground...

Maybe this can be an easy fix to do.

This seems to fix also the problem with 
eso.query_instrument('feros', help=True)
and for harps, since it is using the main query form.

keflavich:
Solved by #840.  Thanks @sousasag!

PR #840, opened 2017-02-09
query not working for feros and harps, and feros missing in instrument list

sousasag:
Following the issue (https://github.com/astropy/astroquery/issues/839)
I am creating my first ever pull request. This small modifications in the core.py file were sufficient to query harps and feros data.

I used the grond instrument logic and just replicated with the same url for harps and feros...

I am not sure if this was the main idea for the query of harps and feros. But for what I want to do, this is perfect.


keflavich:
I think this is fine.   At the very least, it is a good temporary fix until #671 does more.  As long as this solves your immediate problem, I'll go ahead and merge it.  Thanks for the contribution!

PR #847, opened 2017-03-01
Added http header for esasky requests

imbasimba:
To track statistics.

@keflavich @bsipocz 
On another unrelated note... What do you think of updating the official version soon?
The ESASky module has practically been unusable since january, when we changed some things on our servers. We updated the astroquery module to account for this some time ago and would like to get ESASky working in the official version again.


imbasimba:
Perfect, thanks!

PR #977, opened 2017-09-28
Fix a typo in esasky-core

keflavich:
Is there any way we could get tests to cover this?

### DIFF
--- a/astroquery/utils/commons.py
+++ b/astroquery/utils/commons.py
@@ -383,7 +383,7 @@ class FileContainer(object):
             estr = '' if len(e.args) < 1 else (': ' + str(e))
             warn(aud.CacheMissingWarning(msg + e.__class__.__name__ + estr))
 
-        with aud._open_shelve(urlmapfn, True) as url2hash:
+        with _open_shelve(urlmapfn, True) as url2hash:
             if str(self._target) in url2hash:
                 target = url2hash[str(self._target)]
             else:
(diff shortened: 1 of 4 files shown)


keflavich:
If this code snippet were being called anywhere in the tests, we would get a crash, but apparently that's not happening at present.

imbasimba:
That is weird. I will have time to investigate in about a month.

astroquery/utils/commons.py
```diff
@@ -453,3 +453,21 @@ def parse_votable(content):
     """
     tables = votable.parse(six.BytesIO(content), pedantic=False)
     return tables
+
+def _open_shelve(shelffn, withclosing=False):
```
pllim: So weird that you need this still with Python > 3.4. Ah, well.

astroquery/utils/commons.py
```diff
@@ -453,3 +453,22 @@ def parse_votable(content):
...
+        with _open_shelve('somefile',True) as s:
+            ...
+    """
+    import shelve
+    import contextlib
+
+    shelf = shelve.open(shelffn, protocol=2)
+
+    if withclosing:
+        return contextlib.closing(shelf)
```
pllim: At a glance, I don't see why this would behave differently than `__exit__()` in https://github.com/python/cpython/blob/master/Lib/shelve.py . Maybe it is something subtle in the implementation details. Academic discussions for another day...

PR #982, opened 2017-10-04
Login capability for astroquery.mast

ceb8:
This PR is related to the [mast query tool issue](https://github.com/astropy/astroquery/issues/896).

New functionality implemented in this PR is:

- Log in capabilities
   Allows users to log in with their MyST credentials to access proprietary data.
   Can supply username/password or be prompted.
   Can attach existing session token to a mast class instance.
   Can save username/password in keyring if so desired.

- Updated return Tables to be masked
   The returned astropy tables are now masked tables, with missing data values masked and "default values" matching what is found in the [MAST Discovery Portal](mast.stsci.edu).

- Provision for "obstype" selection (science vs calibration data) has been added in preparation for the next Portal release which adds this functionality
   This does not interfere with current functionality, but has no effect on queries until that release.
    Documentation of this option has not yet been included since the functionality is not yet live.

### DIFF
--- a/astroquery/mast/core.py
+++ b/astroquery/mast/core.py
@@ -667,8 +667,7 @@ class MastClass(QueryWithLogin):
         else:
             retrieveAll = False
 
-        # TODO: remove "Clara development" before pull request
-        headers = {"User-Agent": self._session.headers["User-Agent"] + " Clara development",
+        headers = {"User-Agent": self._session.headers["User-Agent"],
                    "Content-type": "application/x-www-form-urlencoded",
                    "Accept": "text/plain"}
 
--- a/astroquery/mast/core.py
+++ b/astroquery/mast/core.py
@@ -43,14 +43,18 @@ __all__ = ['Observations', 'ObservationsClass',
 class ResolverError(Exception):
     pass
 
+
 class SessionTokenError(Exception):
     pass
+
 
 class InputWarning(AstropyWarning):
     pass
 
+
 class AuthenticationWarning(AstropyWarning):
     pass
+
 
 class MaxResultsWarning(AstropyWarning):
     pass
@@ -94,11 +98,11 @@ def _mashup_json_to_table(json_obj, col_config=None):
     dataTable = Table(masked=True)
     absCorr = None
 
-    if not all(x in json_obj.keys() for x in ['fields','data']):
+    if not all(x in json_obj.keys() for x in ['fields', 'data']):
         raise KeyError("Missing required key(s) 'data' and/or 'fields.'")
 
     for col, atype in [(x['name'], x['type']) for x in json_obj['fields']]:
-        
+
         # Removing "_selected_" column
         if col == "_selected_":
             continue
@@ -106,29 +110,29 @@ def _mashup_json_to_table(json_obj, col_config=None):
         # reading the colum config if given
         ignoreValue = None
         if col_config:
-            colProps = col_config.get(col,{})
-            ignoreValue = colProps.get("ignoreValue",None)
+            colProps = col_config.get(col, {})
+            ignoreValue = colProps.get("ignoreValue", None)
 
         # making type adjustments
         if atype == "string":
             atype = "str"
-            ignoreValue = "" if (ignoreValue == None) else ignoreValue
+            ignoreValue = "" if (ignoreValue is None) else ignoreValue
         if atype == "boolean":
             atype = "bool"
-        if atype == "int": # int arrays do not admit Non/nan vals
-            ignoreValue = -999 if (ignoreValue == None) else ignoreValue
+        if atype == "int":  # int arrays do not admit Non/nan vals
+            ignoreValue = -999 if (ignoreValue is None) else ignoreValue
 
         # Make the column list (don't assign final type yet or there will be errors)
-        colData = np.array([x.get(col, ignoreValue) for x in json_obj['data']],dtype=object)
-        if ignoreValue != None:
-            colData[np.where(np.equal(colData,None))] = ignoreValue
+        colData = np.array([x.get(col, ignoreValue) for x in json_obj['data']], dtype=object)
+        if ignoreValue is not None:
+            colData[np.where(np.equal(colData, None))] = ignoreValue
 
         # no consistant way to make the mask because np.equal fails on ''
-        # and array == value fails with None 
-        if atype == 'str': 
+        # and array == value fails with None
+        if atype == 'str':
             colMask = (colData == ignoreValue)
         else:
-            colMask = np.equal(colData,ignoreValue)
+            colMask = np.equal(colData, ignoreValue)
 
         # add the column
         dataTable.add_column(MaskedColumn(colData.astype(atype), name=col, mask=colMask))
@@ -154,19 +158,18 @@ class MastClass(QueryWithLogin):
         self._COLUMNS_CONFIG_URL = conf.server + "/portal/Mashup/Mashup.asmx/columnsconfig"
 
         # shibbolith urls
-        self._SP_TARGET = conf.server + "/api/v0/Mashup/Login/login.html" 
+        self._SP_TARGET = conf.server + "/api/v0/Mashup/Login/login.html"
         self._IDP_ENDPOINT = conf.ssoserver + "/idp/profile/SAML2/SOAP/ECP"
-        self._SESSION_INFO_URL = conf.server + "/Shibboleth.sso/Session" 
+        self._SESSION_INFO_URL = conf.server + "/Shibboleth.sso/Session"
 
         self.TIMEOUT = conf.timeout
         self.PAGESIZE = conf.pagesize
 
         self._column_configs = {}
         self._current_service = None
-        
+
         if username or session_token:
             self.login(username, password, session_token)
-
 
     def _attach_cookie(self, session_token):
         """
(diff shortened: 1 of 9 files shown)


astroquery/mast/core.py
```diff
@@ -40,10 +44,22 @@ class ResolverError(Exception):
     pass
 
 
+class SessionTokenError(Exception):
```
pllim: I feel like such exceptions should be in some higher level module for all sub-modules to share. Just a thought. I'll let the maintainers decide.

ceb8: I could use the more generic LoginError.

bsipocz: yes, ``LoginError`` seems much better to me, thus sharing the behaviour with other modules that do authentication.

ceb8: fixed

astroquery/mast/core.py
```diff
@@ -32,7 +32,8 @@
 from ..query import QueryWithLogin
 from ..utils import commons, async_to_sync
 from ..utils.class_or_instance import class_or_instance
-from ..exceptions import TimeoutError, InvalidQueryError, NoResultsWarning, RemoteServiceError
+from ..exceptions import TimeoutError, InvalidQueryError, RemoteServiceError, LoginError
+from ..exceptions import NoResultsWarning
```
bsipocz: no need to repeat the from part, simply list everything in parenthesis with a linebreak where needed after the first import

ceb8:
@bsipocz I am getting coverage complaints because I don't have tests for all the login stuff.  What should I do about this? I don't think I can really test it automatically because I can't hardcode a password into the tests or anything. I looked at other modules with logins and none of them really seamed to test those features, so I just didn't either, but I am not sure what the optimal solution is.

bsipocz:
@ceb8 - if the MAST API doesn't provide a dummy login for testing, then it should be totally fine to leave that out assuming that locally it works when using some valid credentials.

Actually I'm not sure whether we can even mock the logins in a meaningful way. @keflavich?

ceb8:
@bsipocz  Yeah I don't really have a dummy login, and even if I did the password would change every 90 days, so it wouldn't be a sustainable solution.

I don't think there would be a meaningful way to mock the login, since the important part is  the communication with the shibboleth server which can't be mocked.

ceb8:
@bsipocz I am trying to clear up the last travis error, but I'm having trouble relating the errors to my code.  It looks like these are coming up during testing, but all my tests pass when I run locally so I am not sure how to track these errors down. Any advice?

pllim:
Have you rebased recently? @adamginsburg fixed some monkey-patching in #977.

ceb8:
@pllim I rebased this morning, and I do see those commits in my commit history.

astroquery/mast/tests/test_mast.py
```diff
@@ -49,6 +52,8 @@ def post_mockreturn(method="POST", url=None, data=None, timeout=10, **kwargs):
     else:
         service = re.search(r"service%22%3A%20%22([\w\.]*)%22", data).group(1)
 
+    print(service)
```
bsipocz: I'm fairly certain that this is not needed here

ceb8: Yupppp! Good catch, removed. :)

bsipocz:
I can confirm that the failure is upstream. Builds that passed yesterday are failing now, so will look into astropy dev now.

However in the meantime I've noticed that you've added and then removed significant amount of slightly unrelated code. Would that be possible to squash those commits together? No need to squash everything, just the big removal one into the first.

Once I find the issue, I'll try to do a review.

ceb8:
@bsipocz Commits squashed.

astroquery/mast/core.py
```diff
@@ -63,59 +77,258 @@ def _prepare_service_request_string(json_obj):
...
+
+        exp = re.findall(r'<strong>Session Expiration \(barring inactivity\):</strong> (.*?)\n', response.text)
+        if len(exp) == 0:
+            print(response.text)
+            print("Authentication failed!")
+            return False
+        else:
+            exp = exp[0]
+
+        print("Authentication successful!")
```
bsipocz: is there a way to silence the prints? Either you could use a "verbose" keyword or the logging machinery? These can be then be simple ``log.info``s and the authentication failed ones ``log.warning``s?

astroquery/mast/core.py
```diff
@@ -199,17 +435,193 @@ def _parse_result(self, responses, verbose=False):
...
+            if not password:
+                password = getpass("Enter password for {}: ".format(username))
+
+            # store password if desired
+            if store_password:
+                keyring.set_password("astroquery:mast.stsci.edu", username, password)
+
+            return self._shib_login(username, password)
+
+    def login(self, username=None, password=None, session_token=None,
```
bsipocz: why is this method needed here? to have the docstring?

ceb8: Yes, and to show up in the docs.  When I just used _login, even though the QueryWithLogin class makes a login function it didn't show up in the documentation at all, so users wouldn't know it was there, or what the args are.

bsipocz: OK, I would say then that leave this here for now, but ideally we need to address this issue in ``QueryWithLogin`` and then remove the method from here, too.

astroquery/mast/core.py
```diff
@@ -44,6 +50,14 @@ class InputWarning(AstropyWarning):
     pass
 
 
+class AuthenticationWarning(AstropyWarning):
```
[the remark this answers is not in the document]
ceb8: Done.

astroquery/mast/core.py
ceb8: Done.

astroquery/mast/core.py
```diff
@@ -63,59 +77,258 @@ def _prepare_service_request_string(json_obj):
...
+    col_config : dict, optional
+        Dictionary that defines column properties, e.g. default value.
 
     Returns
     -------
     response : `~astropy.table.Table`
     """
 
-    dataTable = Table()
+    dataTable = Table(masked=True)
```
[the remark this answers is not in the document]
ceb8: I've made sure there is no camelcase in function arguments. If removing all camelcase is required for merging let me know and I will make the changes.

bsipocz:
@keflavich - Given that you've been the one implementing the authentication for the other modules, could you have a quick look at this?

astroquery/mast/core.py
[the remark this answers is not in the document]
bsipocz: no, it's not at all a blocker, I would just like to have Adam having a look at this, this, too.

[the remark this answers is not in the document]
ceb8: OK!

docs/mast/mast.rst
```diff
@@ -293,6 +293,74 @@ Product filtering can also be appllied directly to a table of products without p
...
+
+                >>> sessionInfo = mySession.session_info()
+
+                Session Expiration: 559.0 min
+                Username: testUser@stsci.edu
+                First Name: Test
+                Last Name: User
+
+\* For security passwords should not be typed into a terminal or Jupyter notebook
+but instead input using a more secure method such as `~getpass.getpass`.
```
keflavich: So can we just use getpass in `login`?

keflavich:
Oops, I missed a large file diff!  My review is still in progress, sorry.

docs/mast/mast.rst
ceb8: Yes getpass is the default if a password isn't supplied, but if someone does want to directly supply a password that is also allowed.  I did this partly because that's what the other services with log in capabilities do, and partly because I didn't want to force user interaction for the case when someone wants to run a cron or something else automatic and would rather supply the password directly.

astroquery/mast/core.py
```diff
@@ -187,6 +388,25 @@ def _request(self, method, url, params=None, data=None, headers=None,
 
         return allResponses
 
+    def _get_col_config(self, service):
+        """
+        Gets the columnsConfig entry for given service and stores it in self._column_configs.
```
keflavich: backticks around `self._column_configs`?

docs/mast/mast.rst
keflavich: OK, good plan.  I think that should be mentioned in the docs, though

keflavich: is it and I just missed it?  Sorry, I'm reviewing with 50% attention...

ceb8: In the rst documentation under Proprietary Data Access it says:
> If a password is not supplied, the user will be prompted to enter one.

And in the docstring for the log in function it says:
> If a password is necessary but not supplied it will be prompted for.

If you think I should highlight this more let me know. I want to help people not type their passwords into the terminal as much as possible!

keflavich: No, that's what I am looking for.  Sorry, too many parallel operations here.

astroquery/mast/core.py
```diff
@@ -199,17 +419,193 @@ def _parse_result(self, responses, verbose=False):
...
+        reenter_password : bool, optional
+            Default False.
+            Asks for the password even if it is already stored in the keyring.
+            This is the way to overwrite an already stored password on the keyring.
+        """
+
+        return super(MastClass, self).login(username=username, password=password, session_token=session_token,
+                                            store_password=store_password, reenter_password=reenter_password)
+
+    def logout(self):  # pragma: no cover
```
keflavich: What does `pragma: no cover` mean?  I'm not familiar with the term.

pllim: It is to exclude that block from coverage calculation. You mark it as such when you know that code is not tested.

pllim: There is also `# pragma: py2` for tested in Python 2 only and `# pragma: py3` for Python 3 only. Might be others too but I only used these.

bsipocz: it just tells coveralls to ignore this (as it was a conscious decision that this would/could not have test coverage)

astroquery/mast/core.py
ceb8: Fixed.

astroquery/mast/core.py
bsipocz: oops, I haven't seen your reply here @pllim, would not have repeated you...

docs/mast/mast.rst
ceb8: 👍

astroquery/mast/core.py
pllim: No worries!

keflavich:
I think we're good to merge now, right?  @bsipocz, if you agree, do the honors.

bsipocz:
Indeed we can merge this. Thanks @ceb8!

PR #991, opened 2017-10-18
Fix list parameters

weaverba137:
While cleaning up some old branches I discovered a documentation error in some of the SDSS functions.  In several cases, input parameters were listed as type `float`, when in fact, looking at the code, they have to be a `list` (or `tuple` or other iterable).


weaverba137:
Ping!

PR #997, opened 2017-11-03
Remove vo.astronet.ru from default VO Cone Search list

pllim:
`vo.astronet.ru` cone search service URLs have been broken for many months and I never heard back from their maintainer.

Part of #871


pllim:
Thanks!

PR #1010, opened 2017-12-17
IRSA has changed URLs from http to https

keflavich:


pllim:
For the record, from Vandana Desai:

IRSA web services are transitioning to more secure "https" URLs. On **Wednesday, December 20**, all incoming web requests to http URLs at IRSA will be automatically redirected to https. Most browsers will have no difficulties with the change. However, scripts accessing IRSA's data access APIs may fail if they do not handle the redirect responses. For example, the popular "curl" utility does not follow such redirects automatically by default.

The recommended solution is to update references from "http" to "https" in the URLs used. All IRSA services already support the use of https, so users can proceed with updating and testing their software as soon as possible, to avoid potential disruption that might be caused if the software does not handle the http-to-https redirects smoothly.

keflavich:
The build failure is a [deprecation warning](https://travis-ci.org/astropy/astroquery/jobs/317793892#L3016) fairly deep within votable.  We should ignore this and raise an upstream issue.

keflavich:
It looks like there's already a solved upstream issue? https://github.com/astropy/astropy/issues/6850 solved in https://github.com/astropy/astropy/pull/6959

The failed build was on Tue Dec  5 20:11:19 UTC 2017, while the merge was on Tue Dec 12 17:33:54 2017, so I'll just restart the failed test.

pllim:
Yes, should have been fixed upstream by astropy/astropy#6785.

PR #1016, opened 2018-01-03
Fix units in IRSA dust extinction table

lpsinger:
The IRSA dust 'extinction by band' service, queried by `IrsaDust.get_extinction_table()`, returns wavelength units of `microns` and extinction units of `mags`, which are not recognized by `astropy.units.Unit` because of the plural suffix. Strip off the plural suffix so that the units are correctly recognized.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 0.3.7 (unreleased)
 ------------------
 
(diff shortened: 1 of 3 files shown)


keflavich:
I attempted to "Resolve conflicts" online, which failed.  There are no conflicts, yet github is reporting that there are.  I will merge this manually.

PR #1032, opened 2018-01-31
Raise exception if exceeded output table size

sarikj93:


bsipocz:
Thanks @sarikj93!

Overall it would be good to change all these generic exceptions to something more specific, but that is a separate issue.

sarikj93:
I completely agree!

Kind regards,
Sarik

PR #1043, opened 2018-02-12
Fixing some links and permanent redirects

migueldvb:
The link to the latest `astropy.vo` in index.html is broken because the service has been removed in astropy 3.0. I have removed this link and added an additional reference to the `Simple-Cone-Search-Creator` tool. There are some changes in source files in f9dbecb16e4df432549b3570864b68b9a8f7db0 but these are only to fix URL redirection so the PR could be considered as docs-only.

### DIFF
--- a/docs/splatalogue/splatalogue.rst
+++ b/docs/splatalogue/splatalogue.rst
@@ -23,7 +23,7 @@ Searching for Lines
 -------------------
 
 In the Splatalogue web interface, you select "species" of interest using the left side menu
-seen in the `query interface`_  You can access the line list:
+seen in the `query interface`_.  You can access the line list:
 
 .. code-block:: python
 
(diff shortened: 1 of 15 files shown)


docs/splatalogue/splatalogue.rst
```diff
@@ -89,7 +89,7 @@ online interface.  In principle, we can make a higher level wrapper, but it is
 not obvious what other parameters one might want to query on (whereas with
 catalogs, you almost always need a sky-position based query tool).
 
-Any feature you can change on the `Splatalogue web form <splat_b>`_ can be
+Any feature you can change on the Splatalogue `query interface`_ can be
```
keflavich: Is there a reason for this change?  I think 'web form' is more descriptive than 'query interface' in this case.

migueldvb: This link was broken so I used the named reference to http://www.cv.nrao.edu/php/splat/b.php . I will change the reference text to Splatalogue web form.

keflavich:
The changes look fine.  I think we've seen this test failure recently in another branch... it looks like an astropy-dev:py2.7 incompatibility issue

pllim:
Re: Failure -- Don't use Numpy 1.14 until 2.0.4 is out.

pllim:
> because the service has been removed in astropy 3.0

To clarify, the service was moved to this package itself as `astroquery.vo_conesearch`, so it still exists but we don't need to link to `astropy.vo` anymore. So the change is fine but I just want to make things clear.

migueldvb:
That's what I wanted to say, sorry for the incorrect wording. Since astropy 3.0 is now released even changing the docs link from latest to stable does not help, and the vo functionality is released in astroquery so I thought it was fine to remove the link.

pllim:
Yes, thanks for catching it! I should have removed it when I moved Cone Search over.

PR #1053, opened 2018-02-21
Adding Open Astronomy Catalog API Support

pcowpert:
This module is intended to bring the full functionality of the Open Astronomy Catalog (OAC) REST API to `astroquery`. The OAC API allows rapid retrieval of a variety of data products for supernovae, tidal disruption events (TDEs), and kilonovae. This includes photometric, spectroscopic, radio, and X-ray data. 

A complete description of the API can be found at the [OAC API Github Page](https://github.com/astrocatalogs/OACAPI).

This module include two basic query methods:

- `query_object`: This method can query the API based on an event name or list of event names. 

- `query_region`: This method can query the API based on a single set of coordinates, using either a `cone` or `box` search.

By default, both methods return the available metadata for all queried events. These searches can be refined using the `QUANTITY`, `ATTRIBUTE`, and `ARGUMENT` options outlined in the API documentation.

Both methods return an `Astropy` table by default. Users can also request a `JSON` compliant dictionary by using the `data_format` option. We note that some queries (e.g., retrieving multiple spectra) must be returned by the API as a `JSON` dictionary.

There are an additional three tailored modules designed to allow quick access to common data products. These include:

- `get_photometry`: Retrieve all photometry for a single event or list of events.
-  `get_single_spectrum`: Retrieve a single spectrum for a single event at a specified time in `MJD`.
- `get_spectra`:  Retrieve all available spectra for an event or list of events.

The `get_photometry` and `get_single_spectrum` methods return an `Astropy` table. The `get_spectra` method must return a `JSON` compliant dictionary.

This module was developed by myself and @guillochon

### DIFF
--- a/astroquery/oac/core.py
+++ b/astroquery/oac/core.py
@@ -1,7 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst.
 """
-OPEN ASTRONOMY CATALOG (OAC) API TOOL
--------------------------
 This module allows access to the OAC API and
 all available functionality. For more information
 see: api.astrocats.space.
@@ -80,11 +78,6 @@ class OACClass(BaseQuery):
             When set to `True` the method returns the HTTP request
             parameters as a dict. The actual HTTP request is not made.
             The default value is False.
-
-        Returns
-        -------
-        response : `requests.Response`
-            The HTTP response returned from the service.
 
         Examples
         --------
@@ -187,12 +180,6 @@ class OACClass(BaseQuery):
             When set to `True` the method returns the HTTP request
             parameters as a dict. The actual HTTP request is not made.
             The default value is False.
-
-        Returns
-        -------
-        response : `requests.Response`
-            The HTTP response returned from the service.
-            All async methods should return the raw HTTP response.
 
         Examples
         --------
@@ -343,12 +330,6 @@ class OACClass(BaseQuery):
             A complete list of commands and their usage can be found at:
             https://github.com/astrocatalogs/OACAPI. The default is None.
 
-        Returns
-        -------
-        response : `requests.Response`
-            The HTTP response returned from the service.
-            All async methods should return the raw HTTP response.
-
         Examples
         --------
         The method is used to grab a default light curve for an object:
@@ -395,12 +376,6 @@ class OACClass(BaseQuery):
             A single MJD time to query. This time does not need to be
             exact. The closest spectrum will be returned.
 
-        Returns
-        -------
-        response : `requests.Response`
-            The HTTP response returned from the service.
-            All async methods should return the raw HTTP response.
-
         Examples
         --------
         This method returns a single spectrum for an object at a selected
@@ -441,12 +416,6 @@ class OACClass(BaseQuery):
         event : str, required
             Name of the event to query. Can be a single event or a
             list of events.
-
-        Returns
-        -------
-        response : `requests.Response`
-            The HTTP response returned from the service.
-            All async methods should return the raw HTTP response.
 
         Examples
         --------
--- a/astroquery/oac/core.py
+++ b/astroquery/oac/core.py
@@ -1,5 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst.
 """
+OPEN ASTRONOMY CATALOG (OAC) API TOOL
+-------------------------------------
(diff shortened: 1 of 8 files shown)


keflavich:
@pcowpert Thanks for this contribution!  This is great, and it's wonderful to see a remote service with a real API that we can link to.

It looks like you should do a rebase against the latest master before we evaluate this.  It appears that you're using an out of date astropy-helpers, which will likely cause tests to fail.

astroquery/oac/core.py
```diff
@@ -0,0 +1,602 @@
...
+                 not coordinates.isscalar)):
+
+            request_payload['ra'] = coordinates.ra.deg
+            request_payload['dec'] = coordinates.dec.deg
+
+        else:
+            try:
+                request_payload['ra'] = coordinates[0]
+                request_payload['dec'] = coordinates[1]
+            except Exception:
```
keflavich: It would be better to try to catch only specific exception types (e.g., `IndexError`).  General exception catching can interfere with the interpreter and produce some otherwise bad results.  If the code returns an unexpected Exception type, we should just raise that Exception and deal with it as a bug

astroquery/oac/core.py
```diff
@@ -0,0 +1,602 @@
...
+        if (radius is not None and (height is not None or width is not None)):
+            raise ValueError("Please specify ONLY a radius or "
+                             "height/width pair.")
+
+        if ((radius is None) and ((height is None) or (width is None))):
+            raise ValueError("Please enter both a width and height "
+                             "for a box search.")
+
+        # Check that any values are in the proper format.
+        # Criteria/Code from ../sdss/core.py
```
keflavich: cc @bsipocz - perhaps we should update `commons.radius_to_unit` to handle this case more generally?  https://github.com/astropy/astroquery/blob/master/astroquery/utils/commons.py#L110

(@pcowpert - you don't need to do anything here, this is a note-to-selves that there is probably a more concise and general way to do what you've done that we should have implemented already...)

pcowpert:
@keflavich Thanks for the feedback! I'm not sure how my astropy_helpers got out of date. Something must have gone wrong when updating my fork. I'll work on rebasing everything. 

Once that is complete, I will address other feedback.

keflavich:
OK, great.  Also, before we merge this, we'll want a new entry in the changelog, and you should add the OAC docs to the index table of contents.

guillochon:
FYI @keflavich and @bsipocz we plan to announce the API/`astroquery` module via a RNAAS note once the module goes live. We have a draft of the note [here](https://www.overleaf.com/read/njqnsbbrxysc), if you have any comments on it let us know.

pcowpert:
Hi @keflavich and @bsipocz ,

I attempted to rebase, but it seems this did not affect the astropy_helpers version. I'm guessing the issue is that this is a dependent repo?

Do you have to do the rebase specifically against astropy_helpers? I have not worked with dependent repos before. Thanks!

keflavich:
Try doing `git submodule update` ?   I'm not entirely sure, though, submodules are always super tricky; @bsipocz might have a better answer.

pcowpert:
Okay it seems to have pulled down a new astropy_helpers. However, my fork and the astropy fork still seem to have slightly different hashes? Not sure if that is normal.

My work: `astropy_helpers @ 02a514c`
Astroquery: `astropy_helpers @ 41a6072`

pcowpert:
Okay, I had to force the submodule to update to a specific commit, but I think I am back in sync with the main astroquery repo

guillochon:
Any idea as to why the code coverage of the patch is so low @bsipocz? It appears that the import lines of `core.py` are being hit but the bodies of the functions (which are called in the tests) are not.

keflavich:
I haven't merged this because of the open question about code coverage.  Besides that, it is ready to be merged, right @pcowpert ?  If that's all that's holding it back, we can open a new Issue about code coverage questions and get this in.

pcowpert:
That's correct @keflavich . The issue of getting local tests up and running is all that's left. I know the remote tests work when run and it passes all of the build tests. I believe I have correctly added the docs to the ToC and included the update on the changelog. 

Everything else is in order.

keflavich:
OK.  Could you open that Issue?

pcowpert:
Thank you. The issue has been opened.

PR #1063, opened 2018-03-03
Remove limit in number of epochs in jplhorizons

migueldvb:
There is an issue with the Horizons cgi script that requires giving separate lines in the TLIST argument with at most 256 characters. This limitation may be fixed at some point by Horizons developers. To allow unlimited number of epochs in TLIST I have used a newline separator for the ephemerides, elements and vectors methods, that gets encoded when making the request. The epochs can also be provided as a numpy.ndarray object now.

@mommermi could you take a look at these changes?

### DIFF
--- a/astroquery/jplhorizons/core.py
+++ b/astroquery/jplhorizons/core.py
@@ -6,7 +6,6 @@ from numpy import nan as nan
 from numpy import isnan
 from numpy import ndarray
 from collections import OrderedDict
-import warnings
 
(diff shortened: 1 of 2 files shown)


mommermi:
Thanks, @migueldvb, this looks good to me! Not sure how the limitation to 15 epochs made it in here again, but thanks for fixing it! Also, I should keep an eye open on the remote tests: not sure why the line numbers changed.

PR #1066, opened 2018-03-14
Update splatalogue query to use GET requests

migueldvb:
The splatalogue form `c_export.php` is using GET requests at the moment, although `c.php`
still uses the POST method. `Splatalogue.query_lines`  uses `c_export.php` with POST method and currently returns `Please select a linelist catalogue` for every query.

For example the remote test does not work:

```
____________________________________ test_exclude_remote ______________________________________

    @remote_data
    def test_exclude_remote():
        # regression test for issue 616
        # only entry should be  "D213CO  Formaldehyde 351.96064  3.9e-06   ...."
        results = splatalogue.Splatalogue.query_lines(
            min_frequency=351.9*u.GHz,
            max_frequency=352.*u.GHz,
            chemical_name='Formaldehyde',
            exclude='none')
>       assert len(results) >= 1
E       assert 0 >= 1
E        +  where 0 = len(<Table length=0>\nPlease select a linelist catalogueSpecies,Chemical Name,Freq-MHz,Freq Err,Meas Freq-MHz,Meas Freq Err...        \n--------------------------------------------------------------------------------------------------------------)

astroquery/splatalogue/tests/test_splatalogue.py:165: AssertionError
```

Switching to GET requests everything seems to work. The data file used for testing is also updated in this PR to fix the tests.

@keflavich  do you know if this is a permanent change in the Splatalogue database?


keflavich:
I don't know if this is permanent, I'll ask Tony.  Thanks for the quick fix!

keflavich:
This was an unintentional change.  It may have been forced by some internal change in the php server, which they upgraded from 5.3 to 5.6.  We should monitor this situation closely; it's possible that it will revert to old behavior.

noobulous:
We are working on it now, I'll keep you folks posted with updates, thanks.

noobulous:
I don't think this was a consequence of the PHP upgrade, I think this was a consequence of moving the server from HTTP to HTTPS, which I think happened at around the same time. I think the server now punts all HTTP traffic to HTTPS, and in the process the body of the POST is lost.

Try switching the URL from 'http://www.cv' to 'https://www.cv' and see if that fixes it, the requests library you use should be able to handle that.

Meanwhile I'll see about carving out an exception in the redirect for splat.

48kRAM:
splatalogue is now exempt from the SSL redirect on www.cv.nrao.edu. Please do try to get HTTPS working in the near future, though.

noobulous:
Along the path of investigating this we uncovered other work that needs to happen with splatalogue, once we've done that we will likely move it to its own domain again. We'll be in touch when we get closer to that.

Cheers.

keflavich:
Thanks for the heads up @noobulous & @48kRAM.  We welcome any advance warning you can give us when changes are impending.

PR #1069, opened 2018-03-16
IRSA Dust: call over https

smonkewitz:
Note that besides a commit for the http to https change, this PR also includes a separate commit to ensure that files opened by the dust test cases are closed, thereby silencing quite a few pytest warnings.


keflavich:
This is closely related to #1010

PR #1077, opened 2018-03-26
Implemented search on new comet endpoint

dmcollom:
The MPC added, on request, an endpoint to enable searching for comets in addition to the existing asteroid search endpoint. In this PR:

* MPC module can query both asteroid endpoint and comet endpoint, and now requires a target_type parameter
* Refactor to reduce code reuse

### DIFF
--- a/astroquery/mpc/core.py
+++ b/astroquery/mpc/core.py
@@ -303,4 +303,5 @@ class MPCClass(BaseQuery):
             mpc_endpoint = mpc_endpoint + '/search_comet_orbits'
         return mpc_endpoint
 
+
 MPC = MPCClass()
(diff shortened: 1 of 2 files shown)


keflavich:
Is this ready for final review?

dmcollom:
Yep, it's been ready!

astroquery/mpc/core.py
```diff
@@ -9,39 +9,46 @@
 
 @async_to_sync
 class MPCClass(BaseQuery):
-    MPC_URL = 'http://' + conf.server + '/web_service/search_orbits'
-    # The authentication credentials for the MPC web service are publicly available and
-    # can be openly viewed on the documentation page at https://minorplanetcenter.net/web_service/
+    MPC_URL = 'http://' + conf.server + '/web_service'
```
keflavich: just to check: is it correct that this is http, not https?

keflavich:
OK, sorry about the delay.  It helps to leave a message when (you think) you're done editing; we have been getting a little overwhelmed by the large number of contributions recently.

PR #1091, opened 2018-04-07
Add Gaia to the module list [skip ci]

xuanxu:
`astroquery.gaia` was missing


xuanxu:
Thanks @bsipocz. Actually I've noticied README is out of sync with the list in the main docs landing page. I'll add all the missing modules to the README and PR, if that is ok.

PR #1099, opened 2018-04-13
Issue 1009: Added redirect (303) handling to sync jobs.

jcsegovia:
Fix for issue 1009. Added redirect handling to sync. jobs.

### DIFF
--- a/astroquery/utils/tap/core.py
+++ b/astroquery/utils/tap/core.py
@@ -24,6 +24,7 @@ from astroquery.utils.tap.xmlparser.jobSaxParser import JobSaxParser
 from astroquery.utils.tap.xmlparser.jobListSaxParser import JobListSaxParser
 from astroquery.utils.tap.xmlparser import utils
 from astroquery.utils.tap.model.filter import Filter
+import requests
 
 __all__ = ['Tap', 'TapPlus']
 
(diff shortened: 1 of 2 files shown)


astroquery/utils/tap/core.py
```diff
@@ -207,6 +207,20 @@ def launch_job(self, query, name=None, output_file=None,
...
+        # handle redirection
+        if response.status == 303:
+            # redirection
+            if verbose:
+                print("Redirection found")
+            location = self.__connHandler.find_header(
+                response.getheaders(),
+                "location")
+            if location is None:
+                raise Exception("No location found after redirection was received (303)")
```
keflavich: Could you make this a specific exception type?  Perhaps `requests.exceptions.HTTPError`

keflavich:
pycodestyle is complaining about spaces in the comments.  Would you mind fixing those?  In the test code, the commends are formatted as `#blah` but the standard we have adopted is `# blah`

jcsegovia:
@keflavich do you know why my local pep8 does not show any of the reported errors (E123/E265)...?

keflavich:
@jcsegovia maybe it has to do with some local config file?  I don't know for sure though.

Looks like we're down to just the known & unrelated CI bug.  So I will merge.

PR #1106, opened 2018-04-18
minor: pep8 fix

keflavich:
I'm making this a full PR because I want to see what travis does. I'm sure the last PR I merged had pep8 pass.


pllim:
With `flake8`, sometimes they add/remove stuff to complain about. Maybe it is the same deal with `pycodestyle`; just a change of behavior upstream since you last touched the file. :woman_shrugging:

PR #1119, opened 2018-04-27
Add strip to remove '\n' kept by readline

### LINKED ISSUE #1118: user/password not stripped
In class TapPlus, in login method, when credentials_file is given, the user and password are not stripped (readline keep '\n').

Nikko78:
Fix #1118


Nikko78:
I don't understand what the bot is asking

PR #1158, opened 2018-06-06
Virtual Observatory registry query

tomdonaldson:
This module will enable searching the Virtual Observatory registry to find cone, image, spectral and table services which can then be searched separately.

This work-in-progress has these known limitations:
* Needs more documentation.
* Needs more tests.
* Should have better error checking.
* Should probably explicitly define the keyword arguments.

This implementation also contains a kludgy work-around for VOTable string values being created as byte arrays in the astropy table.


astroquery/vo/core.py
```diff
@@ -0,0 +1,237 @@
...
+
+        tap_params = {
+            "request": "doQuery",
+            "lang": "ADQL",
+            "query": adql
+        }
+
+        response = utils.try_query(url, post_data=tap_params, timeout=self._TIMEOUT, retries=self._RETRIES)
+
+        if kwargs.get('verbose'):
```
keflavich: Better to use a direct kwarg in the __init__, i.e., `__init__(self, verbose=False)`

astroquery/vo/core.py
```diff
@@ -0,0 +1,237 @@
...
+            elif key == 'waveband':
+                waveband = val
+            elif key == 'source':
+                source = val
+            elif key == 'publisher':
+                publisher = val
+            elif key == 'order_by':
+                order_by = val
+            elif key == 'logic_string':
+                logic_string = val
```
keflavich: These can all be specified much more easily as actual keyword arguments

astroquery/vo/core.py
```diff
@@ -0,0 +1,237 @@
...
+            service_type = "simpleimageaccess"
+        elif "spectr" in service_type.lower():
+            service_type = "simplespectralaccess"
+        elif "cone" in service_type.lower():
+            service_type = "conesearch"
+        elif 'tap' in service_type or 'table' in service_type:
+            service_type = "tableaccess"
+        else:
+            print("ERROR: please give a service_type that is one of image, spectral, cone, or table")
+            return None
```
keflavich: you should raise an error instead, i.e.: `raise ValueError("service_type must be one of...`

astroquery/vo/tests/test_registry_remote.py
```diff
@@ -0,0 +1,11 @@
+from astroquery.vo import Registry
+
+# Find all SIA services from HEASARC.
+#heasarc_image_services = Registry.query(source='heasarc', service_type='image')
+def test_basic():
+    # Find all SIA services from HEASARC.
+    heasarc_image_services = Registry.query(source='heasarc',\
```
keflavich: no backslash is required here (please remove it)

astroquery/vo/utils.py
```diff
@@ -0,0 +1,258 @@
...
+    #
+    # (The reader also accepts just a string, but that seems to have two
+    # problems:  It looks for newlines to see if the string is itself a table,
+    # and we need to support unicode content.)
+    #Tracer()()
+    try:
+        file_like_content = io.BytesIO(response.content)
+    except IOError:
+        print("ERROR parsing response content? Got:\n    {}...".format(response.content[0:400]))
+        #  Should this return or raise?
```
keflavich: should raise

astroquery/vo/utils.py
```diff
@@ -0,0 +1,258 @@
...
+        raise
+    except Exception as e:
+        raise e
```
keflavich: this is redundant; just delete these two lines

astroquery/vo/utils.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # Loop through all the columns looking for the UCD
+    for key in table.columns:
+        col = table.columns[key]
+        ucdval = col.meta.get('ucd')
+        if ucdval is not None:
+            if ucd == ucdval:
+                return col
+
+    return None
```
keflavich: this line is redundant

astroquery/vo/utils.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # Loop through all the columns looking for the utype
+    for key in table.columns:
+        col = table.columns[key]
+        utypeval = col.meta.get('utype')
+        if utypeval is not None:
+            if utype == utypeval:
+                return col
+
+    return None
```
keflavich: this line is redundant

pllim:
I could have used this like 5 years ago. :wink:  Looking forward to some documentation.

astroquery/vo/utils.py
```diff
@@ -0,0 +1,258 @@
...
+                print("    Got {} results for parameters[{}]".format(len(result), j))
+                #Tracer()()
+            else:
+                print("    (Got no results for parameters[{}])".format(j))
+
+        service_results.append(result)
+    return service_results
+
+
+def try_query(url, retries=3, timeout=60, get_params=None, post_data=None, files=None):
```
keflavich: this function should be a method of the `Registry`.  Then instead of `bq._request`, you'll use `self._request`.

pllim:
I can see users querying the registry, though that is probably a small base, but I still think this should not be hidden in `utils`.

astroquery/vo/utils.py
trjaffe: This function is used in other classes Image, Cone, etc. and is not specific to Registry.

tomdonaldson:
Yes, this is intended to be user-facing.  The services returned will be the targets for actual data queries.  At this point, I intended to add more to astroquery.vo to support those follow-up queries.

Something like:
```
heasarc_image_services = Registry.query( service_type='image', publisher='heasarc')

image_results = []
for service in heasarc_image_services:
    image_results .append(Image.query_region('M82', radius="0d6m0s", service=service))

# or maybe query all services at once:
results = Image.query_region('M82', radius="0d6m0s", services= heasarc_image_services)
```

pllim:
There is already a `astroquery.vo_conesearch`. Maybe it is less confusing to rename this to `astroquery.vo_registry`. (I don't feel like moving conesearch *again*.)

tomdonaldson:
@pllim  This is an important question.  My initial thinking was to include support for all VO queries (cone search, image search, spectral search, table access (SQL-like) search) in one astroquery.vo package to make everything easy to find.

I have not fully digested all of the existing functionality, but I do know about astroquery.vo_conesearch, and astroquery.utils.tap, the PyVO package, and of course, astropy.io.votable.

One of the goals of this effort is to make VO queries more welcoming for developers, and one possible improvement there is to have a single intuitive location to find these features (such as astroquery.vo).

Where existing features work well, or close to well, then wrapping and delegating to those features makes sense.  So for example, if we were to include Tap query in astroquery.vo, it should use astroquery.utils.tap, making fixes to that if needed.

While I've heard reports from people that had trouble using PyVO, and it hasn't been heavily used/maintained (although there have been some recent contributions).  Reaching out to that developer will be important.

pllim:
I believe @funbaker is the current active maintainer of `pyvo`. https://github.com/pyvirtobs/pyvo

funbaker:
@tomdonaldson Which kind of troublereports do you mean?

tomdonaldson:
@funbaker  At this point I don't have any confirmed functionality problems.  I've only heard word-of-mouth reports of people having trouble getting pyvo to do what they wanted.  So there may be some combination of usability concerns along with maybe some actual functionality problems. 

My time so far hasn't allowed me to become more familiar with pyvo, but I look forward to doing that soon, and discussing any further concerns or plans with you.  It looks like pyvo's goals were similar to many of mine, facilitating user access to VO services.  I don't want to reinvent the wheel, but I would like to take a critical look at what's been done in the past to figure out what's best moving forward.

astroquery/vo/utils.py
tomdonaldson: I agree that this function belongs in the Registry class.  As we propose other classes that share this functionality, I think a small refactoring would be in order so that Registry and any new classes that want that function would inherit from a common superclass.

astroquery/vo/__init__.py
```diff
@@ -0,0 +1,25 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Virtual Observatory
+===============
```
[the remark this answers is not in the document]
tomdonaldson: Addressed

astroquery/vo/core.py
tomdonaldson: Addressed:  kwargs was removed in favor of explicit keywords.

astroquery/vo/core.py
tomdonaldson: Addressed: kwargs was removed in favor of explicit keywords.

astroquery/vo/core.py
tomdonaldson: Addressed

astroquery/vo/utils.py
tomdonaldson: Addressed

astroquery/vo/utils.py
tomdonaldson: Addressed

astroquery/vo/utils.py
tomdonaldson: Addressed

astroquery/vo/utils.py
tomdonaldson: Addressed

astroquery/vo/utils.py
tomdonaldson: Addressed:  The try_query function was moved into the (new) VoBase class, from which Registry now inherits.

keflavich:
@tomdonaldson you should rebase against master.  The differences are only minor ones in the changelog, but if you're trying to trigger a new CI build, that's a better way to do it right now

astroquery/vo/examples/Registry.ipynb
```diff
@@ -0,0 +1,295 @@
+{
```
bsipocz: Could you please remove this notebook (and/or rework the content into normal `rst` files? We don't have docs in astroquery in notebooks, and for tutorials, they are in another repo, https://github.com/astropy/astropy-tutorials.

tomdonaldson: Yes, good point.  Sorry about that.  Just getting back to this PR now, and hope to get it into decent shape soon.

bsipocz: no worries and no rush. I'm not sure about Adam's plans, but I won't be really available in the upcoming weeks to do any detailed review anyway.

tomdonaldson:
Next steps for WIP:

- Address code coverage issues.
- Documentation.

astroquery/vo_service_discovery/core.py
```diff
@@ -0,0 +1,245 @@
...
+                     dump_to_file=False, verbose=False):
+        """
+        Query the Virtual Observatory registry to find what values exist, and counts for each value,
+        for the 'service_type', 'waveband', or 'publisher' fields in the main query.
+
+        Parameters
+        ----------
+        field : str, required
+            One of 'service_type', 'waveband' or 'publisher'
+        minimum : int, optional, default is 1
```
stargaser: Add an explanation on the next line, like:
```
        Include only those results with at least this many counts.
```

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,183 @@
...
+The value of `'access_url'` specifies the endpoint for the service.
+
+Documentation of the service may be found in the `'reference_url'` values.
+We could use capabilities in the standard library to open the first url
+in a web browser.
+
+.. code-block:: python
+
+    >>> import webbrowser
+    >>> webbrowser.open(results['reference_url'][0])
```
stargaser: In the course of testing, I modified this to:
```suggestion
    >>> webbrowser.open(results['reference_url'][1].decode())
```
because the item is a byte-string, which  causes an exception in `web browser.open`.

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,183 @@
...
+10 mm), or radio (greater than 10 mm wavelength). Multiple bands can be specified
+with a comma-separated list.
+
+The ``source`` parameter is a string that is matched to any substring in ivoid.
+
+The ``publisher`` is the name of any publishing organization.
+
+``order_by`` can be specified to sort the result by any of the column names.
+
+The ``logic_string`` is any other string that is needed for the ADQL ``where``
```
stargaser: Delete this paragraph. `logic_string` is not in the function signature.

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,183 @@
...
+clause, and must start with the value ``' and '``.
+
+
+The ADQL query can be printed by specifying ``verbose=True``.
+
```
stargaser: Add to this the other options of `query`.

```suggestion

The results of the query can be saved to a file. Set ``dump_to_file=True`` to save the results table. The format will be ``'votable'`` by default; you can change this with the ``output_format`` option. The result will be saved to a filename using the TAP job ID; you can specify the filename with the ``output_file`` option.

By default, the query is made using the  registry TAP URL. The ``url`` option will override that value. Alternatively, the configuration can be changed to use a different URL.
```

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,183 @@
...
+    waveband short_name ...  publisher   service_type
+    -------- ---------- ... ------------ ------------
+                HEASARC ... NASA/HEASARC  tableaccess
+
+
+Counting available services
+===========================
+
+The `query_counts` method will enumerate how many services are available
+for each of the fields in the results table shown above.
```
stargaser:
```suggestion
for the ``'service_type'``, ``'waveband'`` or ``'publisher'`` fields.
```

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,183 @@
...
+                ceacapability                 56
+                      harvest                 41
+             simpletimeaccess                 31
+                  openskynode                 27
+                       search                  8
+    theoreticalspectralaccess                  5
+                     heliotap                  5
+          protospectralaccess                  4
+             simplelineaccess                  3
+
```
stargaser: Document the options.

```suggestion

The ``minimum`` option defaults to 1 and specifies the minimum counts needed for a row to be listed in the results table. `query_counts` includes the same ``verbose``, ``url`` and file output options as the ``query`` method.

```

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,183 @@
...
+We can get a list of the top 15 publishers:
+
+.. code-block:: python
+
+    >>> results = Registry.query_counts('publisher', 15, verbose=True)
```
stargaser: Recommend to specify the option `minimum`.

```suggestion
    >>> results = Registry.query_counts('publisher', minimum=15, verbose=True)
```

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,183 @@
...
+                      harvest                 41
+             simpletimeaccess                 31
+                  openskynode                 27
+                       search                  8
+    theoreticalspectralaccess                  5
+                     heliotap                  5
+          protospectralaccess                  4
+             simplelineaccess                  3
+
+We can get a list of the top 15 publishers:
```
stargaser: The code block is for all publishers with at least 15 services, not the top 15 publishers.

```suggestion
We can get a list of all publishers with at least 15 services.
```

smoh:
Sorry to chime in only after a lot of efforts have been put in already, but this does seem exactly like re-inventing the wheel. How is this different from Registry search capability of pyvo (https://pyvo.readthedocs.io/en/latest/registry/index.html)?

>There are two other packages with complimentary functionality as Astroquery: pyvo is an Astropy affiliated package, and Simple-Cone-Search-Creator to generate a cone search service complying with the IVOA standard. They are more oriented to general virtual observatory discovery and queries, whereas Astroquery has web service specific interfaces.

tomdonaldson:
@smoh Thanks for the question.  I generally try to avoid reinventing the wheel, and I've thought quite a bit about this question.  The short answer is that I see value in exposing this search in Astroquery, and that there wasn't enough logic overlap to warrant wrapping, depending on, or changing the existing pyvo Registry capability.

Really the most important difference between this and the pyvo feature is that this is in Astroquery.  Astroquery has become the place where people look when they want to search an astronomy resource, so just having the feature here makes it more accessible to the community.

On the technical side, there are some differences as well, again mostly geared towards lowering the bar of entry for users.  While both implementations translate input constraints into a Registry TAP query, the Astroquery implementation *tries* to hide the user a bit more from VO jargon (e.g., SIA, SSA, TAP, ).  Indeed the name of the module, 'vo_service_discovery', is hopefully more descriptive to a user than 'registry search'.  Along the same lines, the Astroquery implementation returns the results in Astropy Table objects, which will require no learning curve for Astroquery users.

tomdonaldson:
@keflavich and @bsipocz, I believe this has transitioned from a WIP to being ready for real review.  Thanks for any and all feedback!

smoh:
>On the technical side, there are some differences as well, again mostly geared towards lowering the bar of entry for users. While both implementations translate input constraints into a Registry TAP query, the Astroquery implementation tries to hide the user a bit more from VO jargon (e.g., SIA, SSA, TAP, ). Indeed the name of the module, 'vo_service_discovery', is hopefully more descriptive to a user than 'registry search'. Along the same lines, the Astroquery implementation returns the results in Astropy Table objects, which will require no learning curve for Astroquery users.

Thanks for your reply. I do see potential advantages of some of the things mentioned but I'd argue these are very _non_-technical changes that can be easily addressed by wrapping.

tomdonaldson:
I don't disagree that wrapping would be fairly easy, but it would also have drawbacks that, to me, make the wrapping not worthwhile:

- Most of both implementations is just wrapping, i.e. deciding the exact input keywords and output formats.  The core functionality (of constructing a TAP query from constraints) is a fairly small section of code.  Having Astroquery wrap on top on pyvo's wrapping would be a little more complex than the base logic itself.

- The Astroquery module would be dependent on pyvo, which is fine, but enough of a potential maintenance complexity to be worth considering.

- The pyvo version doesn't support the exact same constraints, so would need small modifications to match the desired inputs of the Astroquery module.

- I should've mentioned before that this module has a new `query_counts()` feature that can help a user learn to construct the main query.

I suppose, but don't know for sure, that a similar thought process was used when implementing the astroquery.utils.tap module which does not depend on pyvo's tap capabilities.

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,202 @@
...
+
+The ``service_type`` parameter can be set to the VO service types ``'conesearch'``,
+``'simpleimageaccess'``, ``'simplespectralaccess'``, or ``'tableaccess'``. These
+values may be shortened to ``'cone'``, ``'image'``, ``'spectra'`` or
+``'spectrum'``, or ``'tap'`` or ``'table'``.
+
+The value of the ``keyword`` parameter will return results containg that string
+in the ``ivoid``, ``short_name``, or ``res_description`` fields.
+
+Wavelength range can be specified by a string containing one or more band names.
```
keflavich: this paragraph doesn't define the keyword needed here; presumably it's `wavelength`?

keflavich: no, it's `waveband`.

keflavich: also, maybe combine the last sentence with the first?

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,202 @@
...
+The value of ``'access_url'`` specifies the endpoint for the service.
+
+Documentation of the service may be found in the ``'reference_url'`` values.
+You can use a function from the Python standard library to open the second url
+in a web browser.
+
+.. code-block:: python
+
+    >>> import webbrowser
+    >>> webbrowser.open(results['reference_url'][1].decode())
```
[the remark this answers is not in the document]
theresadower: This is basically a safety measure as not all URLs across all registries are stored decoded. While it would be best if the user did not have to decode any results, there is little explicit documentation within the IVOA standards about whether all the URLs should be escaped or not, and some inconsistency remains among the ~20k current resources.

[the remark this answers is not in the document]
pllim: IMHO such technicalities should be hidden from users if possible. But if it is impossible, perhaps an explanation in the doc on why decode is needed here?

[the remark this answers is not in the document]
tomdonaldson: The commit comment for the decode() says it's needed because the reference_url value is a byte-string.  I think that's an issue with the VOTable parser in Astropy, but a little more looking is needed.  

If the reference_url were a str, it may not need decoding regardless of whether the value were already URL-encoded.

https://github.com/astropy/astroquery/pull/1158#discussion_r240796583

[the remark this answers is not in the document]
theresadower: Apologies, I should have looked closer. There have been other places where we safety-decode URLs in strings from the registry and hide it from the user, I think Tom's right that this is a separate thing.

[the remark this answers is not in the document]
pllim: I won't be surprised if this stemmed from `io.votable`, as that package needs some maintenance. 😅

astroquery/vo_service_discovery/core.py
tomdonaldson: Addressed.

docs/vo_service_discovery/vo_service_discovery.rst
tomdonaldson: Addressed.

docs/vo_service_discovery/vo_service_discovery.rst
tomdonaldson: Addressed.

docs/vo_service_discovery/vo_service_discovery.rst
tomdonaldson: Addressed.

docs/vo_service_discovery/vo_service_discovery.rst
tomdonaldson: Addressed.

docs/vo_service_discovery/vo_service_discovery.rst
tomdonaldson: Addressed.

docs/vo_service_discovery/vo_service_discovery.rst
tomdonaldson: I reworded the waveband section to try to address the comments.

Regarding the question, VO registries do not currently have a working ability to query for services that have overlap with some part of the sky.  

However, the IVOA is working on enhancements to allow services to describe their sky coverage and to support queries based on that.  Those features would likely be based on Multi-Order Coverage maps, much like the CDS MOC Server does today (https://astroquery.readthedocs.io/en/latest/api/astroquery.cds.CdsClass.html#astroquery.cds.CdsClass.query_region)

tomdonaldson:
@keflavich @bsipocz  I believe that all the requested changes have either been addressed or are no longer relevant.  Please take a look when you have a chance, and let me know if you have additional feedback.  Thanks!

astroquery/vo_service_discovery/core.py
```diff
@@ -0,0 +1,250 @@
...
+            field = 'role_name'
+            query_where_filter = ' where base_role = \'publisher\' '
+        elif field.lower() == 'service_type':
+            field_table = 'rr.capability'
+            field = 'cap_type'
+
+        if field_table is None:
+            return None
+        else:
+            query_select = 'select ' + field + ' as ' + field_alias + ', count(' + field + ') as count_' + field_alias
```
bsipocz: this may be easier to read if `.format()` is used rather than string additions

astroquery/vo_service_discovery/tests/shared_registry.py
```diff
@@ -0,0 +1,169 @@
...
+class Helpers(object):
+    ""
+
+    DATA_FILES = {}
+
+    def table2ecsv(self, current, filepath):
+        """Dump a table and its meta data to ECSV"""
+        from astropy.io import ascii as ap_ascii
+        try:
+            ap_ascii.write(current, filepath, format='ecsv', overwrite=True)
```
bsipocz: I don't think we need to import `io.ascii` here, the Table writer should be able to deal with it and it's already been imported.

astroquery/vo_service_discovery/tests/shared_registry.py
```diff
@@ -0,0 +1,169 @@
...
+        except Exception as e:
```
bsipocz: why the try/expect if the exception is raised anyway? There are a few other similar cases below, too.

astroquery/vo_service_discovery/tests/shared_registry.py
```diff
@@ -0,0 +1,169 @@
...
+            raise e
+
+    def ecsv2table(self, filepath):
+        """
+        """
+        try:
+            with open(filepath, 'r') as f:
+                table = Table.read(f.read(), format='ascii.ecsv')
```
bsipocz: `Table` should be able to deal with a file name, doesn't need an open file

astroquery/vo_service_discovery/tests/test_registry.py
```diff
@@ -0,0 +1,284 @@
...
+import pytest
+from ...utils.tap.conn.tests.DummyConnHandler import DummyConnHandler
+from ...utils.tap.conn.tests.DummyResponse import DummyResponse
+from ...utils.tap.core import Tap
+
+from ... import vo_service_discovery
+from astroquery.vo_service_discovery import Registry
+from .shared_registry import SharedRegistryTests
+
+# To run just this test,
```
bsipocz: remove these comments

bsipocz:
@tomdonaldson - I have some minor comments, mostly related to make it easier to read the code. Otherwise this looks good and ready to go.

astroquery/vo_service_discovery/core.py
tomdonaldson: I reworked this section using format(), and I agree it looks a bit better.

astroquery/vo_service_discovery/tests/shared_registry.py
tomdonaldson: Good catch.  I removed the import and used the instance write() method for the incoming Table.

astroquery/vo_service_discovery/tests/shared_registry.py
tomdonaldson: No idea why those were there.  Thanks for pointing them out.

astroquery/vo_service_discovery/tests/shared_registry.py
tomdonaldson: Yes, that looks much cleaner.  Thanks for the review!

astroquery/vo_service_discovery/tests/test_registry.py
tomdonaldson: Done.

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,202 @@
...
+at http://vao.stsci.edu/keyword-search/. This module provides methods
+to query this directory by service type, keywords, wavelength band or coverage,
+and source or publisher.
+
+Here is an example showing how to find information about conesearch services
+provided by the NASA Extragalactic Database (NED).
+
+.. code-block:: python
+
+    >>> from astroquery.vo_service_discovery import Registry
```
keflavich: One big-picture suggestion: can we either rename the module `VORegistry` or suggest a user standard of doing that?   e.g., `from astroquery.vo_service_discovery import Registry as VORegistry`?  I don't like the idea of user code existing in the wild with the generic name `Registry`, since that could be a registry for anything.

docs/vo_service_discovery/vo_service_discovery.rst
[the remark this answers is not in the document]
keflavich: Where did we end up coming down on this?  I really don't like the idea of telling users to use `decode`.  I'd slightly prefer a hack-fix to the votable parser within this module over documentation telling users to change string types.

docs/vo_service_discovery/vo_service_discovery.rst
```diff
@@ -0,0 +1,202 @@
...
+The `~astroquery.vo_service_discovery.RegistryClass.query` method provides several
+ways to search for specific services of interest. Each of these options
+can be combined into a single call to the method.
+
+The ``service_type`` parameter can be set to the VO service types ``'conesearch'``,
+``'simpleimageaccess'``, ``'simplespectralaccess'``, or ``'tableaccess'``. These
+values may be shortened to ``'cone'``, ``'image'``, ``'spectra'`` or
+``'spectrum'``, or ``'tap'`` or ``'table'``.
+
+The value of the ``keyword`` parameter will return results containg that string
```
keflavich:
```suggestion
The value of the ``keyword`` parameter will return results containing that string
```

pllim:
I use this to un-escape the URLs in cone search:

https://github.com/astropy/astropy/blob/987be7297fed38e9fba0047419fe54cbe5f0d709/astropy/utils/xml/unescaper.py#L20

tomdonaldson:
Withdrawing this pull request.  These features will instead be integrated with similar features in PyVO (https://github.com/astropy/pyvo).

pllim:
@tomdonaldson , where is it in `pyvo`?

trjaffe:
> @tomdonaldson , where is it in `pyvo`?

pyvo.registry.search()

tomdonaldson:
> > @tomdonaldson , where is it in `pyvo`?
> 
> pyvo.registry.search()

Yes, and also the shortcut pyvo.regsearch()

PR #1161, opened 2018-06-07
GAIA: "Fix" units with .'s and quotes in them

keflavich:
This is a hack to fix some units, like `mas.yr**-1` and `'electron'.s**-1`.  It doesn't work for all the GAIA units, but it works for many.

cc @eteq, @tomdonaldson, @jcsegovia


keflavich:
I'm ignoring the appveyor failure because it was a dependency retrieval failure, but I can't restart appveyor (or, don't know how to...)

PR #1184, opened 2018-07-06
Retrieve calibrations associated to ESO data

saimn:


astroquery/eso/core.py
```diff
@@ -591,6 +593,13 @@ def retrieve_data(self, datasets, continuation=False, destination=None):
         datasets_to_download = []
         files = []
 
+        calib_options = {'none': '', 'raw': 'CalSelectorRaw2Raw',
+                         'processed': 'CalSelectorRaw2Master'}
```
saimn: There are two possibilities for calibrations, raw and processed (though I never managed to get master calib with the second option). So I added an option with three possible values.

astroquery/eso/core.py
```diff
@@ -695,16 +707,52 @@ def retrieve_data(self, datasets, continuation=False, destination=None):
...
+                try:
+                    link = root.select('a[href$=/script]')[0]
+                    if 'downloadRequest' not in link.text:
+                        # Make sure that we found the correct link
+                        raise Exception
+                    href = link.attrs['href']
+                    script = self._request("GET", href, cache=False)
+                    fileLinks = re.findall(
+                        r'"(https://dataportal.eso.org/dataPortal/api/requests/.*)"',
+                        script.content.decode('utf8'))
```
saimn: As explained above, when requested files with calibrations the files path are not accessible without javascript. So instead I get the download script, which contains the de-duplicated file list.

astroquery/eso/core.py
```diff
@@ -695,16 +707,52 @@ def retrieve_data(self, datasets, continuation=False, destination=None):
...
+                    # links with api/ do not work, wtf ???
+                    fileLinks = [
+                        f.replace('https://dataportal.eso.org/dataPortal/api/requests',
+                                  'https://dataportal.eso.org/dataPortal/requests')
```
saimn: This, I don't understand.. the script contains links with "api/", which should work obviously (and works when executing the script), but doesn't work as a direct http request. So I remove "api/" and get the same urls as with the web interface.

saimn:
One thing that is not ideal currently: files already downloaded are checked before getting the full list of files, so it will not skip the calibration files. Are you ok with moving this step just before the download, @keflavich ?

astroquery/eso/core.py
```diff
@@ -695,16 +729,63 @@ def retrieve_data(self, datasets, continuation=False, destination=None):
...
+                    if 'downloadRequest' not in link.text:
+                        # Make sure that we found the correct link
+                        raise Exception
+                    href = link.attrs['href']
+                    script = self._request("GET", href, cache=False)
+                    fileLinks = re.findall(
+                        r'"(https://dataportal.eso.org/dataPortal/api/requests/.*)"',
+                        script.content.decode('utf8'))
+
+                    # links with api/ do not work, wtf ???
```
keflavich: ...really?!  wtf indeed.  Maybe report this upstream?

astroquery/eso/core.py
```diff
@@ -695,16 +729,63 @@ def retrieve_data(self, datasets, continuation=False, destination=None):
...
+                    log.info("Detecting already downloaded datasets, "
+                             "including calibrations...")
+                    fileIds = [f.rsplit('/', maxsplit=1)[1] for f in fileLinks]
+                    filteredIds, files = self._check_existing_files(
+                        fileIds, continuation=continuation,
+                        destination=destination)
+
+                    fileLinks = [f for f, fileId in zip(fileLinks, fileIds)
+                                 if fileId in filteredIds]
+                except Exception:
```
keflavich: Also best to catch a specific exception class here

keflavich: generic Exception catching will also catch a poorly-timed ctrl-c

keflavich:
The change-of-order for checking is fine.

astroquery/eso/core.py
saimn: No, I just didn't know about `.text`, will fix.

astroquery/eso/core.py
```diff
@@ -695,16 +729,63 @@ def retrieve_data(self, datasets, continuation=False, destination=None):
...
+                # directly. So instead we retrieve the download script provided
+                # in the web page, and use it to extract the list of files.
+                # The benefit of this is also that in the download script the
+                # list of files is de-duplicated, whereas on the web page the
+                # calibration files would be duplicated for each exposure.
+                try:
+                    link = root.select('a[href$=/script]')[0]
+                    if 'downloadRequest' not in link.text:
+                        # Make sure that we found the correct link
+                        raise Exception
```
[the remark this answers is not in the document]
saimn: This exception is catched by the except below. At the beginning I added this try/except block to fallback on the other method (without calibration) is something goes wrong. But now that this seems to work, I can probably remove the try/except (and use a better exception here) ?

astroquery/eso/core.py
saimn: Yes, that's weird, without replacing I get an error:

> HTTPError: 500 Server Error: Internal Server Error for url: https://dataportal.eso.org/dataPortal/api/requests/...

There must be something done differently in the download script, but it is not obvious what. 
Do you know where to report this?

astroquery/eso/core.py
saimn: This was really for development and because I was not sure if the solution was going to work. I can remove this I guess.
Btw, `Exception` does not catch keyboard interrupt, it's only bare except and `BaseException` who do.

saimn:
@keflavich - Actually I duplicated the local files check, because it is useful both before making the request to the archive, to speed up things, and after the request, to filter calibration files.

astroquery/eso/core.py
keflavich: Maybe just the ESO helpdesk?  They'll direct you to the right person

astroquery/eso/core.py
keflavich: Oh, ok, I didn't realize that!  Thanks.

astroquery/eso/core.py
saimn: Actually, thinking more about this, this is just because the authentication is managed differently with the two urls: with `https://dataportal.eso.org/dataPortal/api` you need to use Basic Authentication (which the script does), though with `https://dataportal.eso.org/dataPortal/` you can reuse the request session with the cookie based (?) authentication.
So I can make it work with the `api/` urls if I use something like this:
 ```
+            from requests.auth import HTTPBasicAuth
+            auth = HTTPBasicAuth(self.USERNAME,
+                                 keyring.get_password("astroquery:www.eso.org",
+                                                      self.USERNAME))
+
             log.debug("Files:\n{}".format('\n'.join(fileLinks)))
             for fileLink in fileLinks:
                 fileId = fileLink.rsplit('/', maxsplit=1)[1]
                 log.info("Downloading file {0}...".format(fileId))
                 filename = self._request("GET", fileLink, save=True,
-                                         continuation=True)
+                                         continuation=True, auth=auth)
```
But it's better to modify the urls and reuse the request session (the above supposes that the username/password is stored).

saimn:
@keflavich - about unit tests, I had a look at the existing tests but it is not clear to me how to add a new one. 
The MIDI and AMBER exposure currently used do not seem to have associated calibrations.

keflavich:
@saimn Could you write up a unit test for some data set that *does* have associated calibrations?  Do this in another PR; I'll merge this here.

saimn:
Any idea of an instrument for which calibrations are not too big ? I was going to add a MUSE exposure for the test but realized that downloading several Gb of calibrations just for the test is not ideal :grin:

keflavich:
yikes... no, but I would guess a simple slit spectrograph might be the simplest?

saimn:
Okay, out of curiosity, let's do some comparison, for one exposure:
- MUSE, 32Gb :)
- XSHOOTER, 810Mb
- UVES, 291Mb
- FORS2, 90Mb, starts to be more reasonable :slightly_smiling_face:

Issue #1194, opened 2018-07-12
MAST  API changes causing catalog access problems


ceb8:
The newest MAST Portal release is causing problems when accessing HSC and Gaia through Astroquery.

I am working on a fix for this which will be in shortly, but wanted to get it on the record as a known issue.

PR #1203, opened 2018-07-16
Update to fix catalog breaking changes

ceb8:
This PR solves issue #1194 

Updated the HSC and Gaia service names.

Also added preliminary date handling. This is currently just casting to string, which is suboptimal, but prevents datetimes from causing exceptions.


ceb8:
Thanks @bsipocz!

PR #1236, opened 2018-09-17
Splatalogue utils update

keflavich:
Splatalogue has changed several column names, so I've updated the colname dictionaries.


keflavich:
I've had to restart >3 individual builds, which have (so far) all subsequently gone green.  Just a heads up, @bsipocz, that there is a bunch of flakiness across the build systems today.

keflavich:
This is minor and urgent enough that I'm merging myself.

Note that I've added some redundant keywords in the header renaming.  I left these in place because I suspect they will revert the change that made this PR necessary, and I'd rather not have to run another set of corrections when that happens (or at least, minimize those corrections...)

PR #1241, opened 2018-09-19
Fix Fermi queries by switching to https?

keflavich:
We really need a test!

This is trying to address #1211


keflavich:
@cdeil @bsipocz are either of you able to review this?

keflavich:
We're not using wget, just searching for the string

PR #1255, opened 2018-10-12
Clean get/set and improve readability of Gaia tap modules

smoh:
This is related to #964. It does not affect any functionality but makes the code easier to read. Since variables are not really private even those starting with `__`, it does not seem necessary nor to have `__internal_init`. There are still some classes left with these. I have checked that all existing related tests in utils and gaia works minus those that test get/set and now seemed unnecessary.


jcsegovia:
@bsipocz I have a question about underscores: following PEP8, naming conventions (https://www.python.org/dev/peps/pep-0008/#naming-conventions), I understand that a double leading underscore should be used for classes attributes. Do we have to follow that convention or not?

keflavich:
@jcsegovia In general, they should be avoided: ["Generally, double leading underscores should be used only to avoid name conflicts with attributes in classes designed to be subclassed."](https://www.python.org/dev/peps/pep-0008/#id47)

jcsegovia:
@eteq @keflavich  I understand that in python there is nothing really private. But, for an object oriented programming (OOP) point of view, a private attribute is a concept, it means there is no need to access it directly. It is not related to Java, it is a concept. When you design a class (in OOP), the normal approach is make everything private. If someone needs to access to them, create a getter. This is also specified in section 'Designing for Inheritance', it reads 'Always decide whether a class's methods and instance variables (collectively: "attributes") should be public or non-public. If in doubt, choose non-public; it's easier to make it public later than to make a public attribute non-public.'

If in Astroquery the recommendation is to not use double underscore, it is fine to me. It is just that it is not clear. Even if PEP8, section, 'Method names and Instance variables' recommends to avoid its usage, you can read that 'there is some controversy about the use of __names'. And later, it specifies that for private attributes (this is a concept, even if it is not really private in python), you should use double leading underscores.

Again, if in Astroquery the recommendation is to avoid them, it is fine to me.

smoh:
I think the normal approach in designing python class is, in fact, make everything public as others have mentioned already. From my experience, this is [not confined to the astropy community](https://docs.python.org/3/tutorial/classes.html). As far as I know, `__name` (in python) is specifically designed for [name mangling for inheritance](https://docs.python.org/3/tutorial/classes.html#private-variables), which probably is why the PEP statement you quoted is under 'Designing for Inheritance'.

All of the classes act pretty much as a dictionary or namespace to hold variables related to e.g., `Job`.  They do not seem to require the name mangling for inheritance. As a user I find that I frequently want to examine the variables these classes hold (and in fact, it would be very nice if __repr__ can be more useful). With `__name`, accessing them as e.g., `_Job__name` is very cumbersome and that's only after one figures out how they are defined.

If, at later time, an attribute actually really needs to have get/set methods, I think that's what the built-in [property](https://docs.python.org/3/library/functions.html#property) is for.

Because the code style is so far from the python conventions, I find that it works against others trying to understand and contribute.

Many thanks for the review; I am happy to work on any further suggestions.

p.s. Reading PEP, I think `_async` should be `async_` so I will change that if there is no objection.

keflavich:
@jcsegovia I haven't thought about the theory on this for a long while, but in most of the scientific python stack, at least, most attributes default to public over private.  Whether or not it's defined somewhere, it's the de facto standard.  So, yes, we recommend avoiding `__`s most of the time

In the code being modified here, the get/set attributes aren't doing anything useful, they just end up adding a lot of lines of code, which increases the maintenance cost with no particular gain.  If there were some sort of validation happening (i.e., checking that the value set is correct), using `@property`s would be the way to go.

Anyway, thanks @jcsegovia for making this all work in the first place, and thanks @smoh for cleaning it up a bit!  We appreciate both your contributions!

astroquery/utils/tap/model/tapcolumn.py
```diff
@@ -24,176 +24,22 @@ def __init__(self):
...
-
-    def set_flag(self, flag):
-        """Sets the TAP column flag (TAP+)
-
-        Parameters
-        ----------
-        description : str, mandatory
-            TAP column flag
-        """
-        self.__flag = flag
```
keflavich: https://travis-ci.org/astropy/astroquery/jobs/440577014

smoh:
Done; I am done with my changes. Thanks for your help!

PR #1262, opened 2018-10-23
 New module for accessing the HSC-SSP archive 

ruizca:
This adds a new service for accessing data from the Hyper Suprime-Cam Subaru Strategic Program. 
Examples of use are included in the documentation (docs/hsc/hsc.rst). 

Note that you need a valid account in the HSC archive for using this module (see https://hsc-release.mtk.nao.ac.jp/datasearch/new_user/new). Due to this fact, I'm quite lost on how to implement remote tests. Any comments or suggestions about this (or any other issue you find) are most welcome.


keflavich:
Yikes, that's not a very friendly way to implement an archive.  Perhaps we can request that an astroquery account be made so we can run some tests?

For now, you can have remote tests that work only with a valid login.  These tests would have to be skipped by the automated testing system, but you can still write them, and just have step 1 of running the tests be logging in.  i.e., create a test class that, on initialization, logs you in (requesting username & password), and just mark the entire test suite as `skip`.  But, let's also talk to the archive and see if they'll accommodate us.

ruizca:
Yeah, neither the archive or the api are very friendly :-/ I've added the remote test using the approach you suggested. 

Should I contact with the HSC people?

keflavich:
Yes, I think so.  CC me when you contact them, if you like.

Could you clarify a little more - is this the archive for a single survey, or for the whole Subaru archive?  *Is* there a subaru archive?  We'd like to provide access to both if possible.

@bsipocz it looks like this archive uses CAS under the hood.  Maybe we can find a place to put some CAS tools...?

ruizca:
This is for a single survey (details [here](https://hsc.mtk.nao.ac.jp/ssp/)). There is an archive for Subaru, and other japanese telescopes (https://smoka.nao.ac.jp). I'm not familiar with that archive, but it seems that you also need an account for retrieving data, but not for searching.

ruizca:
> Did you ever get a response from the suprime folks?

I'm afraid not. I wrote them again this week, but still no answer :-/

keflavich:
OK, thanks.  Keep me in the loop.  If we don't hear anything further by the end of next week (or... maybe by new year, since we're getting into vacation time for a lot of folks), let's talk again about how to proceed.

ruizca:
Well, I finally got an answer. This is what they said:

> If the test is passed with non-anonymous, new registered test account, it might be better for us,
> but, we are now discussing about anonymous account in HSC public release site for the test.

I guess we can wait after vacation time to see if they implement the anonymous access.

keflavich:
I'm just posting to note that we're waiting on the archive to reply.  I don't think we can have this tool in astroquery proper if they don't provide some general-user access.

That said, I think we also need to consider the case of tools that are only usable by a subset of the community.  Maybe we can have modules like this, and perhaps similar things like tools that specifically query the unreleased part of a survey, and put them in `astroquery.nonpublic` or something?  cc @bsipocz on the design/layout discussion

ruizca:
I've been talking with the archive people. Their main concern about opening a testing account for us is that it could be used as a "free account without registration". If there is a way of implementing the access to the archive without distributing the credentials (user and password) in the public code, they are willing to open the account. But I'm not sure if that is possible. Any thoughts on that?

keflavich:
Yes, we can have a "private" test account shared only between the maintainers (myself and @bsipocz) and you.  Then, we'd only run the tests locally.  It *may* be possible for us to put credentials on the travis-ci servers such that the tests could automatically be run by those servers but not shared with anyone else, but I'm not certain of that

PR #1292, opened 2018-11-26
pip-version identifying tool

keflavich:
cc @bsipocz

This is a tool to indentify the latest pip version of astroquery.  It probably belongs somewhere else, and it might be the wrong approach, but I'm submitting it as a PR to prompt discussion

### DIFF
--- a/version_deployment.py
+++ b/version_deployment.py
@@ -1,0 +1,26 @@
+import subprocess
+import sys
+from distutils import LooseVersion
+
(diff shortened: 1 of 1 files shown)


keflavich:
See also https://github.com/astropy/astropy-helpers/issues/398

keflavich:
I'm going to close this for now, as it isn't something we obviously need.

PR #1296, opened 2018-11-29
Added selcols keyword to astroquery.irsa

jgagneastro:
Added selcols keyword to astroquery.irsa allowing to select subsets or extra columns from IRSA tables

### DIFF
--- a/astroquery/irsa/core.py
+++ b/astroquery/irsa/core.py
@@ -114,7 +114,7 @@ class IrsaClass(BaseQuery):
 
     def query_region(self, coordinates=None, catalog=None, spatial='Cone',
                      radius=10 * u.arcsec, width=None, polygon=None,
-                     get_query_payload=False, verbose=False,selcols=None):
+                     get_query_payload=False, verbose=False, selcols=None):
         """
         This function can be used to perform either cone, box, polygon or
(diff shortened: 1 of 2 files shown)


keflavich:
This mostly looks fine.  Could you add something to the docstring to explain the format of `selcols`?  e.g., include an example of what columns could be selected.  It would also be helpful to link to IRSA's documentation on this parameter.  Is the selcols argument specified somewhere in their docs?

For our reference, this is a fix for Issue #766

jgagneastro:
By docstring do you mean the comment header in core.py ? If so, I’ve already done both of these things. Otherwise, I can add this info if you point me where that docstring is located

keflavich:
Yes, that's what I mean, at least in part.  It would also be helpful to add something to the prose documentation:
https://github.com/astropy/astroquery/blob/master/docs/irsa/irsa.rst

However, as far as I can tell, you have not made any changes to the docstring (the commend section below the function definition) for any functions in `irsa/core.py` in this pull request.  Check that you committed the change?

jgagneastro:
Oh wow, indeed I lost them somewhere along the way. I’ll fix this asap

jgagneastro:
I just added back the header comments about selcols

PR #1307, opened 2018-12-07
Add "_testrun" to user-agent when running tests

keflavich:
This is to help archives separate out tests run by, e.g., our cronjobs from real user queries

### DIFF
--- a/astroquery/conftest.py
+++ b/astroquery/conftest.py
@@ -40,6 +40,9 @@ if int(astropy.__version__[0]) > 1:
 # using Astropy's
 try:
     from .version import version
+    # add '_testrun' to the version name so that the user-agent indicates that
+    # it's being run in a test
+    version += '_testrun'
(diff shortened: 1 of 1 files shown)


astroquery/conftest.py
```diff
@@ -40,6 +40,9 @@
 # using Astropy's
 try:
     from .version import version
+    # add '_testrun' to the version name so that the user-agent indicates that
+    # it's being run in a test
+    version += '_testrun'
```
pllim: Instead of hacking version, maybe define a new var like `IS_RUNNING_TEST`?

keflavich: this approach is very simple, though, since no other changes are required at all (version is already propagated through the user-agent).  I'm not sure there's any real motivation to make any other change, since what we're trying to communicate is a particular variant of astroquery tests being called.

PR #1315, opened 2018-12-13
Detect ESO logout and authenticate again to retrieve data

### LINKED ISSUE #1272: Auto-logout issue with ESO retrieve_data
The ESO portal automatically logs users out after 2hrs. When downloading large datasets using the EsoClass.retrieve_data function in https://github.com/astropy/astroquery/blob/master/astroquery/eso/core.py , this results in all remaining files being replaced by a html file of the ESO login page.

There are a few potential solutions to this. For example, the function could check whether the downloaded files are as expected (i.e. not a html file), or check the authentication status before downloading each file.

saimn:
Should fix #1272. It works locally if I delete the cookies before doing doing the request.

This will work only if the password is stored in a keyring, and raise an error otherwise. I wrapped `_download_file` to detect if we get an html response, which I then interpret as login page as I don't think that we are supposed to get html here otherwise, but maybe we need a better way to detect that it is really a login page. We could look at the content, or at the number of redirections.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 0.3.10 (unreleased)
 -------------------
 
+- ESO: Try to re-authenticate when logged out from the ESO server. [#1315]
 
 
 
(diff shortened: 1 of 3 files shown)


saimn:
@keflavich - I don't have access to the content here, but I found the `.url` attribute which refer in this case to the eso login page, so it should safe to just check this.

PR #1316, opened 2018-12-18
Added access to member_ouss_id

andamian:
Our users at the CADC would like to be able to query alma on the member_ouss_id attribute. Although not available on the public query page, the attribute is supported on the back end and this patch allows it to be used as a constraint.


keflavich:
@andamian Are you sure it's `member_ouss_id` and not `member_ous_id`?  Could you add a test for this?  And, are there any other hidden parameters we should add?

Otherwise, :+1: from me.

andamian:
Good point, @keflavich. At a second thought, member_ous_id is more inclusive and hence more useful. Made the change and added a test. Thanks

PR #1318, opened 2018-12-20
Make the `all_columns` parameter work for `query_planet`.

wtgee:
This adjusts `query_planet` to pass any additional keyword argmunets to the underlying `get_confirmed_planets_table` method.

This seems like the logical way to fix the problem in #1317 but might not be preferred.

### DIFF
--- a/astroquery/nasa_exoplanet_archive/nasa_exoplanet_archive.py
+++ b/astroquery/nasa_exoplanet_archive/nasa_exoplanet_archive.py
@@ -101,8 +101,7 @@ class NasaExoplanetArchiveClass(object):
 
         return self._table
 
-    def query_planet(self, planet_name, table_path=None,
-                     all_columns=False):
+    def query_planet(self, planet_name, **kwargs):
         """
         Get table of exoplanet properties.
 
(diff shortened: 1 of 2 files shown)


astroquery/nasa_exoplanet_archive/nasa_exoplanet_archive.py
```diff
@@ -101,29 +101,24 @@ def get_confirmed_planets_table(self, cache=True, show_progress=True,
...
         planet_name : str
             Name of planet
-        table_path : str (optional)
-            Path to a local table file. Default `None` will trigger a
-            download of the table from the internet.
-        all_columns : bool (optional)
-            Return all available columns. The default returns only the
-            columns in the default category at the link above.
+        kwargs : dict (optional)
+            Extra keyword arguments passed to ``get_confirmed_planets_table``.
```
keflavich: I think you can use single backticks here, since it's referring to a method on the same class.  I'm not 100% sure of this though...

keflavich:
This solution looks fine to me.  Could you confirm that the remote tests pass for you?  I can't run them where I am right now.

wtgee:
> This solution looks fine to me. Could you confirm that the remote tests pass for you? I can't run them where I am right now.

Sorry, holiday travel. Yes, testing with `--remote` and all tests in relevant files passing.

PR #1321, opened 2018-12-28
Fixing kwarg-passing bug

bmorris3:
Presently the `get_planet` method of the `NasaExoplanetArchive` class has a keyword argument for getting `all_columns` of the table, but the user-supplied keyword is not passed on to the `self.get_confirmed_planets_table` method within the function. This PR corrects that small bug.


keflavich:
See #1318, which is a nearly identical PR responding to #1317.  Your review there would be welcome, but I'm closing this one because it's redundant.

PR #1345, opened 2019-01-29
Adding new service esa_hubble

esdc-esac-esa-int:


keflavich:
I'm currently seeing several failures:
```
astroquery/esa_hubble/tests/test_esa_hubble.py .F.FF...                                                                                                                                                                            [ 88%]
```
These are caused by tests that are querying remote services that shouldn't be.

[Remote tests need to be decorated with the `@remote_data` decorator.](https://astroquery.readthedocs.io/en/latest/testing.html#test-module-remote-py)

astroquery/esa_hubble/tests/test_esa_hubble.py
```diff
@@ -0,0 +1,142 @@
...
+        ehst.get_product("J6FL25S4Q", "RAW")
+        dummyHandler.check_call("get_product", parameters)
+
+    def test_get_artifact(self):
+        parameters = {}
+        parameters['artifact_id'] = "O5HKAX030_FLT.FITS"
+        parameters['verbose'] = False
+        dummyHandler = DummyHandler("get_artifact", parameters)
+        ehst = ESAHubbleClass(dummyHandler, self.get_dummy_tap_handler())
+        ehst.get_artifact("O5HKAX030_FLT.FITS")
```
keflavich: This line is triggering a failure for me locally:
`E           FileNotFoundError: [Errno 2] No such file or directory: '/private/var/folders/_r/c4nkvmpn4vx6qc91qygz8h740000gr/T/astroquery-test-_awt8xh6/lib/python3.6/site-packages/astroquery/esa_hubble/tests/data/O5HKAX030_FLT.FITS'`
it may need to be added to the repository or added in the `setup_package.py` file

astroquery/esa_hubble/tests/dummy_handler.py
```diff
@@ -0,0 +1,82 @@
...
+            with open(file, 'rb') as myfile:
+                data = myfile.read()
+        return data
+
+    def get_table(self, url, filename=None, output_format='votable',
+                  verbose=False):
+        print("************* DummyHandler!!!!!!!!")
+        if filename is None:
+            raise ValueError("filename must be specified")
+        urllib.request.urlretrieve(url, filename)
```
keflavich: I'm unclear why this line is here.  The "DummyHandler" seems to exist for testing purposes, yet this line calls a `urlretrieve`, which requires internet access and therefore is attempting to download the real file.  Could you clarify the purpose of this handler?  If it is accessing web-based resources, it cannot be used in non-remote tests.

astroquery/esa_hubble/tests/test_esa_hubble.py
```diff
@@ -0,0 +1,142 @@
...
+
+        parameters = {}
+        parameters['coordinates'] = c
+        parameters['radius'] = None
+        parameters['file_name'] = None
+        parameters['verbose'] = False
+        dummyHandler = DummyHandler("cone_search", parameters)
+
+        ehst = ESAHubbleClass(dummyHandler, dummyTapHandler)
+        ehst.cone_search(parameters['coordinates'])
```
keflavich: This line is triggering an error:
`E               urllib.error.URLError: <urlopen error An attempt was made to connect to the internet by a test that was not marked `remote_data`. The requested host was: archives.esac.esa.int>`
which is related to the remote_data issue I reported

astroquery/esa_hubble/tests/test_esa_hubble.py
```diff
@@ -0,0 +1,142 @@
...
+        ehst.get_product("X0MC5101T")
+        dummyHandler.check_call("get_postcard", parameters)
+
+    def test_query_target(self):
+        parameters = {}
+        parameters['name'] = "m31"
+        parameters['verbose'] = False
+        dummyHandler = DummyHandler("query_target", parameters)
+        ehst = ESAHubbleClass(dummyHandler, self.get_dummy_tap_handler())
+        ehst.query_target(parameters['name'])
```
keflavich: This line is triggering a remote_data issue

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+
+class ESAHubbleHandler(object):
+
+    def __init__(self):
+        return
+
+    def get_file(self, url, filename=None, verbose=False):
+        if filename is None:
+            raise ValueError("filename must be specified")
+        urllib.request.urlretrieve(url, filename)
```
keflavich: Instead of using urllib, you should use astroquery's built-in tools.  See https://github.com/astropy/astroquery/blob/master/astroquery/query.py#L245

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+        if "/" not in filename:
+            print("File " + str(filename) + " downloaded to current directory")
+        else:
+            print("File " + str(filename) + " downloaded")
+
+    def get_table(self, url, filename=None, output_format='votable',
+                  verbose=False):
+        if filename is None:
+            raise ValueError("filename must be specified")
+        urllib.request.urlretrieve(url, filename)
```
keflavich: ditto above - `urllib` should not be needed in astroquery.

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+            chosen. RAW, CALIBRATED, PRODUCT or AUXILIARY
+            filename : string, file name to be used to store the artifact,
+            optional, default None
+            File name for the observation.
+            verbose : bool, optional, default 'False'
+            flag to display information about the process
+
+            Returns
+            -------
+            None. It downloads the observation indicated
```
keflavich: This appears to be incorrect; there is a returned object

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+        """
+
+        obs_id = "OBSERVATION_ID=" + observation_id
+        cal_level = "CALIBRATION_LEVEL=" + calibration_level
+        link = self.data_url + obs_id + "&" + cal_level
+        if filename is None:
+            filename = observation_id + ".tar"
+        print(link)
```
keflavich: Avoid print statements.  If necessary, we prefer log statements using `astropy.log`, but in general, any print statements should at least be hidden under an `if verbose:` clause

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+        return self._handler.get_file(link, filename, verbose)
+
+    def get_artifact(self, artifact_id, filename=None, verbose=False):
+        """ Download artifacts from EHST
```
keflavich: It would be helpful to describe what an artifact is, and perhaps link to the full documentation on artifacts

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+            The identifier of the physical product (file) we want to retrieve.
+            filename : string, file name to be used to store the artifact,
+            optional, default None
+            File name for the artifact
+            verbose : bool, optional, default 'False'
+            flag to display information about the process
+
+            Returns
+            -------
+            None. It downloads the artifact indicated
```
keflavich: Same comment as above - there is a return statement

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+            """
+
+        art_id = "ARTIFACT_ID=" + artifact_id
+        link = self.data_url + art_id
+        if filename is None:
+            filename = artifact_id
+        print(link)
```
keflavich: `if verbose:`

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+            Returns
+            -------
+            None. It downloads the observation postcard indicated
+        """
+
+        retri_type = "RETRIEVAL_TYPE=POSTCARD"
+        obs_id = "OBSERVATION_ID=" + observation_id
+        cal_level = "CALIBRATION_LEVEL=" + calibration_level
+        res = "RESOLUTION=" + str(resolution)
+        link = "".join((self.data_url,
```
keflavich: this looks easier to write as:
`link = self.data_url + "&".join(retri_type, obs_id, cal_level, res)`

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+                        cal_level,
+                        "&",
+                        res))
+        if filename is None:
+            filename = observation_id + ".tar"
+        print(link)
+        return self._handler.get_file(link, filename, verbose)
+
+    def cone_search(self, coordinates, radius=0.0, filename=None,
+                    output_format='votable', verbose=False):
```
keflavich: missing docstring

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+                   "  AND  INTERSECTS(CIRCLE('ICRS'," + str(ra) + ""
+                   "," + str(dec) + "," + str(radiusInGrades) + ""
+                   "),POSITION)=1  AND  PLANE.MAIN_SCIENCE_PLANE='true' "
+                   "ORDER BY PROPOSAL.PROPOSAL_ID "
+                   "DESC",
+                   # "PAGE": "1",
+                   # "PAGE_SIZE": "50",
+                   "RETURN_TYPE": str(output_format)}
+        result = urllib.parse.urlencode(payload,
+                                        quote_via=urllib.parse.quote_plus)
```
keflavich: this should use `requests` instead.  `urlencode` is not needed, you can just do `requests.get(url, payload=dict)` or similar.  In astroquery, you instead do `self._request('GET', url, payload=dict)` or similar

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+        if only_names is True:
+            column_names = []
+            for c in columns:
+                column_names.append(c.name)
+            return column_names
+        else:
+            return columns
+
+    def _checkQuantityInput(self, value, msg):
+        if not (isinstance(value, str) or isinstance(value, units.Quantity)):
```
keflavich: these input validators are useful; they might be worth including in a separate file in `astroquery.utils`.  You don't need to do this now, I'm just pointing out that they are generally useful tools that, if we want to use them in other packages in the future (e.g., other ESA modules), should go into a different location

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+            raise ValueError(str(msg) + ""
+                             " must be either a string or astropy.coordinates")
+
+    def _getQuantityInput(self, value, msg):
```
keflavich: the logic in this one is a bit convoluted.  It would be simpler to restructure as:

```
if value is None:
     error msg
elif isinstance(value, str):
    ....
elif isinstance(value, Quantity):
    ...
else:
    error msg
```

docs/esa_hubble/esa_hubble.rst
```diff
@@ -0,0 +1,109 @@
...
+
+---------------------------
+1. Getting Hubble products
+---------------------------
+
+.. code-block:: python
+
+  >>> from astroquery.esa_hubble import ESAHubble
+  >>>
+  >>> ESAHubble.get_product("J6FL25S4Q", "RAW", "raw_data_for_J6FL25S4Q.tar")
```
keflavich: It would be helpful to explain where these IDs come from

docs/esa_hubble/esa_hubble.rst
```diff
@@ -0,0 +1,109 @@
...
+
+.. code-block:: python
+
+  >>> from astroquery.esa_hubble import ESAHubble
+  >>>
+  >>> ESAHubble.get_postcard("J6FL25S4Q", "RAW", 256, "raw_postcard_for_J6FL25S4Q.jpg")
+  http://archives.esac.esa.int/ehst-sl-server/servlet/data-action?RETRIEVAL_TYPE=POSTCARD&OBSERVATION_ID=\
+  J6FL25S4Q&CALIBRATION_LEVEL=RAW&RESOLUTION=256
+
+This will download the postcard for the observation 'J8VP03010' with low resolution (256) and it will stored in a tar called 'raw_postcard_for_J6FL25S4Q.jpg'. Resolution of 1024 is also available.
```
keflavich: is this a tar file or a jpg?

docs/esa_hubble/esa_hubble.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Calibration levels can be RAW, CALIBRATED, PRODUCT or AUXILIARY.
+
+---------------------------
+3. Getting Hubble artifacts
```
keflavich: as above, describe what artifacts are

docs/esa_hubble/esa_hubble.rst
```diff
@@ -0,0 +1,109 @@
...
+---------------------------
+
+.. code-block:: python
+
+  >>> from astroquery.esa_hubble import ESAHubble
+  >>>
+  >>> ESAHubble.get_artifact("O5HKAX030_FLT.FITS")
```
keflavich: Where does the filename come from?  i.e., how does a user find this file?

docs/esa_hubble/esa_hubble.rst
```diff
@@ -0,0 +1,109 @@
...
+
+.. code-block:: python
+
+  >>> from astroquery.esa_hubble import ESAHubble
+  >>>
+  >>> ESAHubble.query_target("m31", "m31_query.xml")
+  http://archives.esac.esa.int/ehst-sl-server/servlet/metadata-action?RESOURCE_CLASS=OBSERVATION&\
+  SELECTED_FIELDS=OBSERVATION&QUERY=(TARGET.TARGET_NAME=='m31')&RETURN_TYPE=VOTABLE
+
+This will download metadata for all observations associated with target name 'm31'. The result of the query will be stored in file 'm31_query.xml'.
```
keflavich: This doesn't match the astroquery API.  `query_target` should return an `astropy.table.Table` object, which can then be written to file.

docs/esa_hubble/esa_hubble.rst
```diff
@@ -0,0 +1,109 @@
...
+  http://archives.esac.esa.int/ehst-sl-server/servlet/metadata-action?  RESOURCE_CLASS=OBSERVATION&ADQLQUERY=SELECT+DISTINCT+OBSERVATION%2COBSERVATION.TYPE%2CTARGET.MOVING_TARGET%2CTARGET.TARGET_NAME%2CTARGET.TARGET_DESCRIPTION%2CPROPOSAL.PROPOSAL_ID%2CPROPOSAL.PI_NAME%2CPROPOSAL.PROPOSAL_TITLE%2CINSTRUMENT.INSTRUMENT_NAME%2CPLANE.METADATA_PROVENANCE%2CPLANE.DATA_PRODUCT_TYPE%2CPLANE.SOFTWARE_VERSION%2CPOSITION.RA%2CPOSITION.DEC%2CPOSITION.GAL_LAT%2CPOSITION.GAL_LON%2CPOSITION.ECL_LAT%2CPOSITION.ECL_LON%2CPOSITION.FOV_SIZE%2CENERGY.WAVE_CENTRAL%2CENERGY.WAVE_BANDWIDTH%2CENERGY.WAVE_MAX%2CENERGY.WAVE_MIN%2CENERGY.FILTER+FROM+FIELD_NOT_USED++WHERE+OBSERVATION.COLLECTION%3D%27HST%27++AND++PLANE.MAIN_SCIENCE_PLANE%3D%27true%27++AND++%28OBSERVATION.TYPE%3D%27HST+Composite%27+OR+OBSERVATION.TYPE%3D%27HST+Singleton%27%29++AND++INTERSECTS%28CIRCLE%28%27ICRS%27%2C10.685469872614046%2C41.26901534788858%2C0.0%29%2CPOSITION%29%3D1++AND++PLANE.MAIN_SCIENCE_PLANE%3D%27true%27+ORDER+BY+PROPOSAL.PROPOSAL_ID+DESC&RETURN_TYPE=VOTABLE
+
+This will perform a cone search with radius 7 arcmins. The result of the query will be stored in the votable file 'cone_search_m31_5.vot'.
```
keflavich: same here: cone searches that return votables should be converted to astropy table objects.  They can be written to disk after that.

docs/esa_hubble/esa_hubble.rst
```diff
@@ -0,0 +1,109 @@
...
+
+-------------------------------
+5. Getting access to catalogues 
+-------------------------------
+
+The query_hst_tap function provides access to the HST archive database using the Table Access Protocol (TAP) and via the Astronomical Data Query Language (ADQL).
```
keflavich: function names should be wrapped in backticks.  Also, you should quote the full path, e.g.:
```suggestion
The `~ESAHubble.query_hst_tap` function provides access to the HST archive database using the Table Access Protocol (TAP) and via the Astronomical Data Query Language (ADQL).
```
Note that the leading ~ means that this will be rendered as `query_hst_tap`, but it will link to the docs for that function

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+
+        Returns
+        -------
+        A Job object
+        """
+
+        return self._tap.launch_job(query=query, output_file=output_file,
+                                    output_format=output_format,
+                                    verbose=False,
+                                    dump_to_file=output_file is not None)
```
keflavich: Saving to file as an option is acceptable, but this query should return a `Table` object, not a `Job` object.

keflavich: Generally, though, we strongly prefer that users adopt a workflow like:
```
table = ESAHubble.query_hst_tap(parameters)
table.writeto('my_tablename.vot', format='ascii.votable')
```
over something like
```
table = ESAHubble.query_hst_tap(parameters, output_filename='my_votablename.vot')
```

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,374 @@
...
+        urllib.request.urlretrieve(url, filename)
+        table = modelutils.read_results_table_from_file(filename,
+                                                        str(output_format))
+        return table
+
+
+Handler = ESAHubbleHandler()
+
+
+class ESAHubbleClass(object):
```
keflavich: This class should inherit from [`BaseQuery`](https://github.com/astropy/astroquery/blob/master/astroquery/query.py#L142)

esdc-esac-esa-int:
Hi @keflavich,

I am refactoring the code following your advises. I am having problems by replacing this line:

result = urllib.parse.urlencode(payload,
                                                    quote_via=urllib.parse.quote_plus)

Is there any other way to encode a url using astroquery functions?

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,385 @@
...
+                   "OBSERVATION.TYPE='HST Singleton')"
+                   "  AND  INTERSECTS(CIRCLE('ICRS'," + str(ra) + ""
+                   "," + str(dec) + "," + str(radiusInGrades) + ""
+                   "),POSITION)=1  AND  PLANE.MAIN_SCIENCE_PLANE='true' "
+                   "ORDER BY PROPOSAL.PROPOSAL_ID "
+                   "DESC",
+                   # "PAGE": "1",
+                   # "PAGE_SIZE": "50",
+                   "RETURN_TYPE": str(output_format)}
+        result = urllib.parse.urlencode(payload)
```
keflavich: If this is the instance of `urlencode` you were talking about, it simply isn't needed.  You can do something like:
```
request.get(base_url, data=payload)
```
which send the `payload` as part of the URL, e.g.
```
payload = {'key':'val'}
response = requests.get('http://httpbin.org/get', params=payload)
print(response.url)
```
the print will be `http://httpbin.org/get?key=val`.

In astroquery, instead of `requests.get`, do:
`self._request('GET', url, params=...)`

keflavich: The approach I recommend will perform the request; it's not just assembling the URL.

esdc-esac-esa-int: Yes, I know. I will change that tomorrow. I just wanted to make sure the rest is fine

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,396 @@
...
+__all__ = ['EsaHubble', 'EsaHubbleClass']
+
+
+class ESAHubbleHandler(BaseQuery):
+
+    def __init__(self):
+        super(ESAHubbleHandler, self).__init__()
+
+    def get_file(self, filename, response, verbose=False):
+        file = open(filename, 'wb')
```
keflavich: Context managers are a better way to do this:
```suggestion
        with open(filename, 'wb') as file:
             file.write(response.content)
```

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,396 @@
...
+
+            Returns
+            -------
+            None. It downloads the observation indicated
+        """
+
+        obs_id = "OBSERVATION_ID=" + observation_id
+        cal_level = "CALIBRATION_LEVEL=" + calibration_level
+        link = self.data_url + obs_id + "&" + cal_level
+        result = self._handler.request('GET', link, params=None)
```
keflavich: `params=None` is redundant
```suggestion
        result = self._handler.request('GET', link)
```

astroquery/esa_hubble/core.py
```diff
@@ -0,0 +1,396 @@
...
+        file = open(filename, 'wb')
+        file.write(response.content)
+        file.close()
+
+        table = modelutils.read_results_table_from_file(filename,
+                                                        str(output_format))
+        return table
+
+    def request(self, t="GET", link=None, params=None):
+        return self._request(method=t, url=link, params=params)
```
keflavich: Why did you add this extra layer of wrapper?  This looks like unnecessary added complexity to me

esdc-esac-esa-int: What I aimed with the handlers is to provide a mean to separate the core behavior of the module from the accesses to the real services. It is a way to mock the services for testing. This way I can have a default handler for accessing the real service and a dummy handler (that does nothing) for testing.
May I ask you, for the time being, to accept that? I can refactor the code as you please in the upcoming updates. Doing this now will impact our schedule.
Is that ok with you?

esdc-esac-esa-int:
What I aimed with the handlers is to provide a mean to separate the core behavior of the module from the accesses to the real services. It is a way to mock the services for testing. This way I can have a default handler for accessing the real service and a dummy handler (that does nothing) for testing.
May I ask you, for the time being, to accept that? I can refactor the code as you please in the upcoming updates. Doing this now will impact our schedule.
Is that ok with you?

keflavich:
OK, I understand the purpose of that structure.  However, that is at odds with the design of other astropy modules; we use `mock` tools to separate the "real" service from the "testing".

We can accept the workaround using the dummy module for testing, at least for now, but we want it refactored eventually. 

There are a few more minor comments (e.g., you call a file a "tar" file, but it has a "jpg" suffix) that you should resolve before we can merge.

@bsipocz , any further opinions?  We might be ready to merge this, with the caveat that the 'handler' section needs to be factored out.

keflavich:
@esdc-esac-esa-int could you rebase this PR against master?  The travis failure relates to a version issue in another package that is fixed in master; rebasing should fix it.

keflavich:
Something has gone wrong; why are there changes to `ah_bootstrap` in this PR?  Also... the diff shows me a change to the README that isn't a change.  I'm not clear what's going on yet.

keflavich:
While the latest change made tests go green, it reverted several things in master, which is not acceptable.  The problem is that you merged in a different branch.  To merge this into master, we need you to rebase, i.e., we need a clean commit history.  Most importantly, though, there should not be any changes in this PR to files that are not related to the PR: `ah_bootstrap.py`, `README.rst`, and `.gitignore` should not be changed.

PR #1370, opened 2019-02-25
Dace service

julienb54:
Hello,

In Geneva Observatory, we would like to add our API DACE (Data Analysis Center for Exoplanets : https://dace.unige.ch) in astroquery. This is a first small version giving only public data of radial velocities of observed objects (stars, etc.). Next step will be to implement authentication, other useful APIs for astrophysicists and query_object(objectname, ...) (according to your requirements) but will need extra dev on our side. For now we don't have query_regions and get_images but can happen in the future.
This is my first contribution so don't hesitate to tell me if something is missing.

Have a nice day,

Julien

### DIFF
--- a/astroquery/dace/core.py
+++ b/astroquery/dace/core.py
@@ -44,14 +44,14 @@ class DaceClass(BaseQuery):
     def _parse_result(self, response, verbose=False):
         try:
             json_data = response.json()
-            dace_dict = self._transform_data_as_dict(json_data)
+            dace_dict = self.transform_data_as_dict(json_data)
             return Table(dace_dict)
         except Exception as ex:
             raise ParseException(
                 "Failed to parse DACE result! Exception : {} \nRaw content : {}".format(ex, response.text))
 
     @staticmethod
-    def _transform_data_as_dict(json_data):
+    def transform_data_as_dict(json_data):
         """
         Internally DACE data are provided using protobuff. The format is a list of parameters. Here we parse
         these data to give to the user something more readable and ignore the internal stuff
--- a/astroquery/dace/core.py
+++ b/astroquery/dace/core.py
@@ -13,20 +13,19 @@ class ParseException(Exception):
     """Raised when parsing Dace data fails"""
 
 
-"""
-DACE class to access DACE (Data Analysis Center for Exoplanets) data based in Geneva Observatory  
-"""
-
-
 @async_to_sync
 class DaceClass(BaseQuery):
+    """
+    DACE class to access DACE (Data Analysis Center for Exoplanets) data based in Geneva Observatory
+    """
+
     __DACE_URL = conf.server
     __DACE_TIMEOUT = conf.timeout
     __OBSERVATION_ENDPOINT = 'ObsAPI/observation/'
     __RADIAL_VELOCITIES_ENDPOINT = __OBSERVATION_ENDPOINT + 'radialVelocities/'
     __HEADERS = {'Accept': 'application/json'}
 
-    def query_radial_velocities(self, object_name):
+    def query_radial_velocities_async(self, object_name):
         """
         Get radial velocities data for an object_name. For example : HD40307
 
@@ -36,13 +35,12 @@ class DaceClass(BaseQuery):
 
         Return
         ------
-        table : ~astropy.table.Table Table containing radial velocities data for the object_name given
+        Raw HTTP response
         """
-        response = self._request("GET", ''.join([self.__DACE_URL, self.__RADIAL_VELOCITIES_ENDPOINT, object_name]),
-                                 timeout=self.__DACE_TIMEOUT, headers=self.__HEADERS)
-        return self._parse_result(response)
+        return self._request("GET", ''.join([self.__DACE_URL, self.__RADIAL_VELOCITIES_ENDPOINT, object_name]),
+                             timeout=self.__DACE_TIMEOUT, headers=self.__HEADERS)
 
-    def _parse_result(self, response):
+    def _parse_result(self, response, verbose=False):
         try:
             json_data = response.json()
             dace_dict = self._transform_data_as_dict(json_data)
@@ -75,6 +73,3 @@ class DaceClass(BaseQuery):
 
 
 Dace = DaceClass()
-
-# Next you should write the docs in astroquery/docs/module_name
-# using Sphinx.
--- a/astroquery/dace/core.py
+++ b/astroquery/dace/core.py
@@ -31,11 +31,12 @@ class DaceClass(BaseQuery):
 
         Parameters
         ----------
-        object_name : the target you want radial velocities data
+        object_name : str
+            The target you want radial velocities data
 
         Return
         ------
-        Raw HTTP response
+        response : a ``requests.Response`` from DACE
         """
         return self._request("GET", ''.join([self.__DACE_URL, self.__RADIAL_VELOCITIES_ENDPOINT, object_name]),
                              timeout=self.__DACE_TIMEOUT, headers=self.__HEADERS)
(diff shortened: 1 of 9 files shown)


astroquery/dace/core.py
```diff
@@ -0,0 +1,80 @@
...
+__all__ = ['Dace', 'DaceClass']
+
+
+class ParseException(Exception):
+    """Raised when parsing Dace data fails"""
+
+
+"""
+DACE class to access DACE (Data Analysis Center for Exoplanets) data based in Geneva Observatory  
+"""
```
keflavich: this comment block should probably be part of the class

astroquery/dace/core.py
```diff
@@ -0,0 +1,80 @@
...
+        ----------
+        object_name : the target you want radial velocities data
+
+        Return
+        ------
+        table : ~astropy.table.Table Table containing radial velocities data for the object_name given
+        """
+        response = self._request("GET", ''.join([self.__DACE_URL, self.__RADIAL_VELOCITIES_ENDPOINT, object_name]),
+                                 timeout=self.__DACE_TIMEOUT, headers=self.__HEADERS)
+        return self._parse_result(response)
```
keflavich: because of the `async_to_sync` method above, you could (and should) rename this function to `query_radial_velocities_async` and have it return `response`.  `async_to_sync` will automatically create a `query_radial_velocities` function that returns `self._parse_result(response)`.

julienb54: Hello,

Thanks for the quick feedback ^^.

keflavich:
Add the documentation to the docs index and add a changelog entry, that should get the rest of the CI checks passing.

astroquery/dace/core.py
```diff
@@ -0,0 +1,75 @@
...
+    def _transform_data_as_dict(json_data):
+        """
+        Internally DACE data are provided using protobuff. The format is a list of parameters. Here we parse
+        these data to give to the user something more readable and ignore the internal stuff
+        """
+        data = defaultdict(list)
+        parameters = json_data['parameters']
+        for parameter in parameters:
+            variable_name = parameter['variableName']
+            data[variable_name].extend(parameter['doubleValues']) if 'doubleValues' in parameter \
```
bsipocz: this is not very readable like this, what about refactoring this a bit, e.g. setting them to None by default and looping through the possible parameter names? Or even better, use `parameter.get(whatevervaluesyouloop)`? that returns `None` if the key doesn't exist.

astroquery/dace/tests/test_dace_remote.py
```diff
@@ -0,0 +1,15 @@
...
+from astropy.tests.helper import remote_data
+from astroquery.dace import Dace
+
+
+@remote_data
+class TestDaceClass(unittest.TestCase):
+
+    def test_should_get_radial_velocities(self):
+        radial_velocities_table = Dace.query_radial_velocities('HD40307')
+        assert radial_velocities_table is not None and 'rv' in radial_velocities_table.colnames
```
bsipocz: can you add one more assert here, to check reasonable number of observations returned in the table? Currently an empty table passes this test.

astroquery/dace/core.py
```diff
@@ -0,0 +1,75 @@
...
+    DACE class to access DACE (Data Analysis Center for Exoplanets) data based in Geneva Observatory
+    """
+
+    __DACE_URL = conf.server
+    __DACE_TIMEOUT = conf.timeout
+    __OBSERVATION_ENDPOINT = 'ObsAPI/observation/'
+    __RADIAL_VELOCITIES_ENDPOINT = __OBSERVATION_ENDPOINT + 'radialVelocities/'
+    __HEADERS = {'Accept': 'application/json'}
+
+    def query_radial_velocities_async(self, object_name):
```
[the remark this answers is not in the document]
julienb54: Thanks for all your comments ! I will update the code.
About query_object, I think I will implement it in a second step. query_radial_velocities_async is named like this because it really return only radial velocities data.
For query_object we have another API on our side to call but it is not fully implemented.
I created only radial_velocities method for now because this is a request from our astrophysicists.
To summarize, if it is ok for you I'll start with this and will add query_object later.

astroquery/dace/tests/test_dace_remote.py
julienb54: You are right. I will add a more meaningful test.

astroquery/dace/core.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        return self._request("GET", ''.join([self.__DACE_URL, self.__RADIAL_VELOCITIES_ENDPOINT, object_name]),
+                             timeout=self.__DACE_TIMEOUT, headers=self.__HEADERS)
+
+    def _parse_result(self, response, verbose=False):
+        try:
+            json_data = response.json()
+            dace_dict = self._transform_data_as_dict(json_data)
+            return Table(dace_dict)
+        except Exception as ex:
```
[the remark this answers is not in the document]
julienb54: You're right. I will change it.

astroquery/dace/core.py
julienb54: Ok, I'll do the refactoring and come back to you. Thx.

astroquery/dace/core.py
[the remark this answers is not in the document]
bsipocz: And what would `query_object` return in addition to the RV data? I'm also interested in the differences in the data structure of the returns.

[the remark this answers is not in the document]
julienb54: We rediscussed with astrophysicists on our side to understand their needs regarding the DACE APIs.
At the moment we have two kind of data to retrieve for a given object : 

### **1. time series** 
This is final data product of observations. It could span over hours or years. These are mainly used to search for and characterize time dependent astronomical phenomena such as binary stars, exoplanets companions. It could also be used to study stellar activity on different time scales. 
Such series could be related to :
**High resolution spectroscopic data analysis** 
- radial velocities+ errors
- stellar activity indexes+ errors
- other time dependent parameters linked  to high resolution spectroscopic data analysis
 
**Imaging data analysis**
- Light curves + errors

### **2. Reduced data products.**
Possibility to download reduced data products in the "fits" file format for different instruments we have builded ie.   CORALIE, HARPS, ESPRESSO and soon NIRPS

Based on these requirements,  we suggest to implement the following methods :
1. **retrieve_time_series**(object_name, type=spectroscopy/photometry/etc., columns=[list of columns]). For « type » parameter, in this first pull request only spectroscopy will be implemented => rajouter des options. rv, activity index etc...
2. **retrieve_high_resolution_spectroscopy_data**(fits_file_list). A list of fits file to download

Best regards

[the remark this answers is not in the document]
julienb54: Don't hesitate if you have questions, suggestions, etc.
I tried to explain the needs from researchers, hope it will help to take a decision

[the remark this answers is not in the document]
julienb54: About query_object, initially my idea was to return also a Table astropy with stellar data : ra, dec, magv, etc. but it is not implemented on our side at the moment.

astroquery/dace/core.py
julienb54: I refactored using .get() method.

julienb54: Just saw that my test should be in a non remote test file. Working on it.

astroquery/dace/core.py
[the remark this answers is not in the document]
julienb54: Any news ?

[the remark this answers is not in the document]
julienb54: Hello,

Don't hesitate to come back to me. I close the discussion for now. i let you decide what's best and I'll implement it.

Regards,

Julien

keflavich:
@bsipocz we need to review and merge this.

keflavich:
@bsipocz I'm going to rule that this is fine as is.  The eventual API may include `query_radial_velocities`, `query_timeseries`, `query_fits`, but `query_object` doesn't obviously belong here.

@julienb54 for all of these, you might consider using the `get_radial_velocities`, `get_timeseries`, or `retrieve_radial_velocities` etc. approach rather than `query_<x>`, but it's not entirely clear which of these is most appropriate.  I apologize for the long delay in examining this PR.

PR #1388, opened 2019-03-19
changed error types

Himanshu-Garg:
as per issue #1037 

I have changed some Exception to correct Error types...
and removed a useless line (maybe it was a typo)

### DIFF
--- a/astroquery/irsa/core.py
+++ b/astroquery/irsa/core.py
@@ -451,7 +451,7 @@ def _parse_coordinates(coordinates):
             warnings.warn("Coordinate string is being interpreted as an "
                           "ICRS coordinate.")
         except u.UnitsError as ex:
-            warnings.warn("Only ICRS coordinates can be entered as strings \n"
+            warnings.warn("Only ICRS coordinates can be entered as strings\n"
                           "For other systems please use the appropriate "
                           "astropy.coordinates object")
 
--- a/astroquery/irsa/core.py
+++ b/astroquery/irsa/core.py
@@ -454,6 +454,7 @@ def _parse_coordinates(coordinates):
             warnings.warn("Only ICRS coordinates can be entered as strings\n"
                           "For other systems please use the appropriate "
                           "astropy.coordinates object")
+            raise ex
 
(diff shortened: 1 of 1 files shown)


astroquery/irsa/core.py
```diff
@@ -451,10 +451,10 @@ def _parse_coordinates(coordinates):
             warnings.warn("Coordinate string is being interpreted as an "
                           "ICRS coordinate.")
         except u.UnitsError as ex:
-            warnings.warn("Only ICRS coordinates can be entered as strings\n"
+            warnings.warn("Only ICRS coordinates can be entered as strings \n"
                           "For other systems please use the appropriate "
                           "astropy.coordinates object")
-            raise ex
+
```
keflavich: does this Exception still need to be raised?  i.e., can the code proceed from this point?

astroquery/irsa/core.py
```diff
@@ -451,10 +451,10 @@ def _parse_coordinates(coordinates):
             warnings.warn("Coordinate string is being interpreted as an "
                           "ICRS coordinate.")
         except u.UnitsError as ex:
-            warnings.warn("Only ICRS coordinates can be entered as strings\n"
+            warnings.warn("Only ICRS coordinates can be entered as strings \n"
```
keflavich: there's no need for a space before a newline.  A period would be appropriate though.

astroquery/irsa/core.py
Himanshu-Garg: can you elaborate a little pls...

keflavich: Can the code beyond this point execute given the failure here?  An exception is being raised, therefore the code beyond this point has never been executed.  Make sure that it can.  Most likely, the `raise ex` is here intentionally because the code beyond this line _cannot_ be run once you've reached this line.

keflavich:
@bsipocz this one can be merged, though given how old it is (which is our fault), should we rebase and re-trigger CI tests first?

keflavich:
On closer inspection, some of these exception types are incorrect, so I'm closing.

PR #1472, opened 2019-05-23
ALMA login URLs are a moving target

keflavich:
See #1395, #1381.  This is an attempt to make ALMA's login tool more robust
against these changes and give a better error message.


keflavich:
I deem this non-changelog-worthy since it's just yet another upstream correction.

PR #1486, opened 2019-06-07
Add get_images function to CADC module

ellieb:
As per issue https://github.com/astropy/astroquery/issues/1445, adding the [get_images, get_images_async, and get_image_list function](https://astroquery.readthedocs.io/en/latest/api.html#get-images) to the CADC module. (BTW This is my first PR for this project, so it might need a lot of help)

This PR uses pyvo to access the datalink service, so https://github.com/astropy/astroquery/pull/1449 must be merged before this PR gets merged.


astroquery/cadc/core.py
```diff
@@ -237,6 +238,121 @@ def get_collections(self):
...
+
+        Returns
+        -------
+        A list of URLs to data.
+        """
+
+        # TODO: Get access url from publisher ID to find which datalink service to use (i.e. datalink or sc2links)
+        # TODO: Check datalink reponse
+
+        def chunks(l, n):
```
keflavich: Could you use variable names with >1 character?  That's part of our style preference in astroquery, in part so that we can easily search for those variable names.

ellieb: Sure, makes sense :+1:

astroquery/cadc/core.py
```diff
@@ -237,6 +239,136 @@ def get_collections(self):
...
+            Coordinates around which to query
+        radius : str or `astropy.units.Quantity`.
+            The radius of the cone search AND cutout area
+
+        Returns
+        -------
+        A list of URLs to data.
+        """
+
+        def chunks(obj_list, chunk_len):
```
andamian: For brevity, I would replace the function with the oneliner:
`(publisher_ids[pos:pos + batch_size] for pos in range(0, len(publisher_ids), batch_size))` with a short comment at the beginning of the block.

astroquery/cadc/core.py
```diff
@@ -237,6 +239,136 @@ def get_collections(self):
...
+            :return: An iterator that goes through each sublist
+            """
+            for idx in range(0, len(obj_list), chunk_len):
+                yield obj_list[idx:idx + chunk_len]
+
+        if not query_result:
+            raise AttributeError('Missing query_result argument')
+
+        # Send datalink requests in batches of 20 publisher ids
+        n_pids = 20
```
andamian: Please move closer to its usage.

astroquery/cadc/core.py
```diff
@@ -237,6 +239,136 @@ def get_collections(self):
...
+        # Iterate through list of sublists to send datalink requests in batches
+        for pid_sublist in chunks(publisher_ids, n_pids):
+            datalink = DatalinkResults.from_result_url(
+                '{}?{}'.format(self.data_link_url, urlencode({'ID': pid_sublist}, True)))
+            for service_def in datalink.bysemantics('#cutout'):
+                access_url = service_def.access_url.decode('ascii')
+                if '/sync' in access_url:
+                    service_params = service_def.input_params
+                    input_params = {param.name: param.value for param in service_params if
+                                    param.name in ['ID', 'RUNID']}
```
andamian: both `ID` and 'RUNID` are valid?

astroquery/cadc/core.py
```diff
@@ -237,6 +239,136 @@ def get_collections(self):
...
+                'caomPublisherID column missing from query_result argument')
+        result = []
+
+        # Iterate through list of sublists to send datalink requests in batches
+        for pid_sublist in chunks(publisher_ids, n_pids):
+            datalink = DatalinkResults.from_result_url(
+                '{}?{}'.format(self.data_link_url, urlencode({'ID': pid_sublist}, True)))
+            for service_def in datalink.bysemantics('#cutout'):
+                access_url = service_def.access_url.decode('ascii')
+                if '/sync' in access_url:
```
andamian: Do we have `async` too? What is the course of action in that case? What should the function do when `async` is not present?

astroquery/cadc/core.py
```diff
@@ -237,6 +239,136 @@ def get_collections(self):
...
+
+        if get_query_payload:
+            return request_payload
+
+        query = request_payload['query'] + " AND (dataProductType = 'image')"
+        response = self.run_query(query, operation='sync')
+        query_result = response.get_results()
+
+        if query_result and len(query_result) == 2000:
+            logger.debug("Synchronous query results capped at 2000 results - results may be truncated")
```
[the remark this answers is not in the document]
ellieb: Ok! yeah, it might be nice for users to know when they've hit the limit, but I'll take the check out of `get_images`

astroquery/cadc/core.py
```diff
@@ -237,6 +239,136 @@ def get_collections(self):
                     append(row['energy_emBand'])
         return collections
 
+    @class_or_instance
+    def get_images(self, coordinates, radius=0.016666666666667,
+                   collection=None,
+                   get_query_payload=False,
```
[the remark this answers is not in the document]
ellieb: Most have `get_query_payload`, but esasky and skyview (and maybe a few more) don't. It's in the template_module file, but I agree that it doesn't add value. Should I take it out?

astroquery/cadc/core.py
ellieb: sure, i'll change the name to `batch_size` as well

astroquery/cadc/core.py
ellieb: mmm, much cleaner!

astroquery/cadc/core.py
ellieb: `ID` is needed, while `RUNID` is just for logging

astroquery/cadc/core.py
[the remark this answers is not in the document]
andamian: For consistency, let's leave it for now. Thanks for checking

astroquery/cadc/core.py
```diff
@@ -275,6 +276,162 @@ def get_collections(self):
...
+        radius : str or `astropy.units.Quantity`
+            The radius of the cone search AND cutout area.
+        collection: str, optional
+            Name of the CADC collection to query.
+        get_url_list: bool, optional
+            If true, returns the list of data urls rather than
+            the list of context managers. Default is `False`.
+        show_progress: bool, optional
+            Whether to display a progress bar if the file is downloaded
+            from a remote server.  Default is `False`.
```
keflavich: might need to be ``False`` (double backticks) ?

astroquery/cadc/core.py
```diff
@@ -237,6 +239,136 @@ def get_collections(self):
...
+                images.append(obj.get_fits())
+        except HTTPError as err:
+            logger.debug(
+                "{} - Problem retrieving the file: {}".format(str(err), str(err.url)))
+            pass
+
+        return images
+
+    @class_or_instance
+    def get_image_list(self, query_result, coordinates, radius):
```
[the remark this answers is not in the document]
keflavich: It's fine to expose something like this; other astroquery modules do the same sort of thing - we have separate public functions for getting lists of downloadable objects and downloading said objects.

astroquery/cadc/core.py
```diff
@@ -275,6 +276,162 @@ def get_collections(self):
...
+
+    def get_image_list(self, query_result, coordinates, radius):
+        """
+        Function to map the results of a CADC query into URLs to
+        corresponding data and cutouts that can be later downloaded.
+
+        The function uses the IVOA DataLink Service
+        (http://www.ivoa.net/documents/DataLink/) implemented at the CADC.
+        It works directly with the results produced by Cadc.query_region and
+        Cadc.query_name but in principle it can work with other query
```
keflavich: these can probably be backticked to link to the appropriate modules, e.g. `Cadc.query_region`.  I'm not sure whether the full path is needed in scope here...

astroquery/cadc/core.py
```diff
@@ -275,6 +276,162 @@ def get_collections(self):
...
+        results produced with the Cadc query as long as the results
+        contain the 'publisherID' column. This column is part of the
+        caom2.Plane table.
+
+        Parameters
+        ----------
+        query_result : result returned by Cadc.query_region() or
+                    Cadc.query_name(). In general, the result of any
```
keflavich: this should be split up into `query_result : data type` followed by the description on the following lines; as written, this will format funny

keflavich:
I left a few suggestions for fixing sphinx docs; you should check locally how the docs look when you build them, as they might be OK without my suggested changes.

ellieb:
Ok, great! I'll take a look/ make requested changes tomorrow morning. Thank you!!

astroquery/cadc/core.py
ellieb: Ah, yes. That looks better!

astroquery/cadc/core.py
ellieb: You are right! Full path is not needed.

astroquery/cadc/core.py
ellieb: Thanks. Sorry, I'm new to sphinx!

keflavich:
I'm running the remote tests locally and I got a timeout error on the `get_images` test.  It might be necessary to increase the timeout for these tests?

ellieb:
Thanks for taking a look and letting me know! :+1: Is the test itself timing out? Right now the remote `get_images` test is pretty hefty, I might break it up into one regular and one heavy test. Looking at eso and alma modules, they seem to mark really slow tests and skip them, maybe I could do that with the heavy test?

keflavich:
Yes, I can't get the `get_images` test to pass at all locally w/60s timeouts.  I recommend skipping it.

The timeout that needs to be configured for this is `astropy.utils.data.conf.remote_timeout`.  Unfortunately we haven't made the file download timeout use the module-specific astroquery-configured one.

keflavich:
Remote tests now pass locally.  Merging.  We still have a todo item of making a faster `get_images` test, if possible.

PR #1548, opened 2019-09-14
GAIA: Add optional 'columns' parameter to select specific columns.

tcaram:
Hey! Some of the functions from the Gaia submodule (cone_search and query_object) are lacking a 'columns' parameter to query only for specific columns. Querying only for the desired columns would reduce the bandwidth and memory used every time you query for an object or do a cone search. My version of the code is compatible with old ones, as it adds an optional parameter that behaves just like it used to (in previous versions) if it's not set.


jcsegovia:
I agree @keflavich,  I think it is quite useful and I do not see any problem with these changes.
Thanks @tcaram.

PR #1596, opened 2019-12-03
Initial gemini archive support

olyoberdorf:
This is my initial work towards an astroquery API for Gemini's archive.  For now, I have focused on some basic observation data searches:

- search by cone
- search by target name
- search by set of parameters with validation
- search by open ended list of search terms (mimiccing the web url for archive.gemini.edu so all web searches are possible/fully futureproofed)


docs/gemini/gemini.rst
```diff
@@ -0,0 +1,122 @@
...
+                exposure_time detector_roi_setting detector_welldepth_setting  telescope   ...
+                ------------- -------------------- -------------------------- ------------ ...
+                     119.9986           Full Frame                         -- Gemini-North ...
+                     119.9983           Full Frame                         -- Gemini-North ...
+                     119.9986           Full Frame                         -- Gemini-North ...
+                     119.9983           Full Frame                         -- Gemini-North ...
+                      99.9983           Full Frame                         -- Gemini-North ...
+
+
+Observation Name Queries
```
olyoberdorf: note to self: missing underline

astroquery/gemini/core.py
```diff
@@ -0,0 +1,484 @@
...
+
+from ..query import BaseQuery
+from ..utils.class_or_instance import class_or_instance
+from ..utils import async_to_sync
+from . import conf
+
+__all__ = ['Observations', 'ObservationsClass']  # specifies what to import
+
+
+__valid_instruments__ = [
```
olyoberdorf: note to self: consider making these visible sets that users can use as arguments (i.e. instruments.NIRI)

astroquery/gemini/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+from astropy import config as _config
+
+
+class Conf(_config.ConfigNamespace):
+    """
+    Configuration parameters for `astroquery.template_module`.
+    """
+    server = _config.ConfigItem(
+        ['https://archive.gemini.edu',],
+        'Name of the template_module server to use.'
```
olyoberdorf: 'gemini', not 'template_module'

astroquery/gemini/core.py
olyoberdorf: I've decided I'm nervous about defining these options more concretely than the free form strings.  The danger is we see additional options supported server-side and the astroquery data structures will be behind what is actually live.  Tabling for now, but will revisit.

keflavich:
Looking good so far!  Is this ready for a final review?

olyoberdorf:
> Looking good so far! Is this ready for a final review?

Yeah I think I'm ready for feedback - thanks!

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+    @class_or_instance
+    def query_object(self, objectname):
+        """
+        search for Gemini observations by target on the sky.
+
+        Given a sky position and radius, returns a list of Gemini observations.
+
+        Parameters
+        ----------
+        coordinates : str or `~astropy.coordinates` object
```
keflavich: this docstring is copied from above but is no longer true.  coordinates is not an accepted parameter, nor is radius

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+            Can be used to search for data by the PI's name.
+        program_id : str, optional
+            Default None.
+            Can be used to match on program ID
+        utc_date : date or (date,date) tuple, optional
+            Default None.
+            Can be used to search for observations on a particular day or range of days (inclusive).
+        instrument : str, optional
+            Can be used to search for a particular instrument.  Valid values are:
+                'GMOS',
```
keflavich: this list appears to be generated from the `__valid_x__` stuff above; there's probably a good way to reuse those lists here.  Maybe through `f`-strings?  I'm not certain how, but there is definitely a way.

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+        # this as an independent implementation.
+        if radius is not None:
+            if isinstance(radius, (int, float)):
+                radius = radius * units.deg
+            radius = astropy.coordinates.Angle(radius)
+
+        url = "%s/jsonsummary/notengineering/NotFail" % self.server
+        if coordinates is not None:
+            coordinates = commons.parse_coordinates(coordinates)
+            url = "%s/ra=%f/dec=%f" % (url, coordinates.ra.deg, coordinates.dec.deg)
```
keflavich: Instead of building up a URL like this, you could build up a dictionary of keyword/value pairs and format them all later.  But this looks like a RESTful API, maybe?  Does the order of arguments matter?  If not, I suggest a dict refactor.

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+            url = "%s/ProgramText=%s" % (url, program_text)
+        if objectname is not None:
+            url = "%s/object=%s" % (url, objectname)
+        if raw_reduced is not None:
+            if raw_reduced not in __valid_raw_reduced__:
+                raise ValueError("Unrecognized raw/reduced setting: %s" % raw_reduced)
+            url = "%s/%s" % (url, raw_reduced)
+        response = self._request(method="GET", url=url, data={}, timeout=180, cache=False)
+
+        js = json.loads(response.text)
```
keflavich:
```suggestion
        js = response.json()
```

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+            url = "%s/%s" % (url, adaptive_optics)
+        if program_text is not None:
+            url = "%s/ProgramText=%s" % (url, program_text)
+        if objectname is not None:
+            url = "%s/object=%s" % (url, objectname)
+        if raw_reduced is not None:
+            if raw_reduced not in __valid_raw_reduced__:
+                raise ValueError("Unrecognized raw/reduced setting: %s" % raw_reduced)
+            url = "%s/%s" % (url, raw_reduced)
+        response = self._request(method="GET", url=url, data={}, timeout=180, cache=False)
```
keflavich: `timeout` should be `self.TIMEOUT` and should be configurable, with a default to 180

`cache` should _not_ be hard-coded to `False` here, but should be a user parameter, unless there is very good reason.  If there is such a reason, please add it here.

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+        if 'coordinates' in kwargs:
+            coordinates = kwargs['coordinates']
+            coordinates = commons.parse_coordinates(coordinates)
+        else:
+            coordinates = None
+
+        if coordinates is not None:
+            url = "%s/ra=%f/dec=%f" % (url, coordinates.ra.deg, coordinates.dec.deg)
+        if radius is not None:
+            url = "%s/sr=%fd" % (url, radius.deg)
```
keflavich: there must be a supported way to construct these sorts of URLs in python, but I can't find it.  Just an aside - if you can find a library to do this for you, it would be much nicer than these sequential `if` statements.

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+
+        url = "%s/jsonsummary/notengineering/NotFail" % self.server
+        for arg in args:
+            url = "%s/%s" % (url, arg)
+        for key in filter(lambda key: key not in ["coordinates", "radius"], kwargs):
+            url = "%s/%s=%s" % (url, key, kwargs[key])
+
+        response = self._request(method="GET", url=url, data={}, timeout=180, cache=False)
```
keflavich: same comments as [above](https://github.com/astropy/astroquery/pull/1596/files#diff-15c0230f6c9548d8c5a0a1b2fcdab6b9R316)

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+
+        js = json.loads(response.text)
```
keflavich:
```suggestion
     js = response.json()
```

keflavich: be cautious about this suggestion - it looks like I messed up the indentation

docs/gemini/gemini.rst
```diff
@@ -0,0 +1,123 @@
...
+                      99.9983           Full Frame                         -- Gemini-North ...
+
+
+Observation Criteria Queries
+----------------------------
+
+Additional search terms are available as optional arguments to the `~astroquery.gemini.ObservationsClass.query_criteria`
+call.  These all have default values of None, in which case they will not be considered during the search.
+
+The full API is in the api documentation, but some examples are the instrument used, such as GMOS-N, the
```
keflavich: link to the API documentation when you mention it

docs/gemini/gemini.rst
```diff
@@ -0,0 +1,123 @@
...
+                ------------- -------------------- -------------------------- ------------ ------- ...
+                          0.0        Central Stamp                         -- Gemini-North    True ...
+                          0.0           Full Frame                         -- Gemini-North    True ...
+                          0.0           Full Frame                         -- Gemini-North    True ...
+                          0.0           Full Frame                         -- Gemini-North    True ...
+                          0.0           Full Frame                         -- Gemini-North    True ...
+
+
+Observation Raw Queries
+----------------------------
```
keflavich: please match the number of `-`'s to the length of the title string

astroquery/gemini/core.py
olyoberdorf: hmm - if I made the arguments match the REST service's expectations then this would drop out and be handled by the code at the end as generic kw/value pairs.  Though the user would also have to pass the radius with the trailing 'd' to get degrees.  Otherwise the logic to convert from coordinates/radius to this special sauce has to be captured somewhere.  It could also be placed in a dictionary of handlers outside the method and then the loop could just be "if kw in handler_dict, use handler_dict[kw] to modify url, else use default .../kw=val/"

docs/gemini/gemini.rst
olyoberdorf: done - thanks!

docs/gemini/gemini.rst
olyoberdorf: hmm, I meant the reference from the earlier line but that isn't clear.  I've tried rewording it a bit.

astroquery/gemini/core.py
olyoberdorf: done with the right indent - thanks!

astroquery/gemini/core.py
olyoberdorf: another thought, I could make separate code to handle url construction for the service to at least move the complexity out of this area and keep it focused

astroquery/gemini/core.py
olyoberdorf: oops - thanks!

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+
+        Given a sky position and radius, returns a list of Gemini observations.
+
+        Parameters
+        ----------
+        coordinates : str or `~astropy.coordinates` object
+            The target around which to search. It may be specified as a
+            string or as the appropriate `~astropy.coordinates` object.
+        radius : str or `~astropy.units.Quantity` object, optional
+            Default 0.3 degrees.
```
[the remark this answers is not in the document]
olyoberdorf: that's much better, done - thanks!

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+        """ set some parameters """
+        # do login here
+        super().__init__()
+
+    @class_or_instance
+    def query_region(self, coordinates, radius):
+        """
+        search for Gemini observations by target on the sky.
+
+        Given a sky position and radius, returns a list of Gemini observations.
```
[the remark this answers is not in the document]
olyoberdorf: done

astroquery/gemini/core.py
```diff
@@ -0,0 +1,480 @@
...
+    'PROCESSED_ARC'
+]
+
+
+class ObservationsClass(BaseQuery):
+
+    server = conf.server
+
+    def __init__(self, *args):
+        """ set some parameters """
```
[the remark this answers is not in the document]
olyoberdorf: modified it to a brief description of the class purpose

astroquery/gemini/core.py
keflavich: Yeah, that's a good idea since you're reusing the URL construction in two places

olyoberdorf: ok, I've come up with something - definitely could use feedback on if that's the direction you were thinking of or if it's just making things too complicated

astroquery/gemini/core.py
olyoberdorf: see comment on following thread, tried an approach separating out the url logic

astroquery/gemini/urlhelper.py
```diff
@@ -0,0 +1,84 @@
...
+The Gemini REST API is used behind the query calls so this utility is
+used internally by those methods to construct appropriate URLs.
+"""
+import astropy
+from astropy import units
+
+from astroquery.utils import commons
+
+
+def __handle_keyword_arg__(url, key, value):
```
keflavich: None of these functions need to be `__` - sure, they're not meant for public consumption, but the `__` is meant to hide class methods and certain special things - I don't have a comprehensive knowledge of what.  These are just regular old functions.

PR #1606, opened 2019-12-25
Provide for a cleaner API for querying NASA archive

sashank27:
As per the discussions in #1351 and #1605, refactor the NASA exoplanet API to query for a specific planet (using `query_planet()`) as well as for host stellar system (using `query_star()`), using their API.
This prevents unnecessary hacking with the downloaded full table.


astroquery/nasa_exoplanet_archive/nasa_exoplanet_archive.py
```diff
@@ -38,18 +46,16 @@ def param_units(self):
 
         return self._param_units
 
-    def get_confirmed_planets_table(self, cache=True, show_progress=True,
-                                    table_path=None, all_columns=False):
+    def query_planet(self, planet_name, cache=True, show_progress=True,
+                        table_path=None, all_columns=False):
```
keflavich: minor style: make the indentation match the above line

astroquery/nasa_exoplanet_archive/nasa_exoplanet_archive.py
```diff
@@ -65,27 +71,26 @@ def get_confirmed_planets_table(self, cache=True, show_progress=True,
         Returns
         -------
         table : `~astropy.table.QTable`
-            Table of exoplanet properties.
+            Table of one exoplanet's properties.
         """
-        if self._table is None or not cache:
+
+        planet_name = '\'' + planet_name + '\''
```
keflavich: it might be a little easier to read this as `"'"`, but I'm not sure about that...

keflavich: or maybe `'"{0}"'.format(planet_name)`

astroquery/nasa_exoplanet_archive/nasa_exoplanet_archive.py
```diff
@@ -98,28 +103,70 @@ def get_confirmed_planets_table(self, cache=True, show_progress=True,
...
             Table of one exoplanet's properties.
         """
 
-        exoplanet_table = self.get_confirmed_planets_table(**kwargs)
-        return exoplanet_table.loc[planet_name.strip().lower().replace(' ', '')]
+        host_name = '\'' + host_name + '\''
+        EXOPLANET_URL_PLANET = EXOPLANETS_CSV_URL + '&where=pl_hostname=' + quote(host_name)
+
+        if self._table is not None and self._cache_star is host_name and cache:
+            pass
```
keflavich: this block would be a little clearer if you had `return self._table` here instead of `pass`ing.  Even though it repeats a line of code, it makes it clearer what this block of text is doing.

astroquery/nasa_exoplanet_archive/tests/test_nasa_exoplanet_archive_remote.py
```diff
@@ -121,17 +107,3 @@ def test_hd209458b_exoplanet_archive_coords():
     sep = params['sky_coord'].separation(simbad_coords)
 
     assert abs(sep) < 5 * u.arcsec
-
-
-@remote_data
-def test_exoplanet_archive_table_all_columns():
```
keflavich: While it makes sense to remove these now-obsolete tests, we need something to replace them.  Could you add planet-specific tests here?

sashank27: planet-specific tests are already present. Star-specific tests need to be added up. I'm working on it right now.

keflavich:
lgtm!  If you want to add a changelog entry, I'll merge this afterward, otherwise I'll merge next time I have access to command line (I'm writing this from mobile, so I can't easily add the changelog entry myself now)

sashank27:
@keflavich what about the coverage? Also, do @bsipocz also need to review this?

keflavich:
Coverage isn't an issue; we generally see reduction in coverage for any new features.  We don't generally require multiple reviews.

astroquery/nasa_exoplanet_archive/tests/test_nasa_exoplanet_archive_remote.py
```diff
@@ -14,124 +14,159 @@
...
 
-@remote_data
-def test_hd209458b_exoplanet_archive_coords():
-    params = NasaExoplanetArchive.query_planet('HD 209458 b ')
+def test_hd209458b_exoplanet_archive_coords_local():
+    params = NasaExoplanetArchive.query_planet('HD 209458 b ', table_path=LOCAL_TABLE_PATH)
     simbad_coords = SkyCoord(ra='22h03m10.77207s', dec='+18d53m03.5430s')
 
-    sep = params['sky_coord'].separation(simbad_coords)
+    sep = params['sky_coord'][0].separation(simbad_coords)
```
sashank27: @keflavich In this new change, we have to do the indexing to get the data. Is that okay?

keflavich: Yes, that's not a problem, particularly in the tests.  It might be nice to have a comment telling why we're indexing and what the underlying data structure is, but it's not necessary.

sashank27:
@keflavich @bsipocz thanks for the support!

bmorris3:
Hi all – this PR is causing backwards compatibility failures because it removes the `get_confirmed_planets_table` function. Was that intentional?

keflavich:
If I understand correctly, `get_confirmed_planets_table` retrieved the whole table?  I think this should be deprecated, but should not have simply been removed.

Can we re-implement `get_confirmed_planets_table` as a wrapped of `query_planets`, but with a deprecation warning suggesting that you should use the newer syntax?

PR #1638, opened 2020-02-06
New service: NOAO 

pothiers:
Maintainers requested early PR so this is it.  Its not really intended to be merged.  Just here so you can see what I've done and suggest changes.  There is quite a bit more I plan to add but would like to get one successfully merged PR under our belt with NOAO before I commit the more complicated stuff.

NOTE: This uses vanilla "requests" instead of BaseQuery._requests()  The signatures seem to be different so I'm not included to use the _requests version until I understand better how I should use it.  In some of my planned upcoming changes I will be using POST with JSON payload. 

It is not clear to me what that @remote_data decorator is supposed to do in the test files.  Pointers to info would be appreciated.  Since the decorator caused test failure without a good hint about what needed to be change, I removed it.


keflavich:
Thanks @pothiers !

Some answers:
(1) The `@remote_data` decorator is to ensure that tests that require an internet connection are not attempted when one is not available.  See https://astroquery.readthedocs.io/en/latest/testing.html for some details.  Basically, we are trying to avoid spamming all of the services when we run tests.

(2) The `BaseQuery._requests` toolkit adds a few features on top of python's `requests` library, such as caching and some basic login functionality.  You should use it following the examples of other modules, but ask questions as you need.  The call signature is very similar to requests, but instead of `requests.post(...)`, you do `_request('POST', ...)`.  Everything else should be the same, iirc.

pothiers:
Use of _request() causes error: `AttributeError: 'NoaoClass' object has no attribute 'cache_location'` despite NoaoClass inheriting from BaseQuery.  Using it as in: `response = self._request('GET', url)`.  Until I find resolution, reverting to use of "requests" library.

pothiers:
Now do I correct " Changelog entry not present, (or PR number missing) and..."?  How do I see the rest of the message?  Where would this have been explained?

astroquery/noao/core.py
```diff
@@ -0,0 +1,71 @@
+"""\
```
keflavich: no `\` needed here

astroquery/noao/tests/test_noao_remote.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@remote_data
+class TestNoaoClass(object):
+    @classmethod
+    def setup_class(cls):
+        cls.arch = astroquery.noao.Noao
+
+    def test_query_region_1(self):
+        """Insure query gets at least the set of files we expect.
```
keflavich: Ensure?

pothiers:
Thanks. Fixed and pushed.

pothiers:
Travis reports error:
` An attempt was made to connect to the internet by a test that was not marked remote_data. `
This despite the use of "@remote_data" decorator on the only test class in the module.

astroquery/noao/core.py
```diff
@@ -0,0 +1,78 @@
...
+    ADS_URL = f'{NAT_URL}/api/adv_search/fasearch'
+    SIA_URL = f'{NAT_URL}/api/sia/voimg'
+
+    def __init__(self, which='voimg'):
+        """ set some parameters """
+        # Change following to match current https://<root>/api/version
+        KNOWN_GOOD_API_VERSION = 2.0
+        # #!response = self._request('GET', f'{self.NAT_URL}/api/version',
+        # #!                         cache=False)
+        response = requests.get(f'{self.NAT_URL}/api/version')
```
keflavich: This is what's triggering your `remote_data` access.  You can't do a remote call in the `__init__` step, because `__init__` happens at import time.  We actually forbid that explicitly.

A good alternative might be to do something like add a `.api_version` property that performs a (cached) query to NOAO to determine the API version the first time it's called.

pothiers: Thanks the explanation.  I was unlikely to ever find that!  And I love the alternative. I'll implement.

pothiers:
Travis CI complains about sphinx warnings.  I looked for anything related to noao and didn't find anything.  Lots of other warnings though.  I'm that this point I've run out of CI things I can fix.

pothiers:
Below is the bottom of Travis log.  Note "build succeeded", yet "travis build FAILED".

```
build succeeded, 14 warnings.

The HTML pages are in _build/html.
The build_docs travis build FAILED because sphinx issued documentation warnings (scroll up to see the warnings).
travis_time:end:0058ecd8:start=1581112871588920934,finish=1581112911700417722,duration=40111496788,event=script
[0K[31;1mThe command "$MAIN_CMD $SETUP_CMD" exited with 1.[0m


Done. Your build exited with 1.
```

keflavich:
The Sphinx warnings are frustratingly difficult to track down, imo.  

Strangely, though, I don't see any sign of any noao-related warnings in the travis report; I wonder if this is a problem in master.  In any case, don't worry about that particular failure for the moment.

bsipocz:
Yes, the warning is coming from the missing intersphinx of mocpy. Not sure what happened with it. 
As @keflavich said, ignore that failure for now.

bsipocz:
(Oh, and the trick for the sphinx warnings is to case match search for `WARNING`. And to have seen tons of sphinx issues to know what the particular issue with it. I haven't identified a shortcut to this 🤷‍♀ )

pothiers:
I don’t understand “ignore the failure”. The CI fails which means PR doesn’t get merged, right ?

-sp-

keflavich:
Yes, that's right, but at this point we know the source of the error is something else.  @bsipocz just fixed it in #1639, so you didn't have to take any actions.  Anyway, now that it *is* fixed, you should rebase against master - that should fix the travis build.

astroquery/noao/core.py
```diff
@@ -0,0 +1,91 @@
...
+Provide astroquery API access to OIR Lab Astro Data Archive (natica).
+
+This does DB access through web-services. It allows query against ALL
+fields in FITS file headers.
+
+Also possible but not provided here: query through Elasticsearch DSL.  ES will
+be much faster but much more limitted in what can be used in query.  """
+
+# Python library
+# External packages
```
keflavich: you can remove these comments now

astroquery/noao/core.py
```diff
@@ -0,0 +1,91 @@
...
+import requests
+import astropy.table
+# Local packages
+
+# 3. local imports - use relative imports
```
keflavich: similarly, all these comments - they're leftovers from the template, and just provide instructions to you as the developer

astroquery/noao/core.py
```diff
@@ -0,0 +1,91 @@
...
+            self.url = f'{self.NAT_URL}/api/sia/vohdu'
+        if which == 'voimg':
+            self.url = f'{self.NAT_URL}/api/sia/voimg'
+        else:
+            self.url = f'{self.NAT_URL}/api/sia/voimg'
+
+    @property
+    def api_version(self):
+        if self._api_version is None:
+            response = requests.get(f'{self.NAT_URL}/api/version')
```
keflavich: this needs to be `self._request('GET', ...)`, not `requests.get`

astroquery/noao/core.py
```diff
@@ -0,0 +1,91 @@
...
+            # Following gets error:
+            #   AttributeError: 'NoaoClass' object has
+            #      no attribute 'cache_location'
+            # Don't see documenation saying what that should be.
```
keflavich: See https://github.com/astropy/astroquery/blob/master/astroquery/query.py#L166

astroquery/noao/core.py
```diff
@@ -0,0 +1,91 @@
...
+                   f'{self.api_version} from the API.')
+            raise Exception(msg)
+
+    @class_or_instance
+    def query_region(self, coordinate, radius='1'):
+        self.validate_version()
+        ra, dec = coordinate.to_string('decimal').split()
+        size = radius
+        url = f'{self.url}?POS={ra},{dec}&SIZE={size}&format=json'
+        response = requests.get(url)
```
keflavich: same here - this has to be `self._request('GET', ....)`.   You should decide whether the `cache` will be used for this query - generally, for user queries, you want to set `cache=True` by default, or at least let the user specify.  For "setup" queries, like those used to get cookies for a session, you want to set `cache=False`

pothiers: As mentioned in code coments, using self._request instead of "request" gives me error that I cannot resolve. "AttributeError: 'NoaoClass' object has no attribute 'cache_location'" since I couldn't find any documentation telling me what "cache_location" should be.  Any pointer?

pothiers: Found clue in other module.  Needed to do super init  (to get BaseQuery init)

pothiers:
CI fails but seems to be be due to modules that are not mine.

astroquery/noao/core.py
```diff
@@ -0,0 +1,70 @@
...
+                   f'{self.api_version} from the API.')
+            raise Exception(msg)
+
+    @class_or_instance
+    def query_region(self, coordinate, radius='1'):
+        self.validate_version()
+        ra, dec = coordinate.to_string('decimal').split()
+        size = radius
+        url = f'{self.url}?POS={ra},{dec}&SIZE={size}&format=json'
+        response = self._request('GET', url, cache=True)
```
keflavich: Instead of `cache=True` here, put it above (line 57) in `query_region`, and pass the keyword argument on to `_request`.  You should also pass the `TIMEOUT` parameter.  This is done in other modules, for example:
https://github.com/astropy/astroquery/blob/master/astroquery/alma/core.py#L149-L151

astroquery/noao/core.py
```diff
@@ -0,0 +1,70 @@
...
+            self.url = f'{self.NAT_URL}/api/sia/voimg'
+
+        super().__init__()
+
+    @property
+    def api_version(self):
+        if self._api_version is None:
+            response = self._request('GET',
+                                     f'{self.NAT_URL}/api/version',
+                                     cache=True)
```
keflavich: this is a case where I suspect you do _not_ want to cache, since you want to check the remote API version each time you query.  I'm not sure, though.

pothiers: Nope.  As long as someone doesn't create an instance of NOAO class and keep it up for days, it will be fine.  Even then, unlikely to get tripped up.

astroquery/noao/core.py
pothiers: done, good point

pothiers:
The **success/failure reported by CI is different answers in different places**. 

SUCCESS: https://travis-ci.org/astropy/astroquery/requests?utm_medium=notification&utm_source=github_status

FAILURE: https://travis-ci.org/astropy/astroquery/builds/654598245?utm_source=github_status&utm_medium=notification
The Failure is the BlockingIOError that I see is being worked on and not noao module specific.

bsipocz:
yes, the failure is known and unrelated for this PR, so just ignore it for now.

pothiers:
It seems this is still open and missed the intended release.  I was told something else was holding it up and I didn't have to do anything.  The only CI error is related to blocking which is apparently in cds (not my additions).  What's the scope?

bsipocz:
0.4.1 hasn't been released, so this hasn't missed a release.

This still needs a final review.

astroquery/noao/core.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, which='voimg'):
+        """ set some parameters """
+        self._api_version = None
+
+        if which == 'vohdu':
+            self.url = f'{self.NAT_URL}/api/sia/vohdu'
+        if which == 'voimg':
+            self.url = f'{self.NAT_URL}/api/sia/voimg'
+        else:
+            self.url = f'{self.NAT_URL}/api/sia/voimg'
```
keflavich: this whole block reads:
`self.url = f'{self.NAT_URL}/api/sia/voimg'`
no matter what.  I think you meant the middle `if` to be an `elif`?

The documentation here needs to be clearer; what does `which` mean?

astroquery/noao/core.py
```diff
@@ -0,0 +1,74 @@
...
+        if (int(self.api_version) - int(KNOWN_GOOD_API_VERSION)) >= 1:
+            msg = (f'The astroquery.noao module is expecting an older version '
+                   f'of the {self.NAT_URL} API services.  '
+                   f'Please upgrade to latest astroquery.  '
+                   f'Expected version {KNOWN_GOOD_API_VERSION} but got '
+                   f'{self.api_version} from the API.')
+            raise Exception(msg)
+
+    @class_or_instance
+    def query_region(self, coordinate, radius='1', cache=True):
```
keflavich: `radius` here should be a quantity, not a string.

astroquery/noao/core.py
```diff
@@ -0,0 +1,74 @@
...
+        self.validate_version()
+        ra, dec = coordinate.to_string('decimal').split()
+        size = radius
```
keflavich: this seems unnecessary

astroquery/noao/core.py
```diff
@@ -0,0 +1,74 @@
...
+        url = f'{self.url}?POS={ra},{dec}&SIZE={size}&format=json'
+        response = self._request('GET', url,
+                                 timeout=self.TIMEOUT,
+                                 cache=cache)
```
keflavich: might want a `reponse.raise_for_status()` here?

astroquery/noao/core.py
```diff
@@ -0,0 +1,74 @@
...
+        return astropy.table.Table(data=response.json())
+
+    def _args_to_payload(self, *args):
+        # convert arguments to a valid requests payload
+        return dict
```
keflavich: this isn't valid code - do you mean to return `args` or `kwargs` or something like that?  I'm not clear what the intent is here, though it seems you're not using this at all and could just remove the function

docs/noao/noao.rst
```diff
@@ -0,0 +1,18 @@
+.. doctest-skip-all
+
+.. _astroquery.noao:
+
+********************************
+NOAO Queries (`astroquery.noao`)
+********************************
+
+Getting started
+===============
```
keflavich: Need some documentation here!

astroquery/noao/core.py
pothiers: Good catch. Fixed locally for my next push.

astroquery/noao/core.py
pothiers: Its going into a URL query string so doesn't matter.  But changed locally for next push.

astroquery/noao/core.py
pothiers: Not sure what is uncessary.  The astroquery API says "coordinate" is required.  I'd prefer separate ra,dec args if that's what you meant and if its allowed

astroquery/noao/core.py
pothiers: Good point. Done locally for next push.

astroquery/noao/core.py
pothiers: removed

docs/noao/noao.rst
pothiers: I would like to be able to skate through with minimal documentation for this release.  Yes, this is less than minimal.  We have more documentation online so I'll likely point to that if I can figure out how to add links to these docs.  Will you let minimal pass?  I have another version in the works that will add SIGNIFICANTLY MORE search capabilities.  That would be a good time to document heavily.

keflavich: yes, minimal is acceptable as a first pass, but at least a few basic examples - minimum one example per function - is needed.

astroquery/noao/core.py
keflavich: I just mean renaming `radius` to `size` here - why not just use `radius` below?

keflavich:
@pothiers  it looks like your last rebase didn't work!  Let's try to figure out what went wrong via slack

astroquery/noirlab/tests/test_noirlab_remote.py
```diff
@@ -0,0 +1,35 @@
...
+        cls.arch = Noirlab(which='voimg')
+
+    def test_query_region_1(self):
+        """Ensure query gets at least the set of files we expect.
+        Its ok if more files have been added to the remote Archive."""
+
+        c = SkyCoord(ra=10.625*u.degree, dec=41.2*u.degree, frame='icrs')
+        r = self.arch.query_region(c, radius='0.1')
+        actual = set(list(r['md5sum'])[1:])
+        print(f'DBG: query_region_1; actual={actual}')
```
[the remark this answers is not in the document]
pothiers: Added test. Commented out print statements. Passing tests don't show those prints but failing ones do. Makes it easy to overlook prints that are in their just so I can get my "expected results". I was using an obsolete version of my "which" string.  Now use correct version and do it for each value.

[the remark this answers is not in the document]
keflavich: You know, you're right, the print statements on failed tests are useful - if you want to keep print statements here, that's OK.

astroquery/noirlab/core.py
```diff
@@ -0,0 +1,72 @@
...
+        if (int(self.api_version) - int(KNOWN_GOOD_API_VERSION)) >= 1:
+            msg = (f'The astroquery.noirlab module is expecting an older '
+                   f'version of the {self.NAT_URL} API services. '
+                   f'Please upgrade to latest astroquery.  '
+                   f'Expected version {KNOWN_GOOD_API_VERSION} but got '
+                   f'{self.api_version} from the API.')
+            raise Exception(msg)
+
+    @class_or_instance
+    def query_region(self, coordinate, radius=1, cache=True):
```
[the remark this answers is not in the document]
keflavich: This isn't resolved yet - there's still no docstring.  docstrings are the inline documentation, e.g.:
```
def f(blah):
     """
     This is a docstring

    Parameters
    -----------
    blah : str
        Describe parameter here
    """
```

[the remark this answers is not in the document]
pothiers: I think you are talking about _validate_version.  It starts with underscore so I think that means its intended as private.  You asked for docstrings on public methods.  _validate_version is not public.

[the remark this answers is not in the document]
keflavich: this inline comment on github is on the line `def query_region(self, coordinate, radius=1, cache=True):`; you should be able to see that clearly on the github interface.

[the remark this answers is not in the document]
keflavich: it should look like this:
image

astroquery/noirlab/core.py
```diff
@@ -25,7 +25,10 @@ class NoirlabClass(BaseQuery):
     def __init__(self, which='file'):
         """Return object used for searching the NOIRLab Archive.
 
-        Search either for Files (which=file) or HDUs (which=hdu)."""
+        Search either Files (which=file) or HDUs (which=hdu).
+        Files will always be returned.  But if which=hdu, 
+        the
```
keflavich, line 30: Looks like you committed this halfway done?

pothiers: Good catch, fixed.

CHANGES.rst
```diff
@@ -45,8 +45,16 @@ alma
 Infrastructure, Utility and Other Changes and Additions
 -------------------------------------------------------
 
+0.4.1 (2020-02-07)
+================
```
keflavich, line 49: Could you make this change?  We didn't release 0.4.1 on Feb 7.  Let us set the release time later.
```suggestion
0.4.1 (unreleased)
==================
```

keflavich: also, you should be able to make this change through the github interface

astroquery/noirlab/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+The NSF's OIR Lab Astro Data Archive (formerly NOAO Science Archive)
+provides access to data taken with more than 40 telescope and
+instrument combinations, including those operated in partnership with
+the WIYN, SOAR and SMARTS consortia, from semester 2004B to the
+present. In addition to raw data, pipeline-reduced data products from
+the DECam, Mosaic and NEWFIRM imagers are also available, as well as
+advanced data products delivered by teams carrying out surveys and
+other large observing programs with NSF OIR Lab facilities.
+
+Total holdings 9,634,110  V.3
```
bsipocz, line 15: is there a way to auto generate this table? I also feel that being hidden in the this docstring (that is exposed nowhere) is not the ideal place for it, and it should be moved to the narrative docs.

Keeping the Acknowledgement here on the other hand is preferred. Or in fact, you can even repeat it in a section in the narrative docs, too.

astroquery/noirlab/tests/test_noirlab_remote.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_query_region_1(self):
+        """Search FILES.
+        Ensure query gets at least the set of files we expect.
+        Its ok if more files have been added to the remote Archive."""
+
+        c = SkyCoord(ra=10.625*u.degree, dec=41.2*u.degree, frame='icrs')
+        r = Noirlab(which='file').query_region(c, radius='0.1')
+        actual = set(list(r['md5sum']))
+        expected = expsia.query_region_1
```
bsipocz, line 30: I feel that the expected dictionaries could be in this file, rather than separate, but it's super minor

astroquery/noirlab/tests/test_noirlab_remote.py
```diff
@@ -0,0 +1,42 @@
...
+# External packages
+from astropy import units as u
+from astropy.coordinates import SkyCoord
+from astropy.tests.helper import remote_data
+# Local packages
+from .. import Noirlab
+from . import expected as expsia
+# #!import pytest
+
+# performs similar tests as test_module.py, but performs
```
bsipocz, line 13: comments could be cleaned up

docs/noirlab/noirlab.rst
```diff
@@ -0,0 +1,73 @@
...
+.. code-block:: python
+                
+    >>> from astroquery.noirlab import Noirlab
+    >>> import astropy.coordinates as coord
+    >>> from astropy import units as u
+    >>> from astropy.coordinates import SkyCoord
+    >>> coord = SkyCoord(ra=10.625*u.degree, dec=41.2*u.degree, frame='icrs')
+
+    >>> noirlab_file = Noirlab(which='file')
+    >>> results_file = noirlab_file.query_region(c, radius='0.1')
```
bsipocz, line 37:
```suggestion
    >>> results_file = noirlab_file.query_region(coord, radius='0.1')
```

docs/noirlab/noirlab.rst
```diff
@@ -0,0 +1,73 @@
...
+                                                                     string     string ...                           string
+    /net/archive/mtn/20151027/kp4m/2015B-2001/k4m_151028_085849_ori.fits.fz 2015-10-28 ... 2020-02-09T01:17:17.842642+00:00
+    /net/archive/mtn/20151027/kp4m/2015B-2001/k4m_151028_091327_ori.fits.fz 2015-10-28 ... 2020-02-09T01:17:17.875407+00:00
+    /net/archive/mtn/20151027/kp4m/2015B-2001/k4m_151028_084112_ori.fits.fz 2015-10-28 ... 2020-02-09T01:17:18.303911+00:00
+    /net/archive/mtn/20151120/kp4m/2015B-2001/k4m_151121_033641_ori.fits.fz 2015-11-21 ... 2020-02-09T01:24:35.525861+00:00
+    /net/archive/mtn/20151120/kp4m/2015B-2001/k4m_151121_031258_ori.fits.fz 2015-11-21 ... 2020-02-09T01:24:37.873559+00:00
+    /net/archive/mtn/20151120/kp4m/2015B-2001/k4m_151121_041031_ori.fits.fz 2015-11-21 ... 2020-02-09T01:24:38.951230+00:00
+    
+    >>> noirlab_hdu = Noirlab(which='hdu')
+    >>> results_hdu = noirlab_hdu.query_region(c, radius='0.1')
```
bsipocz, line 51:
```suggestion
    >>> results_hdu = noirlab_hdu.query_region(coord, radius='0.1')
```

docs/noirlab/noirlab.rst
```diff
@@ -0,0 +1,73 @@
+.. doctest-skip-all
+.. # To render rst files to HTML: python setup.py build_docs
```
bsipocz, line 2: cleanup comments

astroquery/noirlab/core.py
```diff
@@ -0,0 +1,101 @@
...
+        if (int(self.api_version) - int(KNOWN_GOOD_API_VERSION)) >= 1:
+            msg = (f'The astroquery.noirlab module is expecting an older '
+                   f'version of the {self.NAT_URL} API services. '
+                   f'Please upgrade to latest astroquery.  '
+                   f'Expected version {KNOWN_GOOD_API_VERSION} but got '
+                   f'{self.api_version} from the API.')
+            raise Exception(msg)
+
+    @class_or_instance
+    def query_region(self, coordinate, radius=0.1, cache=True):
```
bsipocz, line 70: with having the `@async_to_sync` decorator on the class this should be an async method. What do you say @keflavich?

keflavich: Yes, let's just verify that it works as expected

bsipocz: other option, we merge this PR as is and address this issue in the next one that builds on top of this anyway?

keflavich: Either way is fine.  But thanks for catching this - I'd like to keep the async methods available.

bsipocz:
@pothiers - after some discussions we decided to merge this and bring over the open questions to the other PR. I fix my minor comments directly here before merging, and the async/sync issue can be addressed then in the other PR.

bsipocz:
(I also squash together the changelog and pep8 commits as a bit of a cleanup at rebase).

bsipocz:
Thanks @pothiers!

PR #1644, opened 2020-02-11
ALMA: Update to new ui

keflavich:
This PR implements a possible solution to #1640, though the urgency for it may have dropped.

Critical features from Alisdair Manning include:
* `<alma_url>/rh/data/expand/<uid>` returns json metadata about the UID now
* Data can be retrieved directly from `<alma_url>/dataPortal/<filename>`


astroquery/alma/core.py
```diff
@@ -16,7 +16,7 @@
 from six.moves.urllib_parse import urljoin, urlparse
 from six import iteritems, StringIO
 import six
-from astropy.table import Table, Column
+from astropy.table import Table, Column, vstack as table_vstack
```
bsipocz: why do you call it `table_vstack`? we use numpy vstack at one place, and even there we call it `np.vstack`

astroquery/alma/core.py
```diff
@@ -236,147 +236,31 @@ def _get_dataarchive_url(self):
...
+                                          for name in table['name']],
+                                    name='URL'))
+            tables.append(table)
+
+        if len(tables) == 0:
+            raise ValueError("No valid UIDs supplied.")
+
+        table = table_vstack(tables)
 
-        Parameters
```
bsipocz: keep the docstrings

bsipocz:
This will also need a changelog even though the changes doesn't seem to be user facing, they will certainly recognise that the previous astroquery version is not working once the servers are updated.

astroquery/alma/__init__.py
```diff
@@ -20,6 +20,14 @@ class Conf(_config.ConfigNamespace):
          'https://beta.cadc-ccda.hia-iha.nrc-cnrc.gc.ca'],
         'The ALMA Archive mirror to use.')
 
+    # archive_url = _config.ConfigItem(
```
bsipocz: I don't think we need this committed into the codebase after this is merged?

astroquery/alma/core.py
```diff
@@ -236,7 +233,73 @@ def _get_dataarchive_url(self):
...
+        table = table_vstack(tables)
+
+        return table
+
+    def stage_data_prefeb2020(self, uids):
+        """
+        Stage ALMA data - old server style
+
+        NOTE: this method will be removed when a new ALMA service is deployed
+        in March 2020
```
bsipocz: will we have a follow-up PR to do the cleanup?

astroquery/alma/core.py
```diff
@@ -262,6 +325,12 @@ def stage_data(self, uids):
         .DEBUG: Data list URL: https://almascience.eso.org/rh/requests/anonymous/786823226 [astroquery.alma.core]
         """
 
+        import time
```
bsipocz: why have been the imports moved down here?

astroquery/alma/core.py
```diff
@@ -514,7 +605,8 @@ def _hack_bad_arraysize_vofix(self, text):
         return b"\n".join(newlines)
 
     def _login(self, username=None, store_password=False,
-               reenter_password=False, auth_urls=['asa.alma.cl',
+               reenter_password=False, auth_urls=['2020feb.asa-test.hq.eso.org',
```
bsipocz: do we need the test server in the codebase?

astroquery/alma/core.py
```diff
@@ -400,6 +469,28 @@ def _HEADER_data_size(self, files):
 
         return data_sizes, totalsize.to(u.GB)
 
+    def _HEADER_data_size(self, files):
+        """
+        Given a list of file URLs, return the data size.  This is useful for
+        assessing how much data you might be downloading!
+        (This is discouraged by the ALMA archive, as it puts unnecessary load
```
bsipocz: is there then an API they prefer?

astroquery/alma/core.py
```diff
@@ -1105,3 +1063,27 @@ def unique(seq):
...
     return filter(lambda x: x in string.printable, s)
+
+
+def uid_json_to_table(jdata,
+                      productlist=['ASDM', 'PIPELINE_PRODUCT',
+                                   'PIPELINE_PRODUCT_TARFILE',
+                                   'PIPELINE_AUXILIARY_TARFILE']):
+    rows = []
+
+    def flatten_jdata(jj):
```
bsipocz: nitpicking but would be nice to use something meaningful rather than `jj` and `kk`. Arguably these are still way much better than single letter vars 😅

astroquery/alma/core.py
keflavich: eh, maybe, let's see if they actually make the upstream change first =)

astroquery/alma/__init__.py
keflavich: you caught me in the middle of a series of commits.  My current solution might be better long term.... ???

astroquery/alma/core.py
keflavich: because I want to delete them when we get rid of `stage_data_prefeb2020` - it's a memory aid.  Also, because `stage_data_prefeb2020` is a collection of hacks, and I want to remind ourselves that this collection of hacks should be burnt in a glorious pyre

astroquery/alma/core.py
keflavich: huh... now in this case... hm.  I've refactored.  Unclear.

astroquery/alma/core.py
keflavich: :shrug: this is old code... I think... why is it green here....

astroquery/alma/core.py
keflavich: They are much better!  You can `s/jj/blahblah/` but you cannot `s/j/blah/`

astroquery/alma/__init__.py
bsipocz: yeah, realized, but only after I submitted the review

astroquery/alma/core.py
keflavich: I'm really confused how `_HEADER_data_size` got duplicated in one of the commits...

astroquery/alma/core.py
keflavich: I don't know.   But I think it's useful to keep them distinct

keflavich:
@bsipocz I think I'm finished hacking for now.   We can get rid of those test_urls before merging.  Leaving the code around to include them isn't crazy, though...

Also... though... sorry, I'm not done hacking.  I realized this breaks the user-facing API at present and doesn't need to.   Let me play a little more

astroquery/alma/core.py
bsipocz: yes, I acknowledged that 😜

keflavich: You said "arguably".  I had to argue.  QED.

astroquery/alma/core.py
bsipocz: thus I vote for keeping the numpy one as np.vstack, and this one vstack, but I agree, this is just bike shedding where there are bigger issues laying around :)

astroquery/alma/core.py
bsipocz: ok. we need a reminder issue or, just remember to clean this up, in ~6 months time or similar.

bsipocz:
Is this ready to go? There are some real pep8 issues, but even I can fix them before merging, if otherwise this is ready to go in.

keflavich:
Not yet,  Andy noted some possible issues with the latest test server and
is trying to figure out if it's upstream or not.

bsipocz:
Can you rebase, so the recent CI fixed are picked up?

keflavich:
I merged in @alipnick's changes, which look good, but they have a somewhat dangerous side-effect that I haven't found a good workaround for:

The user's plaintext password now has to be stored as a class attribute.  The only alternative I see is to copy the full login code to the `download_files` method, which introduces a pain point for users (they may have to reenter a password each time they download) and adds a lot of duplicated code that feels out of place.

@alipnick can you confirm that the download always requires this additional auth information?  I guess it was implemented to allow simpler wget commands, but the side-effect of having the user's data stored in a very accessible attribute is not ideal.

alipnick:
@keflavich, yeah... I noticed that too. It was a quick-and-dirty solution...

What if `auth` is generated and output by `_login()` and an optional input for download_files?

keflavich:
I've adopted pretty much what you suggested.  It needs testing.

fstalm:
If not done already, it could make sense to think about moving the list of authentication servers from alma/core.py/_login to alma/__init__.py so that this can be easily overridden like archive_url can.

alipnick:
@keflavich, I just tested commit d7b5d72 by running a `stage_data()` and then a `download_files()` and found two side-effects:
1) When downloading non-proprietary data, astroquery still demands a login even though it isn't necessary. Is there a way to check if data is proprietary and only demand a login in those cases?
2) Even if you have logged in, you still need to re-enter your password when you go to download the files. This becomes especially annoying if you just logged in for non-proprietary data and then you have to do it again.  I suppose if you stored your password in your keyring this wouldn't happen, but for first time users this may not be obvious.

alipnick:
@fstalm, indeed that has been done in the commits above.

keflavich:
There is a solution proposed by @fstalm, inspect the `jdata['isProprietary']` of a uid, but it only works if you're triggering the download off of metadata you've received through an astroquery query.  If you simply pass in a URL, you can't necessarily obtain those metadata... can you?

I'd like to solve the general case: you have a URL to ALMA data.  Can you determine if it is proprietary or not just from the URL?

More broadly, though, this seems like a server-side issue that shouldn't exist.  Why can't you retrieve proprietary data after logging in and retaining the cookies from the login?

keflavich:
@alipnick I'm trying very hard to get this to work, but there are several problems with the archive:

The test server is not very responsive.  I've had many failures & long delays when using the web interface.  I have gotten nothing but failures from the test server from the command line, both when accessing proprietary and non-proprietary data.

The good news is that authorization is _not_ required by the service.  However, the `stage_data` URLs were being constructed incorrectly.  We therefore do not need to worry about storing the password locally.

This PR will not be a regression, at least, but I still cannot confirm the ability to download data from the archive servers.  If you're _not logged in_, downloading _public_ data works.  When logged in, even downloading public data does not consistently work!  I frequently get errors of this variety:
```
ConnectionError: HTTPConnectionPool(host='na-test-app-2.cv.nrao.edu', port=8082): Max retries exceeded with url: /dataPortal/sso/member.uid___A001_X1290_X46.README.txt (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x121f175c0>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))
```
or, more frustrating,
```
ReadTimeout: HTTPSConnectionPool(host='almascience-dev.nrao.edu', port=443): Read timed out. (read timeout=60)
```

My test script, which can't be run as a normal test because of the required login, is:
```
from astroquery.alma import Alma; import astroquery.alma
from astropy import log
import numpy as np

alma = Alma()
alma.dataarchive_url = 'https://almascience-dev.nrao.edu'
alma.login('keflavich')
log.setLevel('DEBUG')
uid_tbl = Alma.query(payload={'project_code':'2016.1.00620.S'}, public=False)
uid = uid_tbl['Member ous id'][3]
if 'dev' in alma.dataarchive_url:
    print('staging - new way')
    staged_new = alma.stage_data([astroquery.alma.core.clean_uid(uid)])
    assert np.all(staged_new['isProprietary'])

    url = staged_new[staged_new['size'].argmin()]['URL']
    alma.download_files([url])
elif 'dev' not in alma.dataarchive_url:
    # both should yield same result
    staged_new = alma.stage_data([astroquery.alma.core.clean_uid(uid)])
    staged_old = alma.stage_data_prefeb2020([astroquery.alma.core.clean_uid(uid)])
```

When I personally run out of proprietary data (a day I hope never arrives...), I won't be able to test this at all...

keflavich:
I was able to get the dev server to _start_ downloads, but not follow them through; I suspect I'm pushing its limits with these tests.

I'm ready to declare this "done" until we encounter more errors (and there will be more errors).  @alipnick, want to give this one more try?  @bsipocz , let's merge this and see how things crumble on March 11

bsipocz:
@keflavich - my only remaining comment is to do remove the commit with the unrelated fixes for byte types. It only affects dev astropy testing, and would require more investigation, e.g I would like to hunt down whether it's something accidental upstream or only the issue in our tests (which would be somewhat surprising as why now).

keflavich:
Darn, I got caught smuggling!

(done - I rebased and dropped those commits)

bsipocz:
Haha, those issues came up in my cleanup PR, too, and I see them locally, so it  was just bad timing for smuggling :)

keflavich:
Yeah, those fixes definitely belong in a cleanup PR, not here.

I assume we're not expecting these tests to pass?

astroquery/alma/tests/test_alma_remote.py
```diff
@@ -291,3 +292,50 @@ def test_project_metadata():
...
+
+    assert 'mous_uid' in tbl.colnames
+
+    names = [x.split("/")[-1] for x in tbl[tbl['mous_uid'] == 'uid://A001/X147/X92']['URL']]
+
+    assert '2013.1.00269.S_uid___A002_X9de499_X3d6c.asdm.sdm.tar' in names
+
+
+# always skip this: it's only for testing proprietary access
+@pytest.mark.skipif('True')
```
bsipocz: now this is OK, but long term it would be nice to be controlled by an env variable (similar to the slow tests)

bsipocz:
> I assume we're not expecting these tests to pass?

yes, they will fail for now.

astroquery/alma/tests/test_alma_remote.py
keflavich: I like that idea, but I am not sure if that will even work in a pytest environment.  In practice, I had to excise this code from the tests entirely.  I'll remove this test since it's useless.

bsipocz:
OK, this horse has been beaten up quite badly, let's merge and deploy.

bsipocz:
Thanks @keflavich!

Issue #1622, opened 2020-01-24
MAST: astroquery.mast code refactor


ceb8:
Due to changes in MAST's API structure, and the need to plan for future expansion we (@ceb8  and @phicc) intend to heavily refactor the astroquery.mast codebase. These changes should be transparent to the user, but will result in fairly radical changes to the MAST module structure. I am opening this issue to solicit feedback on our proposed codebase structure, described below.

**Files with code for interacting with MAST servers (not containing user facing functions)**
- **discovery_portal.py:** Encompasses functionality to interact with the API associated with the MAST Discovery Portal. This is currently primarily encompassed by the MastClass class in core.py and is considered the "default"  as it is built into the MAST `_request` function. After this refactor this API will no longer be the default, but will be on equal footing with the newer API paradigm.
- **services.py:** Encompasses functionality to interact with MAST's newer RESTful APIs. This is the paradigm currently used by Tesscut as well as some of the catalogs (implemented in `_fabric_request` in core.py). Currently this API paradigm is secondary to the discovery_portal paradigm, after the refactor they will be on equal footing.
- **auth.py:** Encompasses functionality related to authenticating MAST users.
- **cloud.py:** Encompasses functionality related to pulling data from AWS rather than MAST's servers.
- **fpl.py:** "File path lookup" prepares paths for cloud access. This file will be left alone in the refactor as it will be made obsolete in the future based a new MAST service comes on line that provides the same functionality.
- **utils.py:** Utility code, anything that is needed in multiple places but does not fit into any of the other files.

**Files with user facing functionality**
- **observations.py:** Contains ObservationsClass functionality.
- **collections.py:** Contains CatalogsClass functionality. Call this collections rather than catalogs in preparation for future access to other products that follow the same paradigm, but are not "catalogs" per se.
- **cutouts.py:** Contains Tesscut class functionality. Call this cutouts, to allow for future addition of other MAST cutout services.

The goal with this design is to separate the general functionality that handles MAST server interactions from the API functions that offer access to specific MAST resources. This more easily allows the module as a whole to use the general components (authorization, API paradigm etc) without having to place more and more code within the already arguably too long core.py file. This also allows for the future expansion of the MAST module to include not only the current API interfaces, but more mission or otherwise use-case specific API interfaces.

PR #1645, opened 2020-02-12
MAST refactor

ceb8:
The PR encompasses the astroquery.mast refactor as described in issue https://github.com/astropy/astroquery/issues/1622. There are no operational changes in this PR. I'm adding some tests and doing the occasional bugfix, but any work that changes the user interface will be left for other PRs.

Progress of the changes are as follows:

- [x] Move authentication code into auth.py
- [x] Move cloud access code into cloud.py
- [x] Move MAST Discovery Portal API interactions into discovery_portal.py
- [x] Move MAST RESTful service API interactions into service.py
- [x] Move ObservationsClass into observations.py
- [x] Move CatalogsClass into collections.py
- [x] Move TesscutClass into cutouts.py
- [x] Use new MAST file path lookup service instead of maintaining path finding code in astroquery


ceb8:
@bsipocz The test failure message is `KeyError: 'independent_celestial_slices'` which seems to have to do with this astropy issue https://github.com/astropy/astropy/issues/10084 and the reproject package. I don't understand why this is coming up because astroquery does not appear to use the reproject package at all. The fix seems like it would be to update the reproject version, but as I don't see it anywhere in the environment, I am not sure where to look at doing that. Do you have any guidance.

ceb8:
@bsipocz @keflavich This ticket is no longer WIP and is ready for review.

bsipocz:
@ceb8 - yes, that issue is with reproject, I don't know why the new release is not picked up instead of the old, incompatible one. Ignore that failure for now, but please restart the other, now cancelled jobs is the later stages in Travis

astroquery/mast/cloud.py
```diff
@@ -0,0 +1,220 @@
...
+                path = "s3://{}/{}".format(self.pubdata_bucket, path)
+            elif full_url:
+                path = "http://s3.amazonaws.com/{}/{}".format(self._pubdata_bucket, path)
+            return path
+        except self.botocore.exceptions.ClientError as e:
+            if e.response['Error']['Code'] != "404":
+                raise
+
+        warnings.warn("Unable to locate file {}.".format(data_product['productFilename']), NoResultsWarning)
+        return None
```
keflavich, line 130: this line is redudant; `return` would do the same thing, as would leaving this line out.

astroquery/mast/cloud.py
```diff
@@ -0,0 +1,220 @@
...
+                if statinfo.st_size != length:
+                    log.warning("Found cached file {0} with size {1} that is "
+                                "different from expected size {2}"
+                                .format(local_path,
+                                        statinfo.st_size,
+                                        length))
+                else:
+                    log.info("Found cached file {0} with expected size {1}."
+                             .format(local_path, statinfo.st_size))
+                    return
```
keflavich, line 197: not needed

astroquery/mast/collections.py
```diff
@@ -0,0 +1,504 @@
...
+        -------
+        response : list of `~requests.Response`
+        """
+
+        # Put coordinates and radius into consistant format
+        coordinates = commons.parse_coordinates(coordinates)
+
+        # if radius is just a number we assume degrees
+        if isinstance(radius, (int, float)):
+            radius = radius * u.deg
```
keflavich, line 105: not sure the `isinstance` check is great here; maybe `radius = u.Quantity(radius, u.deg)` would work?  Then, no matter what data type is input - string, int, float, array - we'll get the right thing out.

astroquery/mast/collections.py
```diff
@@ -0,0 +1,504 @@
...
+        --------
+        response : list of `~requests.Response`
+        """
+
+        # if spectra is not a Table, put it in a list
+        if isinstance(spectra, Row):
+            spectra = [spectra]
+
+        # set up the download directory and paths
+        if not download_dir:
```
keflavich, line 404: might be better to have `download_dir is not None` here.  `''` will result in a False evaluation here... that may be intentional, though?

astroquery/mast/observations.py
```diff
@@ -0,0 +1,841 @@
...
+            criteria.pop('obstype', None)
+
+        # Temporarily issuing warning about change in behavior
+        # continuing old behavior
+        # grabbing the observation type (science vs calibration)
+        obstype = criteria.pop('obstype', None)
+        if obstype:
+            warn_string = ("Criteria obstype argument will disappear in May 2019. "
+                          "Criteria 'obstype' is now 'intentType', options are 'science' or 'calibration', "
+                          "if intentType is not supplied all observations (science and calibration) are returned.")
```
keflavich, line 168: Maybe change tense, since May 2019 is in the past now

astroquery/mast/cloud.py
ceb8: True. I was following the pep8 guidance about return statement consistency:
"Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function (if reachable)"

astroquery/mast/cloud.py
ceb8: This return statement is definitely needed, if it wasn't there function execution would continue to the `with ProgressBarOrSpinner` block and re-download the file even if a cached copy was found.

astroquery/mast/collections.py
ceb8: Same idea, but one less line of code, adding a unit of `u.deg` to the `coord.Angle` call on the next line takes care of it. 😄

astroquery/mast/collections.py
ceb8: I definitely want `''` to also result in false evaluation. Would it be better to spell that out (`if download_dir is in (None, ''):`)?

astroquery/mast/cloud.py
keflavich: Oh, ok - fair enough!  I didn't know about this pep8 guideline.

astroquery/mast/cloud.py
keflavich: oh, I totally missed... the rest of this function, I got blinded and thought the next section was a new function.  My bad.

astroquery/mast/collections.py
keflavich: eh, it's fine as is.

ceb8:
@bsipocz I ran all the tests individually. Three of them fail, and I verified that they are all showing the same reproject error.

astroquery/mast/auth.py
```diff
@@ -0,0 +1,123 @@
...
+
+from getpass import getpass
+
+from astropy.logger import log
+
+from ..exceptions import AuthenticationWarning
+
+from . import conf
+
+
```
bsipocz, line 22: Add an `__all__` to all new files, even if it's empty, to keep the namespace cleaner.

astroquery/mast/collections.py
```diff
@@ -0,0 +1,500 @@
...
+            base_dir = download_dir.rstrip('/') + "/mastDownload/HSC"
+
+            if not os.path.exists(base_dir):
+                os.makedirs(base_dir)
+
+            manifest_array = []
+            for spec in spectra:
+
+                if spec['SpectrumType'] < 2:
+                    data_url = 'https://hla.stsci.edu/cgi-bin/getdata.cgi?config=ops&dataset=' \
```
bsipocz, line 467: I slightly prefer multiline strings/fstrings (in parens), than line breaks and the usage of `+`. But this is super low level bike shedding, so if you prefer this it certainly can stay as is

astroquery/mast/core.py
```diff
@@ -1,299 +1,46 @@
...
 
-@async_to_sync
-class MastClass(QueryWithLogin):
+class MastQueryWithLogin(QueryWithLogin):
     """
-    MAST query class.
-
-    Class that allows direct programatic access to the MAST Portal,
-    more flexible but less user friendly than `ObservationsClass`.
+    Super class for MAST functionality (should not be called directly by users).
```
bsipocz, line 25: certainly add an `__all__` to this file, so users won't see the class even when they import core

astroquery/mast/core.py
```diff
@@ -314,37 +61,7 @@ def _login(self, token=None, store_token=False, reenter_token=False):  # pragma:
...
-        else:
-            warn_msg = ("MAST API token invalid!\n"
-                        "To make create a new API token visit to following link: " +
-                        auth_link)
-            warnings.warn(warn_msg, AuthenticationWarning)
-
-        return not info["anon"]
+        return self._auth_obj.login(token, store_token, reenter_token)
 
     @deprecated(since="v0.3.9", message=("The get_token function is deprecated, "
```
bsipocz, line 66: maybe this can be now cleaned up and removed?

astroquery/mast/core.py
```diff
@@ -379,458 +96,47 @@ def session_info(self, silent=None, verbose=None):
...
-
-        for prop, value in kwargs.items():
-            mashup_request[prop] = value
-
-        req_string = _prepare_service_request_string(mashup_request)
-        response = self._request("POST", self._MAST_REQUEST_URL, data=req_string, headers=headers,
-                                 retrieve_all=retrieve_all)
-
-        return response
+    @deprecated(since="v0.3.9", alternative="enable_cloud_dataset")
```
bsipocz, line 108: This may be a good time to remove this, too.

bsipocz: (I won't repeat this comment, but there are other deprecations, so consider removing those methods, too.

astroquery/mast/observations.py
```diff
@@ -0,0 +1,835 @@
...
+        if curl_flag:  # don't want to download the files now, just the curl script
+            manifest = self._download_curl_script(products, download_dir)
+
+        else:
+            base_dir = download_dir.rstrip('/') + "/mastDownload"
+            manifest = self._download_files(products, base_dir, cache, cloud_only)
+
+        return manifest
+
+    @deprecated(since="v0.3.9", alternative="get_cloud_uris")
```
bsipocz, line 703: consider removing these

astroquery/mast/utils.py
```diff
@@ -0,0 +1,176 @@
...
+               "Content-type": "application/x-www-form-urlencoded",
+               "Accept": "text/plain"}
+
+    response = requests.get(url, params=params, headers=headers)
+    response.raise_for_status()
+
+    return response
+
+
+def resolve_object(objectname):
```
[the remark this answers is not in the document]
ceb8: This is uses the internal MAST resolver which does call Sesame, but also a number of other services so that it can resolve names that aren't resolvable through Sesame. This is also how we ensure that users get the same results they do using MAST web interfaces.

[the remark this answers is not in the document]
bsipocz: OK to keeping it then. But consider changing the name to be more mast specific if we expose it to the user API, after all other modules do this but not with standalone functions but class methods (e.g. Simbad.query_object), etc.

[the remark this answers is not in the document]
ceb8: It's already exposed as part of the specific classes, which is not new behavior. I can not expose it from utils, which solve the issue of what should/should not be public in utils. The original idea was to try to move from having the resolver within the classes to just having it be a mast.utils function, but if that makes it too general, I can just scrap that idea. Especially since I need to keep it in the classes at least for the time being for backward compatibility.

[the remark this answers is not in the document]
bsipocz: when part of the class I think it's perfectly fine. Having mast.utils is also great as it's a good way to serve multiple of the mast classes. My only issue is with this as a standalone function exposed to the API with a too generic name for a functionality that is available through many providers.
(I don't think we're going into that direction, but one can imagine to have a `resolve_object()` for astroquery with the ability to choose the backend service, whether it's e.g. mast or simbad or whatever)

Maybe @keflavich has some thoughts about it, too.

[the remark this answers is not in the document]
ceb8: OK, then I will keep utils entirely for within mosdule use, and have the resolver function as part of the classes as the user facing one.

astroquery/mast/core.py
ceb8: Oooooooh! I was trying to figure out how to do that! Excellent.

docs/mast/mast.rst
```diff
@@ -975,4 +975,8 @@ Reference/API
 
 .. automodapi:: astroquery.mast
     :no-inheritance-diagram:
+    :inherited-members:
+
+.. automodapi:: astroquery.mast.utils
```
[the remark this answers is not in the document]
ceb8: Got rid of the utils one. But needed to leave the `:inherited-members:` because otherwise all of the functions in `MastQueryWithLogin` like "login" and "session_info" aren't documented as they need to be and also building the docs fail (I think because I refer to those functions).

ceb8:
@bsipocz  Yeah, hopefully the pain that is this PR looking like everything in this module is new will mean that we never have to do something like this with it again.

I ran all the doc examples and they  all run.

With respect to removing the deprecation, I was planning to do that in a separate PR, I wanted this one to just be re-factoring and not have an interface changes.  I can do that here though if you think it would be cleaner.

CHANGES.rst
```diff
@@ -36,6 +37,11 @@ alma
   several columns have been added [#1644]
 
 
+mast
+^^^^
+- Module refactor [#1645]
```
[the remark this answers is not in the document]
ceb8: Removed it. This PR should not have user facing affects.

bsipocz:
> With respect to removing the deprecations, I was planning to do that in a separate PR, I wanted this one to just be re-factoring and not have an interface changes.

that's a totally reasonable plan

docs/mast/mast.rst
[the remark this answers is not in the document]
bsipocz: yes, the comment was for `utils` only

ceb8:
Made an issue for the deprecation removal so I don't forget about it: https://github.com/astropy/astroquery/issues/1708

bsipocz:
Thanks @ceb8!

PR #1648, opened 2020-02-13
Solution for Issue #1646 

alipnick:
Changed metadata inputs for _login() that ALMA needs for authentication.


keflavich:
Thanks @alipnick !  Have you verified this works for you?

alipnick:
With my version of astroquery (which is admittedly older), this changed allowed me to login successfully and access proprietary data.

alipnick:
They seemed to have reordered the hidden metadata info and dropped whatever "lt" was so it ended up being a small change.

PR #1757, opened 2020-07-01
Update gaia.rst

AllenDowney:
```
>>> for column in (gaiadr1_table.get_columns()):
>>>   print(column.get_name())
```

It looks like the `get_columns()` method has been replaced with a `columns` property and `get_name()` has been replaced with `name`

So this works for me:

```
for column in table.columns:
    print(column.name)
```

### DIFF
--- a/docs/gaia/gaia.rst
+++ b/docs/gaia/gaia.rst
@@ -492,7 +492,7 @@ To remove asynchronous
 .. code-block:: python
 
   >>> from astroquery.gaia import Gaia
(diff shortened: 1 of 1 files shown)


keflavich:
Confirmed that this correction is needed.  lgtm!  Will merge when travis says OK

AllenDowney:
Sorry, I just added two commits with additional changes. Would you look again?

ceb8:
Thanks @AllenDowney!

AllenDowney:
Thank you!

PR #1776, opened 2020-07-15
Fix ErfaWarning import

keflavich:
Minor fix suggested by @pllim: astroquery is usign a hidden import that it shouldn't be!


pllim:
CI is acting weird across several repos right now. 🤷

keflavich:
Travis is green now, so I'm merging.  This does not need a changelog.  The coverage loss is not important.

PR #1788, opened 2020-08-04
XMM: Docs updated

lvalerom:
The section of _getting table details of XSA TAP_ was outdated.


keflavich:
Looks fine.  Is this something that will need to be regularly updated?

lvalerom:
No, this is not something that will need to be regularly updated. In fact, we hope that it doesn't need to be updated again. 
Thank you.

keflavich:
Tests pass, this doesn't need a milestone or changelog entry.

PR #1834, opened 2020-10-05
Change rospublic to rosmaster in HEASARC search

EfremBraun:
Fixes issue #1637


keflavich (approved):
Thanks!

Have you verified these examples now work?  If not, we should do that.

EfremBraun:
> Thanks!
> 
> Have you verified these examples now work? If not, we should do that.

Yes I have. During my verification, I saw that not all the examples wrote, which is why I wrote pull request #1835.

PR #1865, opened 2020-10-18
Comment api_key in the default config file

saimn:
Having this key defined breaks the detection of unmodified config file.

Related to (proposed) changes to `astropy.config` (https://github.com/astropy/astropy/pull/10877, https://github.com/astropy/astropy/pull/10893). If this key is not commented, the detection of unmodified config file does not work, unless the config file is strictly the same as the default one.

### DIFF
--- a/astroquery/astroquery.cfg
+++ b/astroquery/astroquery.cfg
@@ -207,4 +207,4 @@
 [astrometry_net]
 
 # The Astrometry.net API key
-api_key =
+#api_key =


bsipocz (approved):
Looks OK to me, but let's wait the sign-off from the other maintainers before merging.

keflavich (approved):

ceb8 (approved):

ceb8:
Thanks @saimn!

PR #1885, opened 2020-11-04
renamed and deprecated Observations and Observations class for the gemini module

tinumide:


CHANGES.rst
```diff
@@ -4,6 +4,14 @@
 Service fixes and enhancements
 ------------------------------
 
+observations
```
ceb8, line 7: There should be two changelog entries, one under MAST and one under Gemeni.

ceb8 (changes requested):
This is a massive breaking change for users, we need to have a deprecation period where the old syntax still works but a deprecation message is displayed. Check out https://docs.astropy.org/en/stable/api/astropy.utils.decorators.deprecated.html#astropy.utils.decorators.deprecated

ceb8:
@olyoberdorf You probably want to take a look at this too.

olyoberdorf:
Gemini is very new, but the deprecation work would be essentially identical to what's needed for MAST.  So you may as well do both.  Also, thanks for working on my module :)

CHANGES.rst
tinumide: Would it be a good idea to assign the old name of the class to the new name
 e.g `ObservationsClass = GeminiObservationsClass()`
and leave a deprecation message such as **ObservationsClass is deprecated. Use GeminiObservationsClass instead **

tinumide: > This is a massive breaking change for users, we need to have a deprecation period where the old syntax still works but a deprecation message is displayed. Check out https://docs.astropy.org/en/stable/api/astropy.utils.decorators.deprecated.html#astropy.utils.decorators.deprecated

Would it be a good idea to assign the old name of the class to the new name
`e.g ObservationsClass = GeminiObservationsClass()`
and display a deprecated message saying **ObservationsClass is deprecated. Use GeminiObservationsClass instead**

bsipocz: You can use this PR as an example for renaming and deprecating a class: https://github.com/astropy/astropy/pull/9445

tinumide: Nice! Thanks @bsipocz

astroquery/gemini/core.py
```diff
@@ -528,4 +530,11 @@ def _gemini_json_to_table(json):
         "release",
         "dec"]
 
+
+@deprecated(since='v0.4.1', alternative='GeminiObservationsClass')
```
bsipocz, line 534: we're in 0.4.2 dev cycle, so this should say deprecated since v0.4.2

CHANGES.rst
```diff
@@ -4,6 +4,18 @@
 Service fixes and enhancements
 ------------------------------
 
+gemini
+^^^^^^
+
+- Renamed and deprecated Observations and ObservationsClass in GEMINI module. [#1885]
```
bsipocz, line 10: this is nitpicking, so I only point out to show better practice: changelog should point out what has changed, and what the new solution is. No need to mention the module name in the changelog as the section above already contains that information. The backticks are to highlight the class names when rendered for the documentation.
```suggestion
- Renamed and deprecated ``Observations`` and ``ObservationsClass`` to ``GeminiObservations`` and ``GeminiObservationsClass`` . [#1885]
```

CHANGES.rst
```diff
@@ -4,6 +4,18 @@
...
+
+
+mast
+^^^^
+
+- Renamed and deprecated Observations and ObservationsClass in MAST module. [#1885]
```
bsipocz, line 16: similarly
```suggestion
- Renamed and deprecated ``Observations`` and ``ObservationsClass`` to  ``MastObservations`` and ``MastObservationsClass``. [#1885]
```

astroquery/gemini/__init__.py
```diff
@@ -18,6 +18,6 @@ class Conf(_config.ConfigNamespace):
 
 conf = Conf()
 
-from .core import ObservationsClass, Observations
+from .core import GeminiObservationsClass, GeminiObservations, Observations, ObservationsClass
 
-__all__ = ['Observations', 'ObservationsClass', 'conf']
+__all__ = ['Observations', 'ObservationsClass', 'GeminiObservations', 'GeminiObservationsClass', 'conf']
```
bsipocz, line 23: I think we don't recommend wildchar imports, so removing the old class names from the namespace is probably safe.

bsipocz (changes requested):
Overall this looks good to me, and I only have some minor comments, one of them needs input from @ceb8. 

However, the documentation should also be updated before we merge this.

bsipocz:
Oh, and as I see there are some conflicts due to a PR that has been merged in the meantime, so this will need a rebase at some point (better to do it once you've done the docs updates, too).

Also, note to maintainers that travis is gone, and I have to fix the CI before we can reasonably merge anything.

ceb8:
@bsipocz  Which of your comments did you need my input on? 

Also thanks for remembering the documentation! Super important!

bsipocz:
This needs a rebase due to conflicts. I may try to get to it before release to make sure the deprecations end up in the release.

astroquery/gemini/core.py
```diff
@@ -21,7 +22,13 @@
 from ..exceptions import AuthenticationWarning
 
 
-__all__ = ['Observations', 'ObservationsClass']  # specifies what to import
+<<<<<<< HEAD
```
bsipocz, line 25: there are some remnants here from the conflict resolutions

astroquery/gemini/core.py
```diff
@@ -526,4 +533,11 @@ def _gemini_json_to_table(json):
         "release",
         "dec"]
 
+
+@deprecated(since='v0.4.2', alternative='GeminiObservationsClass')
```
bsipocz, line 537: this is now 0.4.3

astroquery/mast/observations.py
```diff
@@ -852,5 +851,11 @@ def service_request_async(self, service, params, pagesize=None, page=None, **kwa
         return self._portal_api_connection.service_request_async(service, params, pagesize, page, **kwargs)
 
 
+@deprecated(since='v0.4.2', alternative='MastObservationsClass')
```
bsipocz, line 854: 0.4.3

bsipocz (commented):
Some tiny comments. Fixing these and the pep8 issues should make this ready to go I think

bsipocz:
Good news, I see the docs build error on this branch locally, too. Bad news, I don't see the reason, the same build passes.

bsipocz:
And I spotted one more issue, if you do an import the deprecation warnings should not show. I only have an ugly workaround for the class instance and I'm not sure how I could force it to emit a deprecation when used, but not when the module is imported.
```
Python 3.9.1 (default, Jan 11 2021, 15:50:22) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.20.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import astroquery.mast
WARNING: AstropyDeprecationWarning: The ObservationsClass class is deprecated and may be removed in a future version.
        Use MastObservationsClass instead. [astroquery.mast.observations]
```

Maybe @pllim you have a nice idea to work this around?

astroquery/mast/__init__.py
```diff
@@ -30,13 +30,13 @@ class Conf(_config.ConfigNamespace):
 
 conf = Conf()
 
-from .cutouts import TesscutClass, Tesscut, ZcutClass, Zcut
```
bsipocz, line 33: Keep this line unchanged, that will fix the docs build.

astroquery/gemini/core.py
```diff
@@ -526,4 +527,11 @@ def _gemini_json_to_table(json):
...
         "dec"]
 
+
+@deprecated(since='v0.4.3', alternative='GeminiObservationsClass')
+class ObservationsClass(GeminiObservationsClass):
+    pass
+
+
+GeminiObservations = GeminiObservationsClass()
 Observations = ObservationsClass()
```
pllim, line 537: This line is causing the warning. I guess the brute force way is to silence the warning just for this line...?

bsipocz: Yes, the issue is that if I put a filter in ``__init__.py``, then we rightly don't get the warning at module import time. However, then we don't get it at all for any usage of `Observations`, and that's what probably users used the most as the docs is full of `Observations.xzy()`.

bsipocz:
I'm sorry, but this has to wait for the next release, I just can't find a way to have all our cases covered, to make sure all users who use the old names to see the warning, but those using the new one don't.

bsipocz:
Heads up @ceb8 @jaymedina @tomdonaldson @olyoberdorf: I would like to go ahead with this sooner rather than later. Would you be still on board? Basically we already followed the same logic with the new class in mast, calling it e.g. `MastMissions`.

It is understood that we'll need long deprecation period, for both modules so for a while all the users will experience is a deprecation warning at the time of import.

As I see the last comments in here, this got stalled with some issues around getting the right warning in all situations. Besides fixing that, do you see any blockers with the proposed idea?

olyoberdorf:
Hi Brigitta,

Yeah still on board.  I don’t recall any other problems.  I’m on vacation
but I am back this weekend.

Issue #1740, opened 2020-06-15
ALMA query : "product files" not searched and ReadTimeout problem


mjastro:
Hi,

 when I tried to download all relevant files for the following ALMA project with Alma.query, I was not able to get 'product' tar files (which are accessible through the original archive portal). Only raw data are downloaded. I was wondering whether this is a problem with the following searching method or not. If I need to specify more options or need to take different commands, could you please tell me the necessary commands to download all raw+product tar files? (I am using astroquery version of 0.4.1.dev0)

```
test=Alma.query(payload={'project_code':'2011.0.00064'}) 
uids=np.unique(test['Member ous id'])
Alma.retrieve_data_from_uid(uids)
```

 Another issue that I have with the above commands is I get the following error message from time to time when downloading the files. When I got this error, I just reran `Alma.retrieve_data_from_uid(uids)`

and sometimes it resumes downloading.

`ReadTimeout: HTTPSConnectionPool(host='almascience.eso.org', port=443): Read timed out. (read timeout=60)`

 Can you comment or solve this problem as well? Thank you in advance!

keflavich:
Should that be `2011.0.00064.S`?

Try 
```
files = Alma.stage_data(uids, expand_tarfiles=True)
downloaded = Alma.download_files(files['URL'])
```

I think this is an issue specific to cycle 0 data, in which the data products distributed are labeled differently from other cycles.

The timeout error is an upstream issue.  You can try increasing the `Alma.TIMEOUT` value, but likely if it doesn't work in 60s, there is simply a connection problem with the ESO servers.

mjastro:
> Should that be 2011.0.00064.S?

Oh, yes, it should be with '.S'.

 Downloading worked with the expand_tarfiles option and by modifying TIMEOUT, thank you so much. I'll close it.

alipnick:
I'm having this same issue (along with Issue #1737) with the astroquery 0.4.1.
```python
from astroquery.alma import Alma
alma = Alma()
# Must deal with trailing / character being added to url
alma.dataarchive_url = 'https://almascience.nrao.edu'
links = alma.stage_data('uid://A001/X133d/X3b63')
links['URL']
```
The returned links are only the raw (ASDM) and auxilliary files. If I use `expand_tarfriles=True` I can see all the image files as well, but I actually want the large tarfile rather than the individual FITS files.

keflavich:
With the current version of astroquery,  I get:
```python
In [6]: astroquery.__version__
[...]

Issue #1737, opened 2020-06-09
HTTPError: 404 Client Error: for ALMA astroquery


mjastro:
Hi, 

recently somehow alma astroquery doesn't work properly with the following commands,
```
Alma.retrieve_data_from_uid(uids[0])
Alma.stage_data(uids[0]) 
```

all give the following error message :

```
HTTPError: 404 Client Error:  for url: https://almascience.eso.org/rh/submission/rh
```

"uids" is obtained by doing 
```
uids =  np.unique(test['Member ous id']), 
```
where 
```
test =Alma.query(**specified keywords**)
```

I wonder how I can fix the problem. I am using astroquery 0.4. Severl months ago, with 0.3, it certainly worked though... Thank you in advance for your help!

keflavich:
Could you please specify what `**specfied_keywords**` are?  We can't debug this without knowing what parameters cause the problem.

The ALMA remote tests are presently (mostly) passing, so unless the ESO mirror was down, there's no obvious issue here.

Make sure you're on the latest version of astroquery - if you're using a version more than about a month old, it is likely that the ALMA archive interface has changed and that is the source of your problem.

mjastro:
Thanks, I thought that I was on the very latest version, but I was not (0.4). With the 0.4.1.dev0, it worked well. I think this can be closed.

alipnick:
I'm having this issue with astroquery version 0.4.1. 
```python
>>> from astroquery.alma import Alma
>>> alma = Alma()
>>> links = alma.stage_data('uid://A001/X133d/X3b63')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/users/alipnick/.local/lib/python3.6/site-packages/astroquery/alma/core.py", line 281, in stage_data
    req.raise_for_status()
  File "/opt/local/stow/Python3.6/lib/python3.6/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://almascience.nrao.edu/asax//rh/data/expand/uid___A001_X133d_X3b63
```
The Traceback shows that the url has an extra "/" after "asax" that should not be there. Resetting the Alma.dataarchive_url seems to fix the issue:
```python
>>> alma.dataarchive_url = 'https://almascience.nrao.edu'
>>> links = alma.stage_data('uid://A001/X133d/X3b63')
>>>
```

[...]

PR #1901, opened 2020-11-19
Updates to alma/core.py to fix Issues #1737, #1740

alipnick:
A missing comma in AlmaClass.stage_data() prevented products files from being added to the results table when expand_tarfiles=False. This fixes Issue #1740. 

AlmaClass._get_dataarchive_url() returns an unusable url due to a trailing "/asax/" added on to the url in the _request() call. This needs to be removed. In the future, the trailing characters will likely change to "/aq/" as that is the ALMA standard. This change fixes that issue (Issue #1737).

### DIFF
--- a/astroquery/alma/core.py
+++ b/astroquery/alma/core.py
@@ -450,7 +450,7 @@ class AlmaClass(QueryWithLogin):
                 response.raise_for_status()
                 # Jan 2017: we have to force https because the archive doesn't
                 # tell us it needs https.
-                self.dataarchive_url = response.url.replace("http://", "https://")
+                self.dataarchive_url = response.url.replace("/asax/","").replace("/aq/","").replace("http://", "https://")
             else:
                 self.dataarchive_url = self.archive_url
         elif self.dataarchive_url in ('http://almascience.org',
@@ -531,7 +531,7 @@ class AlmaClass(QueryWithLogin):
                 else:
                     table = uid_json_to_table(jdata,
                                               productlist=['ASDM',
-                                                           'PIPELINE_PRODUCT'
+                                                           'PIPELINE_PRODUCT',
                                                            'PIPELINE_PRODUCT_TARFILE',
                                                            'PIPELINE_AUXILIARY_TARFILE'])
                 table['sizeInBytes'].unit = u.B


keflavich:
Do the extra `/`'s really matter?  I wasn't aware that could ever cause issues.

Could you clarify also, is there supposed to be _no_ `asax` or `aq` at the end of the URL?  That's what the PR presently does.

Anyway, I think the PR looks fine, I just want to verify that it will actually fix the issues noted before we merge it.  

Thanks @alipnick !

alipnick:
Hi @keflavich,
Thanks for taking a look. Indeed, without removing the `/asax/` (or `/aq/` if that returns in the future), `AlmaClass.stage_data()` doesn't work because the `dataarchive_url` parameter gets set as something like `https://almascience.nrao.edu/asax/` and the website 

`https://almascience.nrao.edu/asax//rh/data/expand/[ANY_MOUS_HERE]`

does not exist. Even if you remove the extra `/`, the website doesn't exist. The web address must be in the form:

`https://almascience.nrao.edu/rh/data/expand/[ANY_MOUS_HERE]`

So the returned `dataarchive_url` cannot have the trailing characters. I have a very old version of astroquery running with my software which does remove the `/aq/` so at some point that line was updated and dropped the removal of those characters for some reason. The only reason I didn't notice this error for a long time is because I manually set AlmaClass.archive_url and/or AlmaClass.dataarchive_url which skips the step of finding the `dataarchive_url` altogether.

alipnick:
Here's an example to try. Broken link with extra `/asax/` in it:

`https://almascience.nrao.edu/asax/rh/data/expand/uid___A001_X133d_X3b63`

`/asax/` removed from the link:

`https://almascience.nrao.edu/rh/data/expand/uid___A001_X133d_X3b63`

alipnick:
I should note that I don't use any astroquery functions outside of `AlmaClass.stage_data()`, `AlmaClass.login()`, and `AlmaClass.download_files()` so if the parameter `dataarchive_url` is used outside these functions and requires the `/asax/`, I'm not aware.

keflavich:
I'm trying to verify that this fixes something before I merge it, but I wasn't able to reproduce #1737.    This change appeared to do no harm, though.

keflavich:
@alipnick if you have a chance, could you please confirm that the issues you noted here still occur?  I was not able to reproduce either of them, so I _suspect_ something changed upstream, and these changes are no longer needed.

However, there's still the typo correction you made in this PR - so, if we don't need the `s/asax/aq/` part, we'd still want the added comma.

PR #1914, opened 2020-12-06
Astroquery module for ISO archive (clean pull request)

jesusjuansalgado:
Added new astroquery submodule called:
astroquery.esa.iso

(clean pull request version)
This module allows access to the ESA mission ISO (Infrared Space Observatory) data:

https://www.esa.int/Science_Exploration/Space_Science/ISO_overview

ISO (1995-2001) is an important ESA legacy mission that observed the sky in the
infrared wavelength. These data have special interest these days as it is being used
by scientists e.g. to help on the preparation JWST proposals, both because of the
energy range that ISO covered and for transient data identification.


jesusjuansalgado:
I remove previous pull request and I add this one only with the basic files to be added. I imagine maintainers will need to take the action to add a milestone for it

keflavich:
Thanks @jesusjuansalgado !

PR #1980, opened 2021-02-05
MAST: making AWS clous access anonymous

ceb8:
MAST data on AWS is now free to the world, so this PR changes MAST cloud access from "RequesterPays" to anonymous access.


ceb8:
@bsipocz Looking at the failing tests here, it's not on the mast module, so I'm thinking that failure shouldn't block this PR being merged?

PR #1988, opened 2021-02-10
Add alternative server with the same interface as HEASARC, but different tables (more recent data of INTEGRAL, some CTA tables)

volodymyrss:
[ISDC](https://www.isdc.unige.ch/) is providing INTEGRAL data for the community, and HEASARC imports tables from it, with a delay. ISDC uses [copy of the HEASARC service](https://www.isdc.unige.ch/browse/w3table.pl) - older version, and hence can be added as an alternative source, with some extra handling of exception in decoding, relevant in some cases.

Relation between ISDC and HEASARC is explained here: https://heasarc.gsfc.nasa.gov/docs/integral/inthp_archive.html


volodymyrss:
In addition to ISDC interface, we suggest a modification to validate requests, forbidding unknown parameters, which are otherwise ignored.  Without this, a simple typo could result in unexpected results, without a warning.

volodymyrss:
It is possible that instead it would be better to make another module, say, "isdc",  and re-use most of the heasarc module in it. Though most of the changes to heasarc module here (such as allowed parameter checking, time range) could be  of general use.

volodymyrss:
Thanks @bsipocz for assigning the label!
I wonder if I am following the correct procedure with this PR, by creating it without prior issue or other discussions. I would not want to break conventions and appear inconsiderate.

astroquery/heasarc/core.py
```diff
@@ -188,9 +204,10 @@ def _fallback(self, content):
         colstart = [y for x, y in header.items() if "TBCOL" in x]
         collens = [int(float(y[1:]))
                    for x, y in header.items() if "TFORM" in x]
+
         new_table = []
 
-        old_table = content.split("END")[-1].strip()
+        old_table = content.decode().split("END")[-1].strip()
```
keflavich, line 210: `content.decode` should (almost) never be needed; you may want to use `response.text` instead of `response.content` when receiving the web content.

astroquery/heasarc/core.py
```diff
@@ -228,7 +245,10 @@ def _parse_result(self, response, verbose=False):
             table = Table.read(data, hdu=1)
             return table
         except ValueError:
-            return self._fallback(response.content)
+            try:
+                return self._fallback(response.content)
```
keflavich, line 249: following the note above - maybe this should be `response.text`?

docs/heasarc/heasarc.rst
```diff
@@ -189,6 +189,20 @@ that can be queried.
 The returned table includes both the names and a short description of each 
 mission table.
 
+Using alternative HEASARC servers
+-------------------------------
```
keflavich, line 193:
```suggestion
---------------------------------
```

docs/heasarc/heasarc.rst
```diff
@@ -189,6 +189,20 @@ that can be queried.
...
+
+It is possible to set alternative locations for HEASARC server. One such location
+is hosted by `INTEGRAL Science Data Center <https://www.isdc.unige.ch/>`_, and has further, 
+more recent, tables relevant for INTEGRAL spacecraft.
```
keflavich, line 197: suggestion: add an example grabbing one of those new tables to show users why they might want to use this mirror

astroquery/heasarc/core.py
```diff
@@ -43,7 +48,7 @@ def query_mission_list(self, cache=True, get_query_payload=False):
         Returns a list of all available mission tables with descriptions
         """
         request_payload = self._args_to_payload(
-            Entry='none',
+            entry='none',
```
keflavich, line 51: is this definitely right?  `Entry` is used below, but you changed to `entry` here.  It might be better to use the same capitalization everywhere for consistency; maybe it's better to change all to `Entry`, since that appears to be what the service expects?

keflavich (changes requested):
@volodymyrss Thanks for the PR, and sorry it took us a long time to see it.  I have a few minor suggestions, but this looks like a straightforward and helpful addition.

keflavich:
@volodymyrss you followed the right procedure - we welcome PRs like this one!  It does help to remind us if we haven't gotten to it; we don't mind an occasional "bump".

astroquery/heasarc/core.py
volodymyrss: The `_args_to_payload` function actually documents `entry`:

https://github.com/astropy/astroquery/blob/554d35fee2f1339b90a75862944c968c8eda61e4/astroquery/heasarc/core.py#L241-L245

and also actually uses it:

https://github.com/astropy/astroquery/blob/554d35fee2f1339b90a75862944c968c8eda61e4/astroquery/heasarc/core.py#L283-L286

Passing `Entry` was silently ignored since there was no check for unused parameters. I guess it was just a typo, and the developers anticipated to use `entry`.
It had no impact though, since in this particular call it is set to `none` which is the default when it is not set at all.

Note in the quote above that `action` function argument is similarly transformed to `Action` for the service.
This way the function arguments are always lowercase, and are transformed to different cases as expected by the service, which depends on the argument.

I could certainly change it so that Entry and Action are always capitalized.
Then I would probably also change other parameters, and descriptions.
Things like `radius`:

https://github.com/astropy/astroquery/blob/554d35fee2f1339b90a75862944c968c8eda61e4/astroquery/heasarc/core.py#L325-L327

But what worries me is that this will change user interface:

https://github.com/astropy/astroquery/blob/554d35fee2f1339b90a75862944c968c8eda61e4/astroquery/heasarc/tests/test_heasarc_remote.py#L68

Which is something I was specifically trying to avoid. Would not want to break somebody's (and our own) scripts.

So let me know what you prefer! 
Should I change all arguments, including those provided to the user, to match those expected by the service?
Only change some?

keflavich: You raise great points here.

OK, let's stick with the lowercase for user-facing parameters, and therefore a mix of cases throughout the code (unfortunately).  It may be useful to include a comment briefly summarizing why we're using mixed case - i.e., just summerize what you said above, but only in 1 line.

docs/heasarc/heasarc.rst
volodymyrss: good idea, added. And I also added example for requesting time range in general case (as well as in our tables - it is critical there).

astroquery/heasarc/core.py
volodymyrss: ok, makes sense, I tried to avoid decode here and in other places where response.text can be used instead.

astroquery/heasarc/core.py
volodymyrss: ok, done.

volodymyrss:
Thanks! Please let me know there is anything else I could improve. 

I looked into the non-remote, I could add some, but I am not entirely sure they are useful, since they will not break if the service changes. But the change at the service will render module not useful, and it should be noticed.
Maybe especially in this case, service output post-processing is virtually non-existent, and what is valuable is the interface.

volodymyrss:
Hello @keflavich , as suggested I bump!

astroquery/heasarc/core.py
```diff
@@ -1,7 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 import warnings
-from six import BytesIO
+from six import StringIO, BytesIO
+from astropy.io.fits.column import _AsciiColumnFormat
```
keflavich, line 5: we shouldn't use `_`-preceded tools - they are private and therefore are not guaranteed to be available

astroquery/heasarc/core.py
```diff
@@ -177,20 +174,39 @@ def query_region_async(self, position, mission, radius,
...
+    def _old_w3query_fallback(self, content):
+        # old w3query (such as that used in ISDC) return very strange fits, with all ints
+
+        f = fits.open(BytesIO(content))
+
+        for c in f[1].columns:
+            if c.disp is not None:
+                c.format = c.disp
+            else:
+                c.format = _AsciiColumnFormat(str(c.format).replace("I", "A"))
```
keflavich, line 186: is there another way to achieve this without using a private function?

keflavich (changes requested):
Just one change left before I approve - if possible, we should avoid using private functions from astropy.

astroquery/heasarc/core.py
volodymyrss: Looks like there is a simpler way indeed, just:

```python
 c.format = str(c.format).replace("I", "A")
```

It seems to work the same, I added a small extra test to make sure it is good.

Not sure why I did not use this approach to begin with, it must have not worked, it was some time ago.

keflavich (approved):
@bsipocz @ceb8 either one of you want to give a final once-over and merge?

ceb8:
Thanks @volodymyrss !

volodymyrss:
Thanks to you all!

PR #2145, opened 2021-09-08
Speed up `mast.Observations.get_cloud_uris()`

barentsen:
When using `astroquery.mast`, it is currently very slow to download multiple data products in cloud-enabled mode.

The main reason for the poor performance is that the `Observations.get_cloud_uris()` helper method is very slow at retrieving AWS S3 URIs.  For example, the following example requires >2 minutes to obtain URIs for 186 data products:

Screen Shot 2021-09-07 at 5 44 33 PM

This is unfortunate because slow data access makes the cloud-enabled mode less attractive.

The main reason `Observations.get_cloud_uris()` is slow is because every data products triggers a separate HTTP request to MAST's `path_lookup` API via the `astroquery.mast.utils.mast_relative_path()` helper function.

This PR proposes to modify these functions such that MAST's `path_lookup` API is called in chunks of 50 products at once, allowing `Observations.get_cloud_uris()` to execute much faster.  For example, the demo above is ~7x faster when using this PR branch:

Screen Shot 2021-09-07 at 6 02 26 PM

(Note: the remaining 20s used by `get_cloud_uris()` is spent polling every cloud URI to verify that it exists before returning it to the user, because not every MAST product is mirrored in the cloud.  This will be harder to speed up, but fortunately this may be fast when running in the cloud.)

### DIFF
--- a/astroquery/mast/cloud.py
+++ b/astroquery/mast/cloud.py
@@ -113,7 +113,7 @@ class CloudAccess:  # pragma:no-cover
 
         # Making sure we got at least 1 URI from the query above.
         if uri_list[0] == None:
-            raise InvalidQueryError("No products found on the cloud for this query.")
+            warnings.warn("Unable to locate file {}.".format(data_product), NoResultsWarning)
         else:
             # Output from ``get_cloud_uri_list`` is always a list even when it's only 1 URI
             return uri_list[0]
--- a/astroquery/mast/cloud.py
+++ b/astroquery/mast/cloud.py
@@ -14,7 +14,7 @@ from astroquery import log
 from astropy.utils.console import ProgressBarOrSpinner
 from astropy.utils.exceptions import AstropyDeprecationWarning
 
-from ..exceptions import NoResultsWarning
+from ..exceptions import NoResultsWarning, InvalidQueryError
 
 from . import utils
 
@@ -110,7 +110,12 @@ class CloudAccess:  # pragma:no-cover
         """
 
         uri_list = self.get_cloud_uri_list(data_product, include_bucket=include_bucket, full_url=full_url)
-        return uri_list[0]
+
+        # Making sure we got at least 1 URI from the query above.
+        if len(uri_list) == 0:
+            raise InvalidQueryError("No products found on the cloud for this query.")
+        else:
+            return uri_list[0]
 
     def get_cloud_uri_list(self, data_products, include_bucket=True, full_url=False):
         """
@@ -142,14 +147,9 @@ class CloudAccess:  # pragma:no-cover
 
         uri_list = []
         for path in paths:
-
             if path is None:
                 uri_list.append(None)
-            elif 'galex' in path:
-                path = path.lstrip("/mast/")
             else:
-                path = path.lstrip("/")
-
                 try:
                     # Use `head_object` to verify that the product is available on S3 (not all products are)
                     s3_client.head_object(Bucket=self.pubdata_bucket, Key=path)
--- a/astroquery/mast/cloud.py
+++ b/astroquery/mast/cloud.py
@@ -14,7 +14,7 @@ from astroquery import log
 from astropy.utils.console import ProgressBarOrSpinner
 from astropy.utils.exceptions import AstropyDeprecationWarning
 
-from ..exceptions import NoResultsWarning, InvalidQueryError
+from ..exceptions import NoResultsWarning
 
 from . import utils
 
@@ -109,30 +109,8 @@ class CloudAccess:  # pragma:no-cover
             found in the cloud, None is returned.
         """
 
-        s3_client = self.boto3.client('s3', config=self.config)
-
-        path = utils.mast_relative_path(data_product["dataURI"])
-        if path is None:
-            raise InvalidQueryError("Malformed data uri {}".format(data_product['dataURI']))
-
-        if 'galex' in path:
-            path = path.lstrip("/mast/")
-        else:
-            path = path.lstrip("/")
-
-        try:
-            s3_client.head_object(Bucket=self.pubdata_bucket, Key=path)
-            if include_bucket:
-                path = "s3://{}/{}".format(self.pubdata_bucket, path)
-            elif full_url:
-                path = "http://s3.amazonaws.com/{}/{}".format(self.pubdata_bucket, path)
-            return path
-        except self.botocore.exceptions.ClientError as e:
-            if e.response['Error']['Code'] != "404":
-                raise
-
-        warnings.warn("Unable to locate file {}.".format(data_product['productFilename']), NoResultsWarning)
-        return None
+        uri_list = self.get_cloud_uri_list(data_product, include_bucket=include_bucket, full_url=full_url)
+        return uri_list[0]
 
     def get_cloud_uri_list(self, data_products, include_bucket=True, full_url=False):
         """
@@ -156,8 +134,37 @@ class CloudAccess:  # pragma:no-cover
             List of URIs generated from the data products, list way contain entries that are None
             if data_products includes products not found in the cloud.
         """
+        s3_client = self.boto3.client('s3', config=self.config)
 
-        return [self.get_cloud_uri(product, include_bucket, full_url) for product in data_products]
+        paths = utils.mast_relative_path(data_products["dataURI"])
+        if isinstance(paths, str):  # Handle the case where only one product was requested
+            paths = [paths]
+
+        uri_list = []
+        for path in paths:
+
+            if path is None:
+                uri_list.append(None)
+            elif 'galex' in path:
+                path = path.lstrip("/mast/")
+            else:
+                path = path.lstrip("/")
+
+                try:
+                    # Use `head_object` to verify that the product is available on S3 (not all products are)
+                    s3_client.head_object(Bucket=self.pubdata_bucket, Key=path)
+                    if include_bucket:
+                        path = "s3://{}/{}".format(self.pubdata_bucket, path)
+                    elif full_url:
+                        path = "http://s3.amazonaws.com/{}/{}".format(self._pubdata_bucket, path)
+                    uri_list.append(path)
+                except self.botocore.exceptions.ClientError as e:
+                    if e.response['Error']['Code'] != "404":
+                        raise
+                    warnings.warn("Unable to locate file {}.".format(path), NoResultsWarning)
+                    uri_list.append(None)
+
+        return uri_list
 
     def download_file(self, data_product, local_path, cache=True):
         """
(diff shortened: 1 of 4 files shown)


bsipocz:
This now needs and rebase. Also, @jaymedina could you assign it to the person who are supposed to review this, or cc them on the PR? Thanks!

jaymedina:
Hi @bsipocz I'll review this in conjunction with @tomdonaldson (I think he started a review already?)

astroquery/mast/cloud.py
```diff
@@ -154,8 +134,33 @@ def get_cloud_uri_list(self, data_products, include_bucket=True, full_url=False)
             List of URIs generated from the data products, list way contain entries that are None
             if data_products includes products not found in the cloud.
         """
+        s3_client = self.boto3.client('s3', config=self.config)
 
-        return [self.get_cloud_uri(product, include_bucket, full_url) for product in data_products]
+        paths = utils.mast_relative_path(data_products["dataURI"])
```
jaymedina, line 139: This will break for GALEX data retrieval from the cloud, unfortunately. See here for how I tweaked this output in a recent PR following the new GALEX availability on the cloud: https://github.com/astropy/astroquery/blob/2b3d95498d0b6e73b9c3434ccc7d06cc50642f11/astroquery/mast/cloud.py#L119

Your two options would be to either: 
1. Make a tweak similar to this in `get_cloud_uri_list`
2. Make the tweak in `mast_relative_path` - this is probably the cleanest option. And then we can remove my tweak from `cloud.py`.

astroquery/mast/cloud.py
```diff
@@ -154,8 +134,33 @@ def get_cloud_uri_list(self, data_products, include_bucket=True, full_url=False)
...
+        if isinstance(paths, str):  # Handle the case where only one product was requested
+            paths = [paths]
+
+        uri_list = []
+        for path in paths:
+            if path is None:
+                uri_list.append(None)
```
jaymedina, line 146: Should we raise an error here like how it was done in the original `get_cloud_uri`?:
`raise InvalidQueryError("Malformed data uri {}".format(data_product['dataURI']))`

jaymedina (changes requested):
Looks solid, the main suggestion I have is what I mentioned about the GALEX relative path tweak that I discovered was necessary for GALEX on S3. Also added some minor comments, but in general this looks good. 

I'd say the first thing to do right now is rebase and rerun unit tests, to verify that GALEX indeed breaks without a tweak to the `mast_relative_path` output.

astroquery/mast/cloud.py
bsipocz: re this suggestion: in the cloud WG we're working on generalizing these methods and move them out from mast. So a tweak at the place of usage, or a generically usable kwarg would be preferable rather than a hard-wired conditional on mast specific strings.

bsipocz:
@barentsen @jaymedina - what are the plans with this PR?

bsipocz:
Ping @barentsen @jaymedina @tomdonaldson, do you have any plans for finishing up this PR? I expect it's reasonably close for being ready.

bsipocz:
ping @jaymedina!

jaymedina:
I'll take a look at this and make updates accordingly.

astroquery/mast/cloud.py
jaymedina: What I did was fold in the `path.strip()` logic out of `cloud.py` and into `util.py`'s `mast_relative_path` function. In other words, I essentially went with option `2.` listed above. And because this is a MAST-specific function in `utils`, it makes sense for MAST-specific tweaks to take place in there. Let me know if that works for you!

astroquery/mast/cloud.py
jaymedina: This comment is no longer necessary.

jaymedina:
I've rebased, adapted the code to account for GALEX cloud products, updated error handling and unit tests, moved the MAST-specific string manipulation out of `cloud.py`, and squashed the commits into 4. I think this is ready for another review @bsipocz , thanks in advance.

astroquery/mast/cloud.py
```diff
@@ -109,30 +109,13 @@ def get_cloud_uri(self, data_product, include_bucket=True, full_url=False):
...
-        path = utils.mast_relative_path(data_product["dataURI"])
-        if path is None:
-            raise InvalidQueryError("Malformed data uri {}".format(data_product['dataURI']))
+        uri_list = self.get_cloud_uri_list(data_product, include_bucket=include_bucket, full_url=full_url)
 
-        if 'galex' in path:
-            path = path.lstrip("/mast/")
+        # Making sure we got at least 1 URI from the query above.
+        if len(uri_list) == 0:
+            raise InvalidQueryError("No products found on the cloud for this query.")
```
bsipocz, line 116: Is this an InvalidQueryError or rather a `NoResultsWarning`? We haven't previously raised the error, but if you think the query has an error in it, rather than just there is no result, then we can raise it from now on (probably would require its own changelog entry)

astroquery/mast/cloud.py
```diff
@@ -109,30 +109,13 @@ def get_cloud_uri(self, data_product, include_bucket=True, full_url=False):
...
-            elif full_url:
-                path = "http://s3.amazonaws.com/{}/{}".format(self.pubdata_bucket, path)
-            return path
-        except self.botocore.exceptions.ClientError as e:
-            if e.response['Error']['Code'] != "404":
-                raise
-
-        warnings.warn("Unable to locate file {}.".format(data_product['productFilename']), NoResultsWarning)
-        return None
+            return uri_list[0]
```
bsipocz, line 118: Why only the first element, what is expected to be included in the rest?

astroquery/mast/cloud.py
```diff
@@ -156,8 +139,32 @@ def get_cloud_uri_list(self, data_products, include_bucket=True, full_url=False)
...
+        uri_list = []
+        for path in paths:
+            if path is None:
+                uri_list.append(None)
+            else:
+                try:
+                    # Use `head_object` to verify that the product is available on S3 (not all products are)
+                    s3_client.head_object(Bucket=self.pubdata_bucket, Key=path)
+                    if include_bucket:
+                        path = "s3://{}/{}".format(self.pubdata_bucket, path)
```
bsipocz, line 157: I don't necessarily like the overwriting of variables here as when this has to be debugged it's a bit more difficult to distinguish between the `path` from the iteration vs the path here. Maybe use `s3_path`?

astroquery/mast/utils.py
```diff
@@ -155,22 +155,48 @@ def parse_input_location(coordinates=None, objectname=None):
...
+    # to avoid "414 Client Error: Request-URI Too Large".
+    uri_list_chunks = list(_split_list_into_chunks(uri_list, chunk_size=50))
+
+    result = []
+    for chunk in uri_list_chunks:
+        response = _simple_request("https://mast.stsci.edu/api/v0.1/path_lookup/",
+                                   {"uri": chunk})
+        json_response = response.json()
+        for uri in chunk:
+            path = json_response.get(uri[1])["path"]
```
bsipocz, line 185: I don't see what's expected to be in `uri` here, could you explain it a bit, so I see why the second element?

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -405,28 +405,32 @@ def test_get_cloud_uri(self):
 
     def test_get_cloud_uris(self):
         pytest.importorskip("boto3")
-        test_obs_id = '25568122'
+        test_obs_ids = ["25568122", "31411"]
```
bsipocz, line 408: do this with `pytest.mark.parametrize` instead of an internal loop

bsipocz (commented):
Thanks for picking this up and pushing it through the finish line.

I have some minor comments, mostly asking for clarifications.

astroquery/mast/cloud.py
jaymedina: The output for `get_cloud_uri_list` is always a list, even when the output returns only 1 URI, so the sole element gets indexed out before being returned. I just committed some commentary above this line for clarity.

astroquery/mast/cloud.py
jaymedina: I chose to raise an error rather than a warning so that the output error message is more helpful. Here is what it looks like when I use a `NoResultsWarning`: 

image

The final `TypeError` is an indirect result of the empty uri list, so it's a vague error message. 

Raising a custom `InvalidQueryError` makes things more clear:

image

bsipocz: Maybe then a conditional is needed to return the empty result early, before going into the Typeerror. The point is basically to use the InvalidqueryError when there is an issue with the user input. Not having data, imo is not in this category.

astroquery/mast/cloud.py
jaymedina: There wouldn't be a case where a request for a single cloud URI would return more than one because of L171 in `util.py` of this branch, but I can raise an AssertionError here to ensure that there is only 1 element in the list, what do you think?

bsipocz: I'm really going into bikeshedding now, but if that's the case, why does it have to be a list to begin with?

Also, L173 is never reached? 

Maybe just a comment about this would be enough, no need for the assertion.

astroquery/mast/cloud.py
jaymedina: On second thought, a user will realize that their call to `Observations.get_cloud_uri(product)` returns an empty list before calling `Observations.download_file`, since they have to feed the uri in string form and by default the uri gets returned as a list even when it's 1 URI. So this error handling thing might be redundant.

bsipocz:
Close/reopen to retrigger CI (as logs were long gone now)

bsipocz:
Needs a rebase for RTD fix and some PEP8 fixes before final review.

jaymedina:
Hi @bsipocz, I'm tagging the interim `astroquery.mast` maintainer, Julie Imig (@astrojimig). She will be taking over remaining PRs and will be the liaison for MAST-related issues.

bsipocz:
@astrojimig -  I suppose the conflict is coming from the changes from your recently merged panstarss branch. Anyway, it would be nice to finally wrap this PR up and merge it. 

Could you try and rebase this one? I'm not sure you inherited the access to Geert's branch from Jenny.

astrojimig:
@bsipocz Sure, I'll try to work on this next week!

snbianco:
@bsipocz Working on this now! This might just be my own misunderstanding of GitHub, but do I need some sort of permissions to push to this branch? Whenever I try, I get a permission denied error. Is there an alternate method I should be using?

bsipocz:
> @bsipocz Working on this now! This might just be my own misunderstanding of GitHub, but do I need some sort of permissions to push to this branch? Whenever I try, I get a permission denied error. Is there an alternate method I should be using?

You cannot push to someone else's fork by default, so you will need to take the applicable changes from here. I have reviewed this a couple of times, but things also have changed in the module, so I'm not certain which way is more efficient and useful (trying to keep these as is and rebase and fix from there, or look into the changes and apply the ones that are still relevant to your other PR (and then just rename your method and add fixes to it so old usage of get_cloud_uris() will work with it, too)

PR #2210, opened 2021-11-16
Desi legacy survey

burnout87:


burnout87:
As advised in #2192 , the various renaming have been applied

keflavich:
Thanks.  We still need a few things before this is merged:
1. Documentation.  At least basic documentation describing the purpose and use of this tool should be added as `docs/desi.rst`
2. Non-remote tests.  At least one non-remote test is needed - we can help guide you in how to do this, but following existing examples from other modules is probably the best way to go. 
3. Cleanup.  There is a bunch of commented-out text from the tutorial.  This text is harmless, but we'd prefer to have it removed before merging.

bsipocz:
@burnout87 - thanks for the PR. Before going into a more formal/thorough review, could you please rebase to get rid of the merge commits, as well as to squash the commits that add the test files? Those files need to be cut back a lot in size, currently, the new module's test data takes up 75+% of all of astroquery's size. Thanks!

astroquery/desi/core.py
```diff
@@ -0,0 +1,140 @@
...
+
+__all__ = ['DESILegacySurvey', 'DESILegacySurveyClass']
+
+
+@async_to_sync
+class DESILegacySurveyClass(BaseQuery):
+    URL = conf.server
+    TIMEOUT = conf.timeout
+
+    def query_region_async(self, coordinates, radius, data_release=9):
```
ceb8, line 30: I am unclear as to why this function is written as an async function, it seems like the same results get returned whichever version the user calls.

astroquery/desi/core.py
```diff
@@ -0,0 +1,140 @@
...
+        coordinates : str or `astropy.coordinates`.
+            coordinates around which to query
+        radius : str or `astropy.units.Quantity`.
+            the radius of the cone search
+
+        Returns
+        -------
+        response : `requests.Response`
+            The HTTP response returned from the service.
+            All async methods should return the raw HTTP response.
```
ceb8, line 45: This function does not seem to return an HTTP response as it should if it is going to be an async function.

astroquery/desi/core.py
```diff
@@ -0,0 +1,140 @@
...
+        mask = tap_result['type'] != 'D'
+        filtered_table = tap_result[mask]
+
+        return filtered_table
+
+    def get_images_async(self, position, survey, coordinates=None, data_release=9,
+                         projection=None, pixels=None, scaling=None,
+                         sampler=None, resolver=None, deedger=None, lut=None,
+                         grid=None, gridlabels=None, radius=None, height=None,
+                         width=None, cache=True, show_progress=True, image_band='g'):
```
ceb8, line 66: As in the query_region_async function I don't see any difference between the async and sync versions of this function, async functions should always return HTTP response objects.

astroquery/desi/core.py
```diff
@@ -0,0 +1,140 @@
...
+        A list of context-managers that yield readable file-like objects
+        """
+
+        image_size_arcsec = radius.arcsec
+        pixsize = 2 * image_size_arcsec / pixels
+
+        image_url = 'https://www.legacysurvey.org/viewer/fits-cutout?ra=' + str(position.ra.deg) + '&dec=' + str(position.dec.deg) + '&size=' + str(
+            pixels) + '&layer=ls-dr' + str(data_release) + '&pixscale=' + str(pixsize) + '&bands=' + image_band
+
+        print("image_url: ", image_url)
```
ceb8, line 79: Suggest offering a way to suppress this print statement by having a verbose argument or similar.

astroquery/desi/core.py
```diff
@@ -0,0 +1,140 @@
...
+        return [fits_file]
+
+    def _parse_result(self, responses, verbose=False):
+        tables_list = []
+        output_table = Table()
+        tmp_dir_list = []
+
+        # handles the cases of query_region with TAP which returns a Table
+        # or get_images which should return a list of HDUList, and here in DESILegacySurvey
+        # a list of one HDUList object is returned
```
ceb8, line 98: What are the cases where this is not true?

astroquery/desi/tests/test_module_remote.py
```diff
@@ -0,0 +1,65 @@
...
+
+        ra = Angle('11h04m27s', unit='hourangle').degree
+        dec = Angle('+38d12m32s', unit='hourangle').degree
+        coordinates = SkyCoord(ra, dec, unit='degree')
+
+        radius = Angle(5, unit='arcmin')
+
+        query1 = astroquery.desi.DESILegacySurvey.query_region(coordinates, radius=radius, data_release=9)
+        print("Test completion: ", time.time() - t0)
+        print(query1['ra', 'dec', 'type'])
```
ceb8, line 29: Remove print statements in the testing suite (only mentioning it here, but remove them throughout).

ceb8 (changes requested):
This is looking very good! Most of my comments are small, the two bigger items are:
1. The examples in the docs don't run as-is.
2. The async functions don't seem to actually be async, in that they don't return HTTP responses. They either need to follow the async API specs or not be called async.

astroquery/desi/core.py
```diff
@@ -0,0 +1,395 @@
...
+from numpy import pi, cos
+# export all the public classes and methods
+__all__ = ['DESILegacySurvey', 'DESILegacySurveyClass']
+
+# declare global variables and constants if any
+
+
+# Now begin your main class
+# should be decorated with the async_to_sync imported previously
+from astropy.io.fits import HDUList
```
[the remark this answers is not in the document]
burnout87: This was already moved up, I see the code used for your review is outdated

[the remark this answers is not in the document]
ceb8: Ah woops, I think I started writing comments ages ago and then forgot to check if they were still relevant. Apologies.

bsipocz:
My comment above still stands, please rebase/squash out the commits so bigger test files are not added and then removed in the history

burnout87:
thanks for your time. The first iteration of comments has been addressed.

astroquery/desi/core.py
burnout87: Thanks for the comment, that is a valid point I was not considering. Answer is quite simple: no it does not at the current state. So, it can be removed.

burnout87: It has now been removed.

burnout87:
> My comment above still stands, please rebase/squash out the commits so bigger test files are not added and then removed in the history

Comments squashing executed.

astroquery/desi/core.py
```diff
@@ -0,0 +1,77 @@
...
+
+        tap_result = tap_service.run_sync(qstr)
+        tap_result = tap_result.to_table()
+        # filter out duplicated lines from the table
+        mask = tap_result['type'] != 'D'
+        filtered_table = tap_result[mask]
+
+        return filtered_table
+
+    def get_images(self, position, data_release=9, pixels=None, radius=None, show_progress=True, image_band='g'):
```
bsipocz, line 53: Make optionals kwarg only. Please do this for the rest of the methods, too, I won't comment on it elsewhere.

```suggestion
    def get_images(self, position, *, data_release=9, pixels=None, radius=None, show_progress=True, image_band='g'):
```

astroquery/desi/core.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Returns
+        -------
+        A list of `astropy.io.fits.HDUList` objects.
+        """
+
+        image_size_arcsec = radius.arcsec
+        pixsize = 2 * image_size_arcsec / pixels
```
bsipocz, line 61: This would raise an exception going with the default values

astroquery/desi/core.py
```diff
@@ -0,0 +1,77 @@
...
+        return filtered_table
+
+    def get_images(self, position, data_release=9, pixels=None, radius=None, show_progress=True, image_band='g'):
+        """
+        Returns
+        -------
+        A list of `astropy.io.fits.HDUList` objects.
+        """
+
+        image_size_arcsec = radius.arcsec
```
bsipocz, line 60: This would raise an exception with the default value

astroquery/desi/core.py
```diff
@@ -0,0 +1,77 @@
...
+        pixsize = 2 * image_size_arcsec / pixels
+
+        image_url = 'https://www.legacysurvey.org/viewer/fits-cutout?ra=' + str(position.ra.deg) + '&dec=' + str(position.dec.deg) + '&size=' + str(
```
bsipocz, line 63: Use fstring and url from config

astroquery/desi/tests/setup_package.py
```diff
@@ -0,0 +1,10 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+import os
+
+
+def get_package_data():
+    paths = [os.path.join('data', '*.txt'),
+             os.path.join('data', '*.fits'),
+             os.path.join('data', '*.fits.gz'),
```
bsipocz, line 8: you don't have `fits.gz` files, no need for this line

astroquery/desi/tests/test_module.py
```diff
@@ -0,0 +1,132 @@
+import astropy.io.votable
```
bsipocz, line 1: rename this file from `test_module` to `test_desi`, as well as the remote tests file.

astroquery/desi/tests/test_module.py
```diff
@@ -0,0 +1,132 @@
...
+import pytest
+import os
+import numpy as np
+import pyvo as vo
+
+from astropy.table import Table
+from astropy.io import fits
+from pyvo.dal import TAPResults
```
bsipocz, line 9: why not use this similarly to TAPService as `vo.dal.TAPResults`? I'm OK either way, just wondering if there is a reason.

astroquery/desi/tests/test_module.py
```diff
@@ -0,0 +1,132 @@
...
+pixels = 60
+data_release = 9
+emispheres_list = ['north', 'south']
+
+
+@pytest.fixture
+def patch_get(request):
+    try:
+        mp = request.getfixturevalue("monkeypatch")
+    except AttributeError:  # pytest < 3
```
bsipocz, line 34: no need for this, we only support python 3.7+

astroquery/desi/tests/test_module.py
```diff
@@ -0,0 +1,132 @@
...
+import pyvo as vo
+
+from astropy.table import Table
+from astropy.io import fits
+from pyvo.dal import TAPResults
+from urllib import parse
+from contextlib import contextmanager
+
+from ... import desi
+from ...utils.mocks import MockResponse
```
bsipocz, line 14: do use absolute imports, we're slowly cleaning up the codebase, but new stuff should have it.

astroquery/utils/commons.py
```diff
@@ -42,10 +42,13 @@ def FK4CoordGenerator(*args, **kwargs):
     return coord.SkyCoord(*args, frame='fk4', **kwargs)
 
 
+def ArcminRadiusGenerator(*args, **kwargs):
```
bsipocz, line 45: Why do we need this? Use coord.Angle directly in the code.

docs/desi/desi.rst
```diff
@@ -0,0 +1,111 @@
...
+that can be formulated.
+
+Query a region
+===============
+
+This example shows how to query a certain region with DesiLegacySurvey.
+We'll use a set of coordinates that define the region of interest,
+and search within a 5 arcmin radius.
+
+.. code-block:: python
```
bsipocz, line 21: Use the `.. doctest-remote-data::` directive instead of `code-block`, here and everywhere below, otherwise all these make the tests fail.

docs/desi/desi.rst
```diff
@@ -0,0 +1,111 @@
...
+    >>> radius_input = 3
+    >>> pixels = 1296000
+
+    >>> pos = SkyCoord(ra, dec, unit='degree')
+    >>> radius = Angle(radius_input, unit='arcmin')
+    >>> pixels = 60
+    >>> im = DESILegacySurvey.get_images(pos, data_release=9, radius=radius, pixels=pixels)
+
+.. code-block:: python
+
```
bsipocz, line 110: you need to write the whole example script as one big block, along with the output, have a look at the other modules (e.g. cadc or ipac.irsa) for examples. If you run the remote tests on this file, edit it until it passes. (`python setup.py test -t docs/desi/desi.rst -R`)

bsipocz (changes requested):
It's getting there, thanks.

Here is my first round of review, most of the items are minor, and are needed to fix the failures we see in CI, etc.

astroquery/desi/core.py
```diff
@@ -0,0 +1,77 @@
...
+        Parameters
+        ----------
+        coordinates : str or `astropy.coordinates`.
+            coordinates around which to query
+        radius : str or `astropy.units.Quantity`.
+            the radius of the cone search
+
+        Returns
+        -------
+        response : `astropy.table.Table`
```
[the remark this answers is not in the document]
burnout87: Can you give me more details on this comment? The `query_region` method is no longer async

astroquery/template_module/tests/test_module_remote.py
```diff
@@ -1,7 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 
-# performs similar tests as test_module.py, but performs
+# performs similar tests as test_desi.py, but performs
```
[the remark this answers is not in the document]
bsipocz: (practical rule of thumb: when you `git add`, get into the habit of adding files/changes specifically, refrain from adding "all". It's not much slower in practice, but would avoid committing unrelated changes.)

astroquery/desi/core.py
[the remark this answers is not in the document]
bsipocz: you're right, I likely was stuck in the view of an old version as clearly remember to see `_async` in the name of the method. 

In this case, though I don't think the `async_to_sync` decorator has any effect on the class.

astroquery/template_module/tests/test_module_remote.py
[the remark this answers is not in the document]
burnout87: ok, I realize only now the issue, and indeed I ended up adding those changes without realizing it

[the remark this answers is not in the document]
bsipocz: no worries, it's actually nice that you recovered that other modules didn't update the use of the template module either :)

bsipocz:
@burnout87 - I'll come back to this later during the week for another round of review and minor fixes that would fix e.g. the docs build.

astroquery/desi/core.py
```diff
@@ -0,0 +1,97 @@
...
+
+        return filtered_table
+
+    def get_images(self, position, pixels, radius, *, data_release=9, show_progress=True, image_band='g'):
+        """
+        Downloads the images for a certain region of interest.
+
+        Parameters
+        -------
+        position: `astropy.coordinates`.
```
bsipocz, line 61: Do add all the arguments to the docstring. Also, pixel and radius can be optional.

astroquery/desi/core.py
```diff
@@ -0,0 +1,97 @@
...
+                    f"size={pixels}&"
+                    f"layer=ls-dr{data_release}&"
+                    f"pixscale={pixsize}&"
+                    f"bands={image_band}")
+
+        file_container = commons.FileContainer(image_url, encoding='binary', show_progress=show_progress)
+
+        try:
+            fits_file = file_container.get_fits()
+        except (requests.exceptions.HTTPError, urllib.error.HTTPError) as e:
```
bsipocz, line 90: nitpick: don't use single letter variables, it's difficult to do grep etc for them. Something a little longer e.g. 'exp' or 'err' would do

astroquery/desi/core.py
```diff
@@ -0,0 +1,97 @@
...
+            fits_file = None
+            warnings.warn(f"{str(e)} - Problem retrieving the file at the url: {image_url}", NoResultsWarning)
```
bsipocz, line 92: are you sure you don't want to raise this as an exception? Why and when does this return an HTTPError?

astroquery/desi/tests/test_desi_remote.py
```diff
@@ -0,0 +1,49 @@
...
+from astropy.table import Table
+from astroquery.exceptions import NoResultsWarning
+
+
+@pytest.mark.remote_data
+class TestLegacySurveyClass:
+
+    def test_query_region(self):
+
+        ra = Angle('11h04m27s', unit='hourangle').degree
```
bsipocz, line 16: this is just a float, why the extra Angle, etc wrapping here and below? we don't need to test that `Angle` and `SkyCoord` work as expected here, that's an upstream task for astropy.

astroquery/desi/tests/test_desi_remote.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+import astroquery.desi
```
bsipocz, line 2: extreme nitpick: Do import the class rather than using the full namespace 
```suggestion
from  astroquery.desi import DESILegacySurvey
```

astroquery/desi/tests/test_desi_remote.py
```diff
@@ -0,0 +1,49 @@
...
+    def test_get_images(self, valid_inputs):
+
+        if valid_inputs:
+            ra = Angle('11h04m27s', unit='hourangle').degree
+            dec = Angle('+38d12m32s', unit='hourangle').degree
+            radius_input = 0.5  # arcmin
+            pixels = 60
+        else:
+            ra = Angle('86.633212', unit='degree').degree
+            dec = Angle('22.01446', unit='degree').degree
```
bsipocz, line 36: this is just a float, why the extra Angle, etc wrapping?

bsipocz (changes requested):
Some major and minor nitpicking comments. I'm pushing the fix for the docs build separately as it's an automodapi peculiarity.

Also, as you've squashed down all the commits to one you lost the authorship credit. I'm happy to get some of that fixed by sorting out the duplicated commits/extra data etc squashes from the backup branch once we're otherwise ready to merge.

astroquery/desi/core.py
```diff
@@ -0,0 +1,97 @@
...
+        response : `astropy.table.Table`
+        """
+
+        tap_service = vo.dal.TAPService(conf.tap_service_url)
+        coordinates_transformed = coordinates.transform_to(coord.ICRS)
+
+        qstr = (f"SELECT all * FROM ls_dr{data_release}.tractor WHERE "
+               f"dec>{coordinates_transformed.dec.deg - radius.deg} and "
+               f"dec<{coordinates_transformed.dec.deg + radius.deg} and "
+               f"ra>{coordinates_transformed.ra.deg - radius.deg / np.cos(coordinates_transformed.dec.deg * np.pi / 180.)} and "
```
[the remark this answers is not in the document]
burnout87: Thanks for pointing this out, I also noticed in the parameters doc. I will specify that radius is an `Angle` object

astroquery/desi/core.py
burnout87: An HTTPError happens in the case of not valid inputs, more specifically when these are outside of the survey coverage.

We, for our internal applications, want to display a warning message in case this situation occurs.

bsipocz: yes, my point is that if the URL throws an error then I think we should throw it, too. E.g. the user expects a file as a return, but here they get a valid return yet no files.
Is there any use case where this is more useful for the users than to receive an error?

astroquery/desi/core.py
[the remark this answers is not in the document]
bsipocz: see my code example above, it doesn't have to be an angle, but can be a quantity and the lines suggested will work.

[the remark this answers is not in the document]
burnout87: we have consistently used Angle also within other contexts, so I prefer to stick with that, unless there is a reason which makes `Quantity` a better option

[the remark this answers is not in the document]
eerovaher: `Angle` is a subclass of `Quantity`, so even if a parameter type is declared to be `Quantity` the users can still provide an `Angle`.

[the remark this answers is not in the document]
bsipocz: yes, many users don't use the subclass, and most of astroquery has radius as a quantity, so I would prefer to keep that as none of the conveniences offered by angle should be a dealbreaker in this module.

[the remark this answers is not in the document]
burnout87: > (coordinates_transformed.ra - radius).to(u.deg)

I suppose this has to be followed by `.value`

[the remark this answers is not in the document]
bsipocz: >I suppose this has to be followed by .value. Am I correct?

yes, indeed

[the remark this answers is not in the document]
eerovaher: `.to_value(unit)` is faster than `.to(unit).value`.

burnout87:
> 
> Also, as you've squashed down all the commits to one you lost the authorship credit. I'm happy to get some of that fixed by sorting out the duplicated commits/extra data etc squashes from the backup branch once we're otherwise ready to merge.

Thanks! I was not considering that when doing the squash.

astroquery/desi/core.py
```diff
@@ -17,32 +19,37 @@
...
 
         Returns
         -------
-        response : `astropy.table.Table`
+        response : `~astropy.table.Table`
         """
 
+        if radius is None:
+            radius = Quantity(0.5, u.arcmin)
+
```
eerovaher, line 44:
```suggestion
```

astroquery/desi/core.py
```diff
@@ -17,32 +19,37 @@
...
 
         qstr = (f"SELECT all * FROM ls_dr{data_release}.tractor WHERE "
-               f"dec>{coordinates_transformed.dec.deg - radius.deg} and "
-               f"dec<{coordinates_transformed.dec.deg + radius.deg} and "
-               f"ra>{coordinates_transformed.ra.deg - radius.deg / np.cos(coordinates_transformed.dec.deg * np.pi / 180.)} and "
-               f"ra<{coordinates_transformed.ra.deg + radius.deg / np.cos(coordinates_transformed.dec.deg * np.pi / 180)}")
+               f"dec>{(coordinates_transformed.dec - radius).to_value(u.deg)} and "
+               f"dec<{(coordinates_transformed.dec + radius).to_value(u.deg)} and "
+               f"ra>{coordinates_transformed.ra.to_value(u.deg) - radius.to_value(u.deg) / np.cos(coordinates_transformed.dec.to_value(u.deg) * np.pi / 180.)} and "
+               f"ra<{coordinates_transformed.ra.to_value(u.deg) + radius.to_value(u.deg) / np.cos(coordinates_transformed.dec.to_value(u.deg) * np.pi / 180)}")
```
eerovaher, line 52:
```suggestion
               f"ra>{coordinates_transformed.ra.to_value(u.deg) - radius.to_value(u.deg) / np.cos(coordinates_transformed.dec)} and "
               f"ra<{coordinates_transformed.ra.to_value(u.deg) + radius.to_value(u.deg) / np.cos(coordinates_transformed.dec)}")
```

astroquery/desi/core.py
```diff
@@ -52,24 +59,38 @@ def query_region(self, coordinates, radius, *, data_release=9):
 
         return filtered_table
 
-    def get_images(self, position, pixels, radius, *, data_release=9, show_progress=True, image_band='g'):
+    def get_images(self, position, pixels=None, radius=None, *, data_release=9, show_progress=True, image_band='g'):
```
eerovaher, line 62:
```suggestion
    def get_images(self, position, pixels=None, radius=0.5*u.arcmin, *, data_release=9, show_progress=True, image_band='g'):
```

astroquery/desi/core.py
```diff
@@ -52,24 +59,38 @@ def query_region(self, coordinates, radius, *, data_release=9):
...
 
         Returns
         -------
-        list: A list of `astropy.io.fits.HDUList` objects.
+        list: A list of `~astropy.io.fits.HDUList` objects.
         """
 
+        if radius is None:
+            radius = Quantity(0.5, u.arcmin)
+
```
eerovaher, line 93:
```suggestion
```

astroquery/desi/tests/test_desi.py
```diff
@@ -0,0 +1,126 @@
...
+        file_obj = data_path(DATA_FILES['dummy_hdu_list_fits'])  # TODO: add images option
+        encoding = kwargs.get('encoding', None)
+        f = None
+        try:
+            if encoding == 'binary':
+                f = open(file_obj, 'rb')
+                yield f
+            else:
+                f = open(file_obj, 'rb')
+                yield f
```
eerovaher, line 51: The if-else block currently serves no purpose.

astroquery/desi/tests/test_desi.py
```diff
@@ -0,0 +1,126 @@
...
+        finally:
+            if f is not None:
+                f.close()
```
eerovaher, line 54: The try-finally block would not be needed if the following pattern was used:
```python
with open(...) as f:
    ...
```

astroquery/desi/tests/test_desi.py
```diff
@@ -0,0 +1,118 @@
...
+    parsed_url = parse.urlparse(url)
+    splitted_parsed_url = parsed_url.path.split('/')
+    url_filename = splitted_parsed_url[-1]
+    filename = None
+    content = None
+    if url_filename.startswith('tractor-'):
+        filename = data_path(DATA_FILES['dummy_tractor_fits'])
+
+    if filename is not None:
+        content = open(filename, 'rb').read()
```
eerovaher, line 76: It's better to use `open()` together with the `with` keyword, see https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files
```suggestion
        with open(filename, 'rb') as f:
            content = f.read()
```

astroquery/desi/core.py
burnout87: The server returns an `HTTPError` in the case of  internal error of the legacysurvey but also in the case I mentioned (ie case of not valid inputs, outside of the survey coverage), so at the moment is impossible for us to distinguish these two situations.

For this reason, we decided just to show the message in the warning to the user in both of those cases.

burnout87: Hi @bsipocz , what do you think? Does it make sense to you?

bsipocz: It's OK to customize the message but my point still stands, it should the raised. I don't see a use case where a warning rather than an error is preferred after all the query errored out and the `None` return is not really a valid results for these cases (`None` return would be fine if there is a valid query and there is no match for ti in the DB, etc.)

ceb8 (approved):

docs/desi/desi.rst
keflavich: I believe this issue is still holding tests back from passing

docs/desi/desi.rst
```diff
@@ -0,0 +1,99 @@
...
+    >>> pixels = 60
+    >>>
+    >>> pos = SkyCoord(ra, dec, unit='degree')
+    >>> radius = Angle(radius_input, unit='arcmin')
+    >>> im = DESILegacySurvey.get_images(pos, pixels, radius, data_release=9)
+
+All the information we need can be found within the object "im".
+
+.. doctest-remote-data::
+
```
keflavich, line 65: This small fix may address @bsipocz 's comment below
```suggestion
    >>> # All the information we need can be found within the object "im".
```

keflavich:
@burnout87 could you rebase this so we get all the infrastructure updates included?

Sorry this has gone on long - I'll try to be responsive on this one and get it merged on your schedule.

keflavich:
It looks like the last commit did fix the docs issue, but now there are a few small other things that have cropped up.  I'll point them out inline.

astroquery/desi/core.py
```diff
@@ -0,0 +1,112 @@
...
+import urllib.error
+import requests
+import warnings
+
+import pyvo as vo
+import numpy as np
+import astropy.coordinates as coord
+
+from astropy import units as u
+from astropy.units import Quantity
```
keflavich, line 11: This import is unused, triggering a code style error: https://github.com/astropy/astroquery/actions/runs/3327073137/jobs/5501457364#step:5:19
```suggestion
```

astroquery/desi/tests/test_desi.py
```diff
@@ -0,0 +1,115 @@
...
+from contextlib import contextmanager
+
+
+DATA_FILES = {
+    'dummy_tap_table': 'dummy_table.txt',
+    'dummy_tractor_fits': 'dummy_tractor.fits',
+    'dummy_hdu_list_fits': 'hdu_list_images.fits'
+}
+
+coords = commons.ICRSCoord('11h04m27s +38d12m32s')
```
keflavich, line 23:
```suggestion
coords = coord.SkyCoord('11h04m27s +38d12m32s', frame='icrs')
```

astroquery/desi/core.py
```diff
@@ -0,0 +1,111 @@
...
+                    f"pixscale={pixsize}&"
+                    f"bands={image_band}")
+
+        file_container = commons.FileContainer(image_url, encoding='binary', show_progress=show_progress)
+
+        try:
+            fits_file = file_container.get_fits()
+        except (requests.exceptions.HTTPError, urllib.error.HTTPError) as exp:
+            fits_file = None
+            warnings.warn(f"{str(exp)} - Problem retrieving the file at the url: {image_url}", NoResultsWarning)
```
keflavich, line 106: Following @bsipocz 's unresolved comment: if the download was unsuccessful, we should raise an Exception.
```suggestion
            raise urllib.error.HTTPError(f"{str(exp)} - Problem retrieving the file at the url: {image_url}")
```

docs/desi/desi.rst
```diff
@@ -0,0 +1,97 @@
...
+    >>> ra = Angle('11h04m27s', unit='hourangle').degree
+    >>> dec = Angle('+38d12m32s', unit='hourangle').degree
+    >>> radius_input = 0.5
+    >>> pixels = 60
+    >>>
+    >>> pos = SkyCoord(ra, dec, unit='degree')
+    >>> radius = Angle(radius_input, unit='arcmin')
+    >>> im = DESILegacySurvey.get_images(pos, pixels, radius, data_release=9)
+
+    >>> All the information we need can be found within the object "im".
```
keflavich, line 62:
```suggestion
    >>> # All the information we need can be found within the object "im".
```

keflavich:
I'm a bit confused by our current error:
https://github.com/astropy/astroquery/actions/runs/3345568625/jobs/5541264991#step:5:673
It's an "unclosed file" `ResourceWarning: unclosed file <_io.BufferedReader name='/home/runner/work/astroquery/astroquery/.tox/py310-test-alldeps-devdeps-cov/lib/python3.10/site-packages/astroquery/desi/tests/data/hdu_list_images.fits'>`, but shouldn't the file get closed when it leaves scope?  I don't recall having to explicitly close `.fits` files (or load them with a context manager) in other packages.

astroquery/desi/core.py
```diff
@@ -0,0 +1,97 @@
...
+    def query_region(self, coordinates, radius, *, data_release=9):
+        """
+        Queries a region around the specified coordinates.
+
+        Parameters
+        ----------
+        coordinates : `astropy.coordinates`
+            coordinates around which to query.
+        radius : `astropy.units.Quantity`
+            the radius of the cone search.
```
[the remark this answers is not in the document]
keflavich: this is still incorrectly specified in the docstring

astroquery/desi/core.py
```diff
@@ -0,0 +1,111 @@
...
+from astroquery.query import BaseQuery
+from astroquery.utils import commons
+from astroquery.desi import conf
+
+__all__ = ['DESILegacySurvey', 'DESILegacySurveyClass']
+
+
+class DESILegacySurveyClass(BaseQuery):
+
+    def query_region(self, coordinates, radius=0.5*u.arcmin, *, data_release=9):
```
keflavich, line 21: There was a suggestion earlier to make `radius` an optional keyword argument, like this:
```suggestion
    def query_region(self, coordinates, *, radius=0.5*u.arcmin, data_release=9):
```
I resolved the conversation unintentionally, but I'll re-raise it now: is there any reason not to have it be optional?

astroquery/desi/tests/test_desi.py
```diff
@@ -0,0 +1,115 @@
...
+            assert np.array_equal(result[col], data[col])
+        else:
+            np.testing.assert_allclose(result[col], data[col])
+
+
+def image_tester(images, filetype):
+    assert type(images) == list
+    data = fits.open(data_path(DATA_FILES[filetype]))
+    assert images[0][0].header == data[0].header
+    assert np.array_equal(images[0][0].data, data[0].data)
```
keflavich, line 101: I don't think this should be necessary, but this is my attempt to fix the outstanding CI issue.  I'll merge it to see if it gets tests passing.
```suggestion
    with fits.open(data_path(DATA_FILES[filetype])) as data:
        assert images[0][0].header == data[0].header
        assert np.array_equal(images[0][0].data, data[0].data)
```

docs/desi/desi.rst
```diff
@@ -0,0 +1,97 @@
...
+    >>> dec = Angle('+38d12m32s', unit='hourangle').degree
+    >>> radius_input = 0.5
+    >>> pixels = 60
+    >>>
+    >>> pos = SkyCoord(ra, dec, unit='degree')
+    >>> radius = Angle(radius_input, unit='arcmin')
+    >>> im = DESILegacySurvey.get_images(pos, pixels, radius, data_release=9)
+
+    >>> # All the information we need can be found within the object "im".
+
```
keflavich, line 63: Not certain, but I think the whitespace here prevented the variable from being carried along to the next section, resulting in the current build failure.
```suggestion
    >>>
    >>> # All the information we need can be found within the object "im".
    >>>
```

astroquery/desi/core.py
eerovaher: In some modules the region can be a circle or a rectangle, and in those cases `radius` should definitely be a keyword-only argument. Although here the region can only be a circle, it might still better to make it a keyword-only argument for consistency's sake with the other sub-packages, and to make it very explicit to all users that the region is indeed a circle and not a rectangle.

keflavich: ...actually it is a (projected) square, so `radius` is the wrong kwarg.
```suggestion
    def query_region(self, coordinates, width=0.5*u.arcmin, *, data_release=9):
```
https://github.com/astropy/astroquery/pull/2210/files#diff-79317ce77daf56401a2a4150461a92c123e7061152b9219d313ac47515f9ad2aR44-R48

Probably this should be refactored to have `width` and `height`, and maybe even specifically `ra_width` and `dec_height`, as arguments.

bsipocz:
@burnout87 - This is rather close to be able to be merged, but currently there are a few cases where the code run into regression. Also, it would be nice to actually test more of the parameter space of the possible input values, e.g. the used data release is very much hardwired. I expect all this remaining stuff can be fixed with a reasonably straightforward commit, but do let us know if there are any questions.

emoravec:
Thanks very much for making DESI LS a part of astroquery. I have been using the branch that contains `DESILegacySurvey.query_region()` and have been using it over the past few weeks. I've noticed a bug. I have run the following command for about 20 different coordinate pairs that is along the lines of `DESILegacySurvey.query_region(coords, width=Angle(5, unit='arcmin'), data_release=9)`. The query table that it returns has the correct RA range (+- 5')  and the Dec range is + ~5 armin but minus typically ~60 _**degrees**_. Thus I am getting a lot of extra data and sometimes these queries take a really long time to get the data. See attached image (Example_query_no_Dec_lower_lim where the "box" (it doesn't look like a box due to the wide range of dec returned in the query) outlined in dotted lines is the region in which I would expect the query to return data ([RA-5',Dec-5'],[RA-5',Dec+5'],[RA+5',Dec+5'],[RA+5',Dec-5']) and the blue dots show the full range of data returned by the query.

I believe I’ve found the problem in the sql query in that there is no bound in the lower limit of Dec. 

Currently the sql query is:
`qstr = (f"SELECT all * FROM ls_dr{data_release}.tractor WHERE "
        f"dec<{(coordinates_transformed.dec + width).to_value(u.deg)} and "
        f"ra>{coordinates_transformed.ra.to_value(u.deg) - width.to_value(u.deg) / np.cos(coordinates_transformed.dec)} and "
        f"ra<{coordinates_transformed.ra.to_value(u.deg) + width.to_value(u.deg) / np.cos(coordinates_transformed.dec)}")`
        
But I _think_ it should be: 
`qstr = (f"SELECT all * FROM ls_dr{data_release}.tractor WHERE "
        f"dec<{(coordinates_transformed.dec + width).to_value(u.deg)} and "
        f"dec>{(coordinates_transformed.dec - width).to_value(u.deg)} and "
        f"ra>{coordinates_transformed.ra.to_value(u.deg) - width.to_value(u.deg) / np.cos(coordinates_transformed.dec)} and "
        f"ra<{coordinates_transformed.ra.to_value(u.deg) + width.to_value(u.deg) / np.cos(coordinates_transformed.dec)}")`

emoravec:
DR10 for DESI LS just came out in December 2022. It would be great to be able to access the DR10 data via a similar command `DESILegacySurvey.query_region(coords, width=Angle(30, unit='arcsec'), data_release=10)`.

PR #2290, opened 2022-02-09
Add how to change SIMBAD_URL to the docs

kelle:
Suggested change based on conversation in #493 .

### DIFF
--- a/docs/simbad/simbad.rst
+++ b/docs/simbad/simbad.rst
@@ -565,6 +565,17 @@ Astroquery configuration file.
 
     >>> from astroquery.simbad import Simbad
     >>> Simbad.TIMEOUT = 60 # sets the timeout to 60s
+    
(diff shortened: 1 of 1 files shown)


keflavich (approved):
This is a fine documentation improvement; we should _also_ fix the URL inheritance stuff to be a little more intuitive

bsipocz (changes requested):
We talked a lot lately about class and instance attributes and kind of agreed that the URL one should be one that the users are not expected to interact with, let alone changing it. 

So I don't think this should be added to the docs, use the config system instead as intended. And if you use this solution, do it as a workaround until the bug in the runtime change of config items is not fixed.

keflavich:
OK, that's fine - this is a workaround, not the solution.  It's good to have the issue open to remind us / show up on google.

PR #2374, opened 2022-04-25
A number of CADC authentication and doc fixes

andamian:
- Added fix to use an authentication connection with `datalink` service (and not only with the tap service)
- Fixed a problem to handle a different type of exception raised the upstream stack (`HTTPError`)
- A few fixes to make documentation examples testable.

### DIFF
--- a/astroquery/cadc/core.py
+++ b/astroquery/cadc/core.py
@@ -104,15 +104,15 @@ class CadcClass(BaseQuery):
 
         super(CadcClass, self).__init__()
         self.baseurl = url
+        # _auth_session contains the credentials that are used by both
+        # the cadc tap and cadc datalink services
         if auth_session:
             self._auth_session = auth_session
         else:
-            self._auth_session = None
+            self._auth_session = authsession.AuthSession()
 
     @property
     def cadctap(self):
-        if not self._auth_session:
-            self._auth_session = authsession.AuthSession()
         if not hasattr(self, '_cadctap'):
             if self.baseurl is None:
                 self.baseurl = get_access_url(self.CADCTAP_SERVICE_URI)
@@ -124,6 +124,13 @@ class CadcClass(BaseQuery):
                 self._cadctap = pyvo.dal.TAPService(
                     self.baseurl, session=self._auth_session)
         return self._cadctap
+
+    @property
+    def cadcdatalink(self):
+        if not hasattr(self, '_datalink'):
+            self._datalink = pyvo.dal.adhoc.DatalinkService(
+                self.data_link_url, session=self._auth_session)
+        return self._datalink
 
     @property
     def data_link_url(self):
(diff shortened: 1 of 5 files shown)


astroquery/cadc/core.py
```diff
@@ -596,6 +617,9 @@ def exec_sync(self, query, maxrec=None, uploads=None, output_file=None):
             Temporary tables to upload and run with the queries
         output_file: str or file handler:
```
bsipocz, line 618: nitpick: add spaces before `:`, here and elsewhere, to make this rendered properly

setup.cfg
```diff
@@ -32,7 +32,7 @@ astropy_header = true
 text_file_format = rst
 xfail_strict = true
 remote_data_strict = true
-addopts = --doctest-rst
+#addopts = --doctest-rst
```
bsipocz, line 35: Don't do this change, it's keep coming back, I suspect it maybe something outdated on the main brach of your fork?

bsipocz (changes requested):
This needs a rebase to get rid of the merge commit. I think that there maybe something off with on your fork as the change to `setup.cfg` keeps coming back.

Also, please add a changelog.

setup.cfg
andamian: Sorry - Didn't mean to commit this change.

andamian:
> This needs a rebase to get rid of the merge commit. I think that there maybe something off with on your fork as the change to `setup.cfg` keeps coming back.
> 
> Also, please add a changelog.

Do I need a changelog even if the API and functionality are essentially the same (just added an extra attribute with a default to a method)?

docs/cadc/cadc.rst
```diff
@@ -461,24 +461,15 @@ Query saving results in a file:
 1.5 Synchronous query with temporary uploaded table
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-A table can be uploaded to the server in order to be used in a query.
-
-
-.. doctest-skip::
-
-    >>> from astroquery.cadc import Cadc
```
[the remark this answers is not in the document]
andamian: You suggest I should added back with the `doctest-skip`?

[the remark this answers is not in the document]
bsipocz: No, I only set it to skip as the example was broken. (Well, we would need a dummy file anyway, but the problem wasn't with the file itself). So if you suggest to temporarily remove it I'm on board with that one,  but for the long term I would find it useful to have a working examples and not just the narrative description of how the file upload should work.

bsipocz:
> Do I need a changelog even if the API and functionality are essentially the same (just added an extra attribute with a default to a method)?

OK, it's fine not to add one if you don't expect this to be directly used by users.

docs/cadc/cadc.rst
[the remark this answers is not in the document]
andamian: Done. I hope it's the correct way.

astroquery/cadc/core.py
andamian: Done

bsipocz (approved):

bsipocz:
Thanks @andamian!

Version is on pypi.

PR #2376, opened 2022-04-25
Gaia astroquery 1.1

mhsarmiento:
Dear Astroquery Team,
This pull request contains a new feature in the Gaia module and an update in the default value of one of the parameters of the method 'load_data'
- Now the users of the Astroquery/Gaia module will know if there is a planned downtime of the Archive or if there is any issue on-going on the Gaia archive.
- For the method 'load_data' it has been updated the default value of the parameter 'valid_data' to 'False'.  With this change the epoch photometry service will returns all data associated to a given source. This change also mimics the behaviours of the current UI of the archive.

Thank you in advance,
all the best,
Maria

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,11 +4,19 @@
 New Tools and Services
 ----------------------
 
+gaia
+^^^^
+
+- TAP notifications service is now available for Gaia. If there is notification for the users,
+e.g planned or our unplanned downtimes of the archive, etc. The notification
+will be also visible when accessing the archive through Astroquery. [#2376]
 
 hsa
 ^^^
 
 - New module to access ESA Herschel mission. [#2122]
+
+
 
 Service fixes and enhancements
 ------------------------------
@@ -48,6 +56,13 @@ oac
 
 - Fix bug in parsing events that contain html tags (e.g. in their alias
   field). [#2423]
+
+gaia
+^^^^
+
+- Method 'load_data' now has the parameter 'valid_data' set to False by default.
+With this change the epoch photometry service returns all data associated to
+a given source. [#2376]
 
 
 Infrastructure, Utility and Other Changes and Additions
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,14 +8,13 @@ gaia
 ^^^^
 
 - TAP notifications service is now available for Gaia. If there is notification for the users,
-e.g planned or our unplanned downtimes of the archive, etc. The notification
-will be also visible when accessing the archive through Astroquery. [#2376]
+  for example planned or our unplanned downtimes of the archive, etc. The notification
+  will be also visible when accessing the archive through Astroquery. [#2376]
 
 hsa
 ^^^
 
 - New module to access ESA Herschel mission. [#2122]
-
 
 
 Service fixes and enhancements
@@ -61,8 +60,8 @@ gaia
 ^^^^
 
 - Method 'load_data' now has the parameter 'valid_data' set to False by default.
-With this change the epoch photometry service returns all data associated to
-a given source. [#2376]
+  With this change the epoch photometry service returns all data associated
+  to a given source. [#2376]
 
 
 Infrastructure, Utility and Other Changes and Additions
--- a/astroquery/gaia/__init__.py
+++ b/astroquery/gaia/__init__.py
@@ -39,6 +39,8 @@ class Conf(_config.ConfigNamespace):
                                       'MCMC_GSPPHOT',
                                       'MCMC_MSC']
 
+    GAIA_MESSAGES = _config.ConfigItem("notification?action=GetNotifications", "Gaia Messages")
+
 
 conf = Conf()
 
--- a/astroquery/gaia/__init__.py
+++ b/astroquery/gaia/__init__.py
@@ -39,8 +39,6 @@ class Conf(_config.ConfigNamespace):
                                       'MCMC_GSPPHOT',
                                       'MCMC_MSC']
 
-    GAIA_MESSAGES = _config.ConfigItem("notification?action=GetNotifications", "Gaia Messages")
-
 
 conf = Conf()
 
--- a/astroquery/gaia/core.py
+++ b/astroquery/gaia/core.py
@@ -920,6 +920,7 @@ class GaiaClass(TapPlus):
         the status of JWST TAP
         """
         try:
+            print("parsing notification messages")
             subContext = conf.GAIA_MESSAGES
             connHandler = self._TapPlus__getconnhandler()
             response = connHandler.execute_tapget(subContext, False)
--- a/astroquery/gaia/core.py
+++ b/astroquery/gaia/core.py
@@ -162,7 +162,7 @@ class GaiaClass(TapPlus):
         except HTTPError as err:
             log.error("Error logging out data server")
 
-    def load_data(self, ids, data_release=None, data_structure='INDIVIDUAL', retrieval_type="ALL", valid_data=True,
+    def load_data(self, ids, data_release=None, data_structure='INDIVIDUAL', retrieval_type="ALL", valid_data=False,
                   band=None, avoid_datatype_check=False, format="votable", output_file=None,
                   overwrite_output_file=False, verbose=False):
         """Loads the specified table
@@ -189,12 +189,10 @@ class GaiaClass(TapPlus):
             retrieval type identifier. For GAIA DR2 possible values are ['EPOCH_PHOTOMETRY']
             For future GAIA DR3 (Once published), possible values will be ['EPOC_PHOTOMETRY', 'RVS', 'XP_CONTINUOUS',
             'XP_SAMPLED', 'MCMC_GSPPHOT' or 'MCMC_MSC']
-        valid_data : bool, optional, default True
-            By default, the epoch photometry service returns only valid data,
-            that is, all data rows where flux is not null and
-            rejected_by_photometry flag is not true. In order to retrieve
-            all data associated to a given source without this filter,
-            this request parameter should be included (valid_data=False)
+        valid_data : bool, optional, default False
+            By default, the epoch photometry service returns all data associated to a given source.
+            In order to retrieve only valid data, that is, all data rows where flux is not null and
+            rejected_by_photometry flag is not true, this parameter should be included (valid_data=True)
         band : str, optional, default None, valid values: G, BP, RP
             By default, the epoch photometry service returns all the
             available photometry bands for the requested source.
@@ -250,9 +248,9 @@ class GaiaClass(TapPlus):
         params_dict = {}
 
         if not valid_data or str(retrieval_type) == 'ALL':
-            params_dict['VALID_DATA'] = "false"
-        elif valid_data:
             params_dict['VALID_DATA'] = "true"
+        elif valid_data:
+            params_dict['VALID_DATA'] = "false"
 
         if band is not None:
             if band != 'G' and band != 'BP' and band != 'RP':
--- a/astroquery/gaia/core.py
+++ b/astroquery/gaia/core.py
@@ -33,6 +33,7 @@ from astropy.table import Table
 from astropy import units as u
 import warnings
 from astroquery.exceptions import InputWarning
+from collections.abc import Iterable
 
 
 class GaiaClass(TapPlus):
@@ -925,10 +926,14 @@ class GaiaClass(TapPlus):
             connHandler = self._TapPlus__getconnhandler()
             response = connHandler.execute_tapget(subContext, False)
             if response.status == 200:
-                for line in response:
-                    string_message = line.decode("utf-8")
-                    print(string_message[string_message.index('=') + 1:])
-        except OSError as e:
+                if isinstance(response, Iterable):
+                    for line in response:
+                        try:
+                            print(line.decode("utf-8").split('=', 1)[1])
+                        except ValueError as e:
+                            print(e)
+                            pass
+        except OSError:
             print("Status messages could not be retrieved")
 
 
--- a/astroquery/gaia/core.py
+++ b/astroquery/gaia/core.py
@@ -45,6 +45,7 @@ class GaiaClass(TapPlus):
     MAIN_GAIA_TABLE_DEC = conf.MAIN_GAIA_TABLE_DEC
     ROW_LIMIT = conf.ROW_LIMIT
     VALID_DATALINK_RETRIEVAL_TYPES = conf.VALID_DATALINK_RETRIEVAL_TYPES
+    GAIA_MESSAGES = "notification?action=GetNotifications"
 
     def __init__(self, tap_plus_conn_handler=None,
                  datalink_handler=None,
@@ -52,7 +53,7 @@ class GaiaClass(TapPlus):
                  gaia_data_server='https://gea.esac.esa.int/',
                  tap_server_context="tap-server",
                  data_server_context="data-server",
-                 verbose=False, show_messages=True):
+                 verbose=False, show_server_messages=True):
         super(GaiaClass, self).__init__(url=gaia_tap_server,
                                         server_context=tap_server_context,
                                         tap_context="tap",
@@ -76,7 +77,7 @@ class GaiaClass(TapPlus):
             self.__gaiadata = datalink_handler
 
         # Enable notifications
-        if show_messages:
+        if show_server_messages:
             self.get_status_messages()
 
     def login(self, user=None, password=None, credentials_file=None,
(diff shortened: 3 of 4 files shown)


astroquery/gaia/__init__.py
```diff
@@ -39,6 +39,8 @@ class Conf(_config.ConfigNamespace):
                                       'MCMC_GSPPHOT',
                                       'MCMC_MSC']
 
+    GAIA_MESSAGES = _config.ConfigItem("notification?action=GetNotifications", "Gaia Messages")
```
bsipocz, line 42: What are the other possibilities for this configitem?

astroquery/gaia/core.py
```diff
@@ -911,5 +915,20 @@ def launch_job_async(self, query, name=None, output_file=None,
...
+        try:
+            subContext = conf.GAIA_MESSAGES
+            connHandler = self._TapPlus__getconnhandler()
+            response = connHandler.execute_tapget(subContext, False)
+            if response.status == 200:
+                for line in response:
+                    string_message = line.decode("utf-8")
+                    print(string_message[string_message.index('=') + 1:])
+        except OSError as e:
+            print("Status messages could not be retrieved")
```
bsipocz, line 931: Why the try/except, why not raise this as an exception instead? Is it expected to be unstable?

Also, no need for catching the exception (`as e`) if you don't use it later.

bsipocz (changes requested):
There are some commit duplications and merge commits, please rebase and maybe squash them down a bit, too.

astroquery/gaia/core.py
```diff
@@ -911,5 +915,20 @@ def launch_job_async(self, query, name=None, output_file=None,
...
+        the status of Gaia TAP
+        """
+        try:
+            subContext = conf.GAIA_MESSAGES
+            connHandler = self._TapPlus__getconnhandler()
+            response = connHandler.execute_tapget(subContext, False)
+            if response.status == 200:
+                for line in response:
+                    string_message = line.decode("utf-8")
+                    print(string_message[string_message.index('=') + 1:])
```
eerovaher, line 929: Equivalent, but shorter and clearer code:
```suggestion
                    print(line.decode("utf-8").split('=', 1)[1])
```

bsipocz:
A rebase went wrong here. Make sure you rebase on `astropy/main` instead of the main of your fork or esdc's fork.

bsipocz:
> A rebase went wrong here. Make sure you rebase on astropy/main instead of the main of your fork or esdc's fork.

OK, so I correct this to:

A rebase went wrong here. Make sure you rebase on the freshly fetched astropy/main instead of the main of your fork or esdc's fork.

astroquery/gaia/core.py
```diff
@@ -911,5 +915,20 @@ def launch_job_async(self, query, name=None, output_file=None,
                                         upload_table_name=upload_table_name,
                                         autorun=autorun)
 
+    def get_status_messages(self):
+        """Retrieve the messages to inform users about
+        the status of JWST TAP
```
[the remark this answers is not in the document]
mhsarmiento: Updated. Thank you

astroquery/gaia/core.py
```diff
@@ -74,6 +74,10 @@ def __init__(self, tap_plus_conn_handler=None,
         else:
             self.__gaiadata = datalink_handler
 
+        # Enable notifications
+        if show_messages:
```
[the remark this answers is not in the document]
mhsarmiento: Test added with show_message=True

astroquery/gaia/core.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, tap_plus_conn_handler=None,
                  gaia_data_server='https://gea.esac.esa.int/',
                  tap_server_context="tap-server",
                  data_server_context="data-server",
-                 verbose=False):
+                 verbose=False, show_messages=True):
```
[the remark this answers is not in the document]
mhsarmiento: Thanks for your comment. I have brought your question to the team for internal discussion. We will came back to you soon.

CHANGES.rst
```diff
@@ -3,21 +3,27 @@
...
 
+
+
 Service fixes and enhancements
 ------------------------------
 
-casda
-^^^^^
-
-- Add the ability to produce 2D and 3D cutouts from ASKAP images and cubes. [#2366]
```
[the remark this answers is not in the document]
mhsarmiento: Restored! Thanks

astroquery/gaia/__init__.py
mhsarmiento: Sorry. Not sure if I have understood your question. This is the way we use to add a modifier to the request.

astroquery/gaia/core.py
mhsarmiento: updated. Thank you!

astroquery/gaia/core.py
mhsarmiento: thanks for your inputs. Updated!

CHANGES.rst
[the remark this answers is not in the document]
eerovaher: The statement about not removing change log entries applies to all of them, not just this one entry individually.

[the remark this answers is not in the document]
eerovaher: The statement about not removing change log entries applies to all of them, not just this one entry individually.

astroquery/gaia/__init__.py
eerovaher: If this configuration item can only take one value then it is not configurable and should not be managed by the configuration system. So if you have implemented this as a configuration item then there must be more values it can take, but they are not documented anywhere.

eerovaher: If this configuration item can only take one value then it is not configurable and should not be managed by the configuration system. So if you have implemented this as a configuration item then there must be more values it can take, but they are not documented anywhere.

astroquery/gaia/core.py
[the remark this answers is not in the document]
eerovaher: I recommend you replace `show_messages` with something more descriptive so that it would be more obvious how it differs  from `verbose`.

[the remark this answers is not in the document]
eerovaher: I recommend you replace `show_messages` with something more descriptive so that it would be more obvious how it differs  from `verbose`.

astroquery/gaia/core.py
```diff
@@ -911,5 +916,24 @@ def launch_job_async(self, query, name=None, output_file=None,
...
+            response = connHandler.execute_tapget(subContext, False)
+            if response.status == 200:
+                if isinstance(response, Iterable):
+                    for line in response:
+                        try:
+                            print(line.decode("utf-8").split('=', 1)[1])
+                        except ValueError as e:
+                            print(e)
+                            pass
+        except OSError:
```
eerovaher, line 935: It is best to use the most specific error type possible. In this case it looks like `ConnectionError` or perhaps even one of its subclasses would be more appropriate.

eerovaher, line 935: It is best to use the most specific error type possible. In this case it looks like `ConnectionError` or perhaps even one of its subclasses would be more appropriate.

astroquery/gaia/tests/test_gaiatap.py
```diff
@@ -14,6 +14,7 @@
 
 
 """
+import unittest
```
eerovaher, line 17: The use of `unittest` was just removed by 6ec71cecceef295c78408aaa056eda706e467ea1, it should not be reintroduced without a very good reason.

eerovaher, line 17: The use of `unittest` was just removed by 6ec71cecceef295c78408aaa056eda706e467ea1, it should not be reintroduced without a very good reason.

eerovaher:
This pull request makes unannounced and unrelated changes to the XMM-Newton sub-package. It would be better to open a separate pull request for those.

CHANGES.rst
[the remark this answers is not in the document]
bsipocz: @eerovaher - This is a rebase issue, I can fix it when the PR otherwise ready to be merged rather than nitpick on the need of multiple round of rebases.

jespinosaar:
I have rebased again and removed the unnecessary commits, @mhsarmiento you can now proceed with the required changes.

cc @esdc-esac-esa-int

mhsarmiento:
Thanks @bsipocz  and @eerovaher for your support to the Gaia module in Astroquery. Your comments are very helpful and will be also applied to other ESA projects in Astrouquery. Those projects already pushed the same code to Astroquery to show the TAP notifications to the users (for example at esa/jwst module).

bsipocz:
I did some investigating to see what went wrong with the previous rebases and concluded that it was rebased on `esdc-esac-esa-int/main` instead of `astropy/main`. 
The issue is that the `esdc-esac-esa-int/main` is not tracking the canonical copy, but frozen in time to the point when the fork was created. So whenever you rebased on that branch, it will pull in a lot of extra commits and causing a conflicting nightmare.

My assumption is because of the confusing nature of calling the personal fork and these origin/upstream. I think that this git default naming is unfortunate and not is not fit for purpose for any workflow that deals with multiple accounts and is therefore a cause of great pain.

So to fix the workflow my suggestion is to rename all the remotes you work with to reflect the user/organization name on github. That way it's always clear which versions you are using as e.g. the base of a rebase. The workflow for a rebase, in that case, would be:

```
# check that all the remotes are set
$ git remote -v

esdc-esac-esa-int	git@github.com:esdc-esac-esa-int/astroquery (fetch)
esdc-esac-esa-int	git@github.com:esdc-esac-esa-int/astroquery (push)

# check that you're on the correct feature branch
$ git branch --show-current

gaia-astroquery-1.1

# Update all remotes. If you list all the branches, this command will update what's in `remotes/<username>/<branchname>`

$ git fetch --all  

Fetching astropy
Fetching esdc-esac-esa-int

$ # Optionally remove all outdated branches as well with the following. (also, strictly speaking --all is not mandatory, but I think it's good practice)
$ # git fetch --all --prune

# Do the rebase, it pops up the editor, select commits for squash/removal/move around. Fix conflicts and keep doing `git rebase --continue` until done.
$ git -rebase -i astropy/main

# when done, force push it to esdc... You don't need to be, but I think it's a good practice to be specific.
$ git push esdc-esac-esa-int HEAD:<name_of_branch> --force
```

Hope it helps.

(of course, the other approach is what you did, is to stash all commits done to one, but that also removes all granularity in the PR that may be useful later when one tries to trace the origin of a bug (which can be a bit problematic for very large PRs))

bsipocz:
Given that there are still issues with the changelog (some other unrelated entries got removed) I go ahead and push another rebase/squash approach that has the cleanup conflicts.

bsipocz (changes requested):
The only remaining item is the GAIA_MESSAGES conf item. Is there any situation where a user is expected to interact with it? If not, then hardwire it in `core.py` rather than as a confitem.

eerovaher:
The documentation needs to be updated, because the current code examples do not expect any notifications from the server, and I would rename the new `show_messages` argument to something like `show_server_notifications` so that it would be clear how it differs from the already existing `verbose` argument.

jespinosaar:
> I did some investigating to see what went wrong with the previous rebases and concluded that it was rebased on `esdc-esac-esa-int/main` instead of `astropy/main`. The issue is that the `esdc-esac-esa-int/main` is not tracking the canonical copy, but frozen in time to the point when the fork was created. So whenever you rebased on that branch, it will pull in a lot of extra commits and causing a conflicting nightmare.
> 
> My assumption is because of the confusing nature of calling the personal fork and these origin/upstream. I think that this git default naming is unfortunate and not is not fit for purpose for any workflow that deals with multiple accounts and is therefore a cause of great pain.
> 
> So to fix the workflow my suggestion is to rename all the remotes you work with to reflect the user/organization name on github. That way it's always clear which versions you are using as e.g. the base of a rebase. The workflow for a rebase, in that case, would be:
> 
> ```
> # check that all the remotes are set
> $ git remote -v
> 
> astropy	git@github.com:astropy/astroquery.git (fetch)
> astropy	git@github.com:astropy/astroquery.git (push)
> esdc-esac-esa-int	git@github.com:esdc-esac-esa-int/astroquery (fetch)
> esdc-esac-esa-int	git@github.com:esdc-esac-esa-int/astroquery (push)
> 
> # check that you're on the correct feature branch
> $ git branch --show-current
> 
> gaia-astroquery-1.1
> 
> # Update all remotes. If you list all the branches, this command will update what's in `remotes/<username>/<branchname>`
> 
> $ git fetch --all  
> 
> Fetching astropy
> Fetching esdc-esac-esa-int
> 
> $ # Optionally remove all outdated branches as well with the following. (also, strictly speaking --all is not mandatory, but I think it's good practice)
> $ # git fetch --all --prune
> 
> # Do the rebase, it pops up the editor, select commits for squash/removal/move around. Fix conflicts and keep doing `git rebase --continue` until done.
> $ git -rebase -i astropy/main
> 
> # when done, force push it to esdc... You don't need to be, but I think it's a good practice to be specific.
> $ git push esdc-esac-esa-int HEAD:<name_of_branch> --force
> ```
> 
> Hope it helps.
> 
> (of course, the other approach is what you did, is to stash all commits done to one, but that also removes all granularity in the PR that may be useful later when one tries to trace the origin of a bug (which can be a bit problematic for very large PRs))

Thanks for the analysis and the support @bsipocz! We already have the remotes with different names and a well-defined procedure documented, so maybe this has been just a small slip.

jespinosaar:
> The documentation needs to be updated, because the current code examples do not expect any notifications from the server, and I would rename the new `show_messages` argument to something like `show_server_notifications` so that it would be clear how it differs from the already existing `verbose` argument.

@mhsarmiento , why not 'show_tap_notifications'?

bsipocz:
> We already have the remotes with different names and a well-defined procedure documented, so maybe this has been just a small slip.

no worries, this takes time and practice for everyone, I feel the workflow got much easier already (at least on the reviewer end), so thank you!

astroquery/gaia/core.py
```diff
@@ -911,5 +916,25 @@ def launch_job_async(self, query, name=None, output_file=None,
                                         upload_table_name=upload_table_name,
                                         autorun=autorun)
 
+    def get_status_messages(self):
+        """Retrieve the messages to inform users about
+        the status of Gaia TAP
+        """
+        try:
+            print("parsing notification messages")
```
bsipocz, line 924: I know that many users don't like default verbosity. OTOH from the archive's side, I understand why these messages are desirable. 
So maybe a middle ground here to add another sentence to say how to turn it off?

Also, add a short paragraph about it in the narrative docs, users should know how to turn this off.

bsipocz:
>>    The documentation needs to be updated, because the current code examples do not expect any notifications from the server, and I would rename the new show_messages argument to something like show_server_notifications so that it would be clear how it differs from the already existing verbose argument.

> @mhsarmiento , why not 'show_tap_notifications'?


I don't think we have a precedent yet, in other modules (maybe @keflavich you can recall something). Do we know of any other servers where this would be useful?  But I would suggest using something that can be reused generally for the other, non-tap modules, too. I don't really have a strong preference, beyond the potential reusability of the name for other modules. What about `show_server_messages`?

astroquery/gaia/core.py
jespinosaar: This call to our TAP will tell the user if the TAP is under maintenance or it is not working, so they know if what they will do will be working in advance. But it is ok, we can add in the documentation how to turn it off. @mhsarmiento, could you please add this in the documentation of the python file and the .rst file?

jespinosaar:
> 

It is ok from our side.

astroquery/gaia/core.py
```diff
@@ -911,5 +917,29 @@ def launch_job_async(self, query, name=None, output_file=None,
...
+                    for line in response:
+
+                        try:
+                            print(line.decode("utf-8").split('=', 1)[1])
+                        except ValueError as e:
+                            print(e)
+                            pass
+                        except IndexError as e:
+                            print("Archive down for maintenance")
+                            pass
```
eerovaher, line 939: `pass` is a null operation which is meant for situations where the Python syntax requires an indented code block, but no code should be executed. There is no need to use `pass` here because the `print()` calls form the indented code blocks that Python expects. Furthermore, in the case of an `IndexError`, `e` would be an unused variable.
```suggestion
                        except ValueError as e:
                            print(e)
                        except IndexError:
                            print("Archive down for maintenance")
```

mhsarmiento:
> The only remaining item is the GAIA_MESSAGES conf item. Is there any situation where a user is expected to interact with it? If not, then hardwire it in `core.py` rather than as a confitem.

Thanks @bsipocz!. GAIA_MESSAGES has been removed as conf item and hardwired at 'core.py'. If in the future there is a situation in which the user can interact with that we will put it back as conf item but for now there isn't any plan for that.
Thanks again for your help.

astroquery/gaia/__init__.py
bsipocz: This has to be deleted now

bsipocz (approved):
One more comment, otherwise I think this is good to go. (The archive is now down, so I see test failures with the docs, but that is unrelated to this PR, so address any actual issues that may come up separately when DR3 is out.)

bsipocz:
I'm cleaning up the duplicate merges etc, and will do the __init__ cleanups and merge this.

mhsarmiento:
> One more comment, otherwise I think this is good to go. (The archive is now down, so I see test failures with the docs, but that is unrelated to this PR, so address any actual issues that may come up separately when DR3 is out.)

Thanks! I think that I forgot to commit this file. sorry for that. 

Yes, today at 12h CEST DR3 will be out! Very exciting times! :-)

bsipocz:
Thank you @mhsarmiento!

bsipocz:
@mhsarmiento @jespinosaar - should we change the default from DR2 to DR3 or it's too early?

PR #2477, opened 2022-08-02
Convert query_region to cone search

### LINKED ISSUE #585: Make SDSS query region a circle
This issue was discussed in #545, but we opted to do it as a separate fix so the other work in #545 could get in sooner.

Basically, the SDSS region queries (which have a `radius`) keyword in fact search a box instead of a circular region on the sky.  That needs to be fixed.  I think @weaverba137 _might_ be working on this?

weaverba137:
This PR closes #585.

`query_region()` now uses `query_crossid()` internally with slightly different parameters.  Effectively, a circular/cone region around each coordinate is searched instead of a rectangle in previous versions.

In addition, this PR adds lots of updates to tests, both online and offline, and cleans up the documentation of several of the methods.

Other comments:

* `query_region()` and `query_crossid()` accept `astropy.coordinates.Angle` for search radius as well as string or Quantity; internally, Angle is used to validate the search radius input and convert the angle to arcmin.
* The maximum search radius imposed by SDSS is 3 arcmin.  This is now enforced by the code.
* Added DR17 to offline tests; it had already been enabled for online tests.

### DIFF
--- a/astroquery/sdss/core.py
+++ b/astroquery/sdss/core.py
@@ -7,7 +7,7 @@ import warnings
 import numpy as np
 
 from astropy import units as u
-import astropy.coordinates as coord
+from astropy.coordinates import Angle
 from astropy.table import Table, Column
 
 from ..query import BaseQuery
@@ -27,6 +27,7 @@ sdss_arcsec_per_pixel = 0.396 * u.arcsec / u.pixel
 @async_to_sync
 class SDSSClass(BaseQuery):
     TIMEOUT = conf.timeout
+    MAX_CROSSID_RADIUS = 3.0 * u.arcmin
     QUERY_URL_SUFFIX_DR_OLD = '/dr{dr}/en/tools/search/x_sql.asp'
     QUERY_URL_SUFFIX_DR_10 = '/dr{dr}/en/tools/search/x_sql.aspx'
     QUERY_URL_SUFFIX_DR_NEW = '/dr{dr}/en/tools/search/x_results.aspx'
@@ -55,19 +56,21 @@ class SDSSClass(BaseQuery):
                            'qso_bal': [30, 31], 'qso_bright': 32
                            }
 
-    def query_crossid_async(self, coordinates, obj_names=None,
-                            photoobj_fields=None, specobj_fields=None,
-                            get_query_payload=False, timeout=TIMEOUT,
-                            radius=5. * u.arcsec,
+    def query_crossid_async(self, coordinates, radius=5. * u.arcsec, timeout=TIMEOUT,
+                            fields=None, photoobj_fields=None, specobj_fields=None, obj_names=None,
+                            spectro=False, region=False, field_help=False, get_query_payload=False,
                             data_release=conf.default_release, cache=True):
         """
         Query using the cross-identification web interface.
 
+        This query returns the nearest `primary object`_.
+
+        .. _`primary object`: https://www.sdss.org/dr17/help/glossary/#surveyprimary
+
         Parameters
         ----------
-        coordinates : str or `astropy.coordinates` object or list of
-            coordinates or `~astropy.table.Column` of coordinates The
-            target(s) around which to search. It may be specified as a
+        coordinates : str or `astropy.coordinates` object or list of coordinates or `~astropy.table.Column` of coordinates
+            The target(s) around which to search. It may be specified as a
             string in which case it is resolved using online services or as
             the appropriate `astropy.coordinates` object. ICRS coordinates
             may also be entered as strings as specified in the
@@ -77,13 +80,19 @@ class SDSSClass(BaseQuery):
             ra = np.array([220.064728084,220.064728467,220.06473483])
             dec = np.array([0.870131920218,0.87013210119,0.870138329659])
             coordinates = SkyCoord(ra, dec, frame='icrs', unit='deg')
-        radius : str or `~astropy.units.Quantity` object, optional The
-            string must be parsable by `~astropy.coordinates.Angle`. The
+        radius : str or `~astropy.units.Quantity` object or `~astropy.coordinates.Angle` object, optional
+            The string must be parsable by `~astropy.coordinates.Angle`. The
             appropriate `~astropy.units.Quantity` object from
-            `astropy.units` may also be used. Defaults to 5 arcsec.
+            `astropy.units` or `~astropy.coordinates.Angle` object from
+            `astropy.coordinates` may also be used. Defaults to 5 arcsec.
+            The maximum allowed value is 3 arcmin.
         timeout : float, optional
             Time limit (in seconds) for establishing successful connection with
             remote server.  Defaults to `SDSSClass.TIMEOUT`.
+        fields : list, optional
+            SDSS PhotoObj or SpecObj quantities to return. If None, defaults
+            to quantities required to find corresponding spectra and images
+            of matched objects (e.g. plate, fiberID, mjd, etc.).
         photoobj_fields : list, optional
             PhotoObj quantities to return. If photoobj_fields is None and
             specobj_fields is None then the value of fields is used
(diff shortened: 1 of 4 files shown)


keflavich:
This is a pretty big PR that looks good at a glance.  @bsipocz, will you want to do a detailed review?  I'm ready to approve this, but I haven't checked every line of the tests.

bsipocz:
Yes, I definitely want to review this with an ETA of maybe next week as this one is already been rather full.

astroquery/sdss/core.py
```diff
@@ -94,76 +103,86 @@ def query_crossid_async(self, coordinates, obj_names=None,
...
-        if specobj_fields is None:
-            if photoobj_fields is None:
-                photoobj_fields = crossid_defs
-            photobj_fields = ['p.{0}'.format(i) for i in photoobj_fields]
-            photobj_fields.append('p.objID as obj_id')
-            specobj_fields = []
+        if (not isinstance(coordinates, list) and
+            not isinstance(coordinates, Column) and
+            not (isinstance(coordinates, commons.CoordClasses) and
+                 not coordinates.isscalar)):
```
eerovaher, line 147: `astropy` has a fitting function for this: https://github.com/astropy/astropy/blob/c33e53f68388d4b4d68e85956bcfe4500a94c642/astropy/utils/misc.py#L45 The only case that needs to be handled separately is `str` because every `str` is iterable.

astroquery/sdss/core.py
```diff
@@ -94,76 +103,86 @@ def query_crossid_async(self, coordinates, obj_names=None,
...
+            coordinates = [coordinates]
+        if obj_names is None:
+            obj_names = ['obj_{0}'.format(i) for i in range(len(coordinates))]
```
eerovaher, line 150: Why not use f-strings here?
```suggestion
            obj_names = [f'obj_{i}' for i in range(len(coordinates))]
```

astroquery/sdss/core.py
```diff
@@ -94,76 +103,86 @@ def query_crossid_async(self, coordinates, obj_names=None,
...
+        elif len(obj_names) != len(coordinates):
+            raise ValueError("Number of coordinates and obj_names should "
+                         "be equal")
```
eerovaher, line 153: I like very explicit error messages.
```suggestion
            raise ValueError(
                f"number of coordinates {len(coordinates)} is not equal to number of object names {len(obj_names)}"
            )
```

astroquery/sdss/core.py
```diff
@@ -94,76 +103,86 @@ def query_crossid_async(self, coordinates, obj_names=None,
...
-        data = "obj_id ra dec \n"
-        data += " \n ".join(['{0} {1} {2}'.format(obj_names[i],
-                                                  coordinates[i].ra.deg,
-                                                  coordinates[i].dec.deg)
+            data = "name ra dec \n"  # SDSS's own examples default to 'name'.  'obj_id' is too easy to confuse with 'objID'
+            data_format = '{obj} {ra} {dec}'
+        data += " \n ".join([data_format.format(obj=obj_names[i],
+                                                ra=coordinates[i].ra.deg,
+                                                dec=coordinates[i].dec.deg)
                              for i in range(len(coordinates))])
```
eerovaher, line 163: The `list` created here is discarded after iterating over it once, so it would be better to use a generator expression here. Furthermore, using `zip` is preferable over accessing elements by index.
```suggestion
        data += " \n ".join(data_format.format(obj=obj,
                                               ra=coord.ra.deg,
                                               dec=coord.dec.deg)
                             for obj, coord in zip(obj_names, coordinates))
```

astroquery/sdss/core.py
```diff
@@ -1055,12 +1076,20 @@ def _args_to_payload(self, coordinates=None, radius=2. * u.arcsec,
                 raise ValueError('must specify at least one of `coordinates`, '
                                  '`run`, `camcol` or `field`')
 
-        sql = "{0} {1} {2} {3}".format(q_select, q_from, q_join, q_where)
-
-        request_payload = dict(cmd=sql, format='csv')
+        sql = "{0} {1} {2}".format(q_select, q_from, q_where)
```
eerovaher, line 1079:
```suggestion
        sql = f"{q_select} {q_from} {q_where}"
```

astroquery/sdss/tests/test_sdss.py
```diff
@@ -318,42 +309,186 @@ def test_query_crossid(patch_post, dr):
     url_tester_crossid(dr)
 
 
+def test_query_crossid_large_radius(patch_request):
+    """Test raising an exception if too large a search radius.
+    """
+    with pytest.raises(ValueError):
```
eerovaher, line 315: I suggest letting `pytest` [check the error message with the `match` argument](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions) to ensure you are indeed getting the `ValueError` you are expecting. Furthermore, the error message often conveys the purpose of the test clearly enough that an additional comment is not needed.

eerovaher (commented):
I have a few comments, but I wont claim to have done a thorough review.

astroquery/sdss/core.py
```diff
@@ -77,13 +80,19 @@ def query_crossid_async(self, coordinates, obj_names=None,
             ra = np.array([220.064728084,220.064728467,220.06473483])
             dec = np.array([0.870131920218,0.87013210119,0.870138329659])
             coordinates = SkyCoord(ra, dec, frame='icrs', unit='deg')
-        radius : str or `~astropy.units.Quantity` object, optional The
-            string must be parsable by `~astropy.coordinates.Angle`. The
+        radius : str or `~astropy.units.Quantity` object or `~astropy.coordinates.Angle` object, optional
```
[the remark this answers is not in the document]
weaverba137: True, but not every `Quantity` is an acceptable input either.  One way or another, it has to be an angular quantity.

astroquery/sdss/core.py
```diff
@@ -55,19 +56,21 @@ class SDSSClass(BaseQuery):
                            'qso_bal': [30, 31], 'qso_bright': 32
                            }
 
-    def query_crossid_async(self, coordinates, obj_names=None,
-                            photoobj_fields=None, specobj_fields=None,
-                            get_query_payload=False, timeout=TIMEOUT,
-                            radius=5. * u.arcsec,
+    def query_crossid_async(self, coordinates, radius=5. * u.arcsec, timeout=TIMEOUT,
+                            fields=None, photoobj_fields=None, specobj_fields=None, obj_names=None,
+                            spectro=False, region=False, field_help=False, get_query_payload=False,
```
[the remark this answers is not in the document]
weaverba137: OK, I can rearrange.

astroquery/sdss/core.py
```diff
@@ -94,76 +103,86 @@ def query_crossid_async(self, coordinates, obj_names=None,
...
+            True, objects will only count as a match if photometry *and*
+            spectroscopy exist. If False, will look for photometric matches
+            only.
+        region : bool, optional
+            Normally cross-id only returns the closest primary object.
+            Setting this to ``True`` will return all objects.
+        field_help: str or bool, optional
+            Field name to check whether a valid PhotoObjAll or SpecObjAll
+            field name. If `True` or it is an invalid field name all the valid
+            field names are returned as a dict.
```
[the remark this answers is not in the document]
weaverba137: I agree, but for historical reasons, that is how this is implemented. Perhaps changing the API in that manner should be a separate PR.

astroquery/sdss/core.py
```diff
@@ -94,76 +103,86 @@ def query_crossid_async(self, coordinates, obj_names=None,
...
 
-        if obj_names is None:
-            obj_names = ['obj_{0}'.format(i) for i in range(len(coordinates))]
-        elif len(obj_names) != len(coordinates):
-            raise ValueError("Number of coordinates and obj_names should "
-                             "be equal")
+        """
 
-        if isinstance(radius, u.Quantity):
+        if isinstance(radius, Angle):
```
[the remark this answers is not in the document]
weaverba137: If the radius is already an Angle, there is no need to test whether it is convertible to Angle.

astroquery/sdss/core.py
weaverba137: Historical reasons.  Is there an active policy in astroquery to convert to f-strings wherever possible?

astroquery/sdss/core.py
weaverba137: I was trying to make a minimal change.  Again, is there an explicit policy that replacing explicit lists with implicit iterables (which are less obvious) should be undertaken wherever possible?

astroquery/sdss/core.py
weaverba137: I was trying to make a minimal change. My personal experience is that PR reviewers are more likely to say "don't change what isn't broken".

astroquery/sdss/tests/test_sdss.py
weaverba137: I'll take a look at `match`, but honestly, I'm going to leave the docstring.

astroquery/sdss/core.py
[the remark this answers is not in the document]
bsipocz: please do not rearrange, rather make them mandatory kwarg. We incrementally doing this change in other modules, and examples like these are excellent cases where optional parameters should be forced to spelled out. 
(Of course it will be an API break, but I plan to have multiple API breaks in the next release anyway by doing significant cleanups).

astroquery/sdss/core.py
weaverba137: I'll give that a try.

astroquery/sdss/core.py
```diff
@@ -94,76 +103,86 @@ def query_crossid_async(self, coordinates, obj_names=None,
...
             radius = radius.to(u.arcmin).value
         else:
             try:
-                float(radius)
-            except TypeError:
+                radius = Angle(radius).to(u.arcmin).value
```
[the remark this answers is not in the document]
weaverba137: Is it measurably faster on a scalar?

astroquery/sdss/core.py
weaverba137: Again, minimal change.

astroquery/sdss/core.py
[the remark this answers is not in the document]
weaverba137: Is there a good example of that for me to look at?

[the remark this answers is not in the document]
bsipocz: Yeap, here is the PEP: https://peps.python.org/pep-3102/

For astroquery e.g. alma, esasky, mast are all moved on this approach.

In practice you add a `*` between positional and keyword arguments, and then can rearrange the order of the keyword ones as you wish (I fully agree that in many cases it makes sense to rearrange them, but if we don't enforce them to be keyword only, any addition is a breaking API change.)

astroquery/sdss/core.py
bsipocz: we incrementally cleaning up old python syntax, and as f-strings are cleaner to read moving towards using them more.

astroquery/sdss/core.py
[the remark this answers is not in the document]
eerovaher: I think using keyword-only arguments is a good idea, but it might be better to apply that change to the entire `sdss` subpackage at once instead of doing it one function at a time.

astroquery/sdss/core.py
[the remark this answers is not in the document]
eerovaher: If this is how it works already then a separate pull request would indeed be preferable.

astroquery/sdss/core.py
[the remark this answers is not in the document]
eerovaher: Yes, but the previous code also worked with a bare `Quantity` without trying to construct an `Angle` from it.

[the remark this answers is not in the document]
weaverba137: And I have not eliminated that functionality, if you examine the entire function.

astroquery/sdss/core.py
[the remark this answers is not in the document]
eerovaher: Using `IPython`:
```python
>>> from astropy import units as u
>>> angle = 1 * u.deg
>>> %timeit angle.to(u.rad).value
8.11 µs ± 199 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
>>> %timeit angle.to_value(u.rad)
4.58 µs ± 170 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

astroquery/sdss/core.py
weaverba137: Will do.

astroquery/sdss/core.py
eerovaher: If this is an already existing error message then it should be modified in a separate pull request.

astroquery/sdss/core.py
weaverba137: In the interest of getting this PR merged, I'm *not* going to attempt to change this, in the interest of minimizing additional changes.

astroquery/sdss/core.py
eerovaher: Pull requests are not required to update unrelated code, but these lines are being touched anyways.

Replacing list comprehensions with generator expressions is faster and uses less memory, but the values can be accessed only once. In this case the values are accessed only once anyways, so there really are no downsides to using a generator expression here.

Combining iterables with `zip` is faster than looping over indices. Using `IPython`:
```python
>>> word = "foo"
>>> letters = list(word)
>>> %timeit for i in range(len(word)): assert word[i] == letters[i]
392 ns ± 32.4 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
>>> %timeit for x, y in zip(word, letters): assert x == y
280 ns ± 2.81 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

By using an f-string the code snippet could be made even shorter:
```suggestion
        data += " \n ".join(f"{obj} {coord.ra.deg} {coord.dec.deg}" for obj, coord in zip(obj_names, coordinates))
```

astroquery/sdss/core.py
eerovaher: These lines are being edited anyways and the change is not user-facing.

weaverba137: Done

astroquery/sdss/core.py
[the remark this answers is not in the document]
eerovaher: The code still works with a bare `Quantity`, but it now converts a bare `Quantity` to an `Angle` even though none of the additional functionality implemented by `Angle` is required.

weaverba137:
In regards keyword-only arguments: I have converted the entire module except for the `get_images` and `get_spectra` functions, which have a fairly complicated call signature.  I think that is sufficient for the current PR, given that the original target of the PR was to convert to cone-search, not specifically to change to keyword-only arguments.

bsipocz (approved):
Overall it look good, bar one comment about the removal of `get_query_payload`, which I would like to be back, and to e.g. return the SQL string.
(It proved to be rather useful in many cases in the past.)

Anyway, I don't think it's worth dragging this largeish PR with that point further and will open an issue for a follow-up instead.

As the other review pointed out, there are a few places where a little bit of cleanup could be done, but I agree that that one is definitely beyond the scope here.

bsipocz:
Thanks @weaverba137!

bsipocz:
After merging this PR, I've gone through and triaged the `sdss` labelled issues, and happy to say it has fixed a bunch of them. There are still some long-reported bugs, but after the triage, we have half the number of open issues than before! 

🎉

pllim:
Thanks!

emirkmo:
Just out of curiosity, is it still possible to use astroquery to make larger radius queries similar to the box version previously used by astroquery.query_region?

You know how the saying goes with old code and user workflows…, but we had built an automated pipeline around large box size crossmatches 😅. 

the limit of 3 arcmin is way too small and needs a hefty workaround based on SDSS fields..

we can currently pin the old astroquery version and patch in future fixes that don’t touch this code but it’s not ideal.

 As a note from a user, it would have been nice to get a depreciation warning for an API breaking change like this, one that lasted a little while to give us time to adjust. It seems to me the switchover happened rapidly, and we have to do workarounds such as building the old functionality from the git history ourselves and patching in future fixes, or switch to our own SQL queries/casjobs. 
 
One way to enable backwards compatibility is to rename the old version to something like query_rectangle_region or to add a parameter such as rectangular: boolean = False, that preserves the breakage while the deprecation warning is active.

Anyway, hope this can be considered.

bsipocz:
@emirkmo - you can write any custom SQL query, so I would suggest looking into using `query_sql()` for larger sized queries?

emirkmo:
Thanks for the advice. It is actually easy to solve as you say, but moving to a custom SQL query is exactly what I was hoping to avoid by using astroquery. I made an issue to track what I guess is actually a feature request to re-enable to rectangular search, as it has a tangible benefit, and the old implementation in astroquery from a few months ago still just works.

pllim:
I am also interested in this request because the 3 arcmin limit seems pretty arbitrary and I am not sure how to explain to users why suddenly we have this limit downstream.

bsipocz:
cc @weaverba137

bsipocz:
Frankly, we can still roll back that API change (it hasn't landed in a tagged release), but I would need a bit more feedback (doing the proper conesearch was raised as a request a few times in the past...)


cc @keflavich

weaverba137:
3 arcmin is a *default*, but it is not *hard-coded* as far as I'm aware. What happens when you increase that limit?

keflavich:
I've been listening in on this, and I'd like to hear @weaverba137 's take.  I think rolling back the API change makes sense; probably best would be to allow either rectangular or circular searches, which is what we do with some other modules (vizier, I think?  I'd have to dig more).

Generally I agree that, for something as basic as a rectangle or circle query, we want to provide the simplified user interface and avoid having users write SQL.

bsipocz:
OK, but here the underlying difference is huge. E.g. for the rectangular one we just wrote a brute force +/- ra/dec box, while the radius search is using a different server-side service.

emirkmo:
> Frankly, we can still roll back that API change (it hasn't landed in a tagged release), but I would need a bit more feedback (doing the proper conesearch was raised as a request a few times in the past...)
> 
> cc @keflavich

I think the changes are actually very nice, especially with more proper reporting of errors and test failures. Run query also works as expected now. I would just appreciate having access to the rectangular search as well. I will leave any decisions up to the maintainers of course.

bsipocz:
I would see value in not having API change for the old setup, but then we run into the core of the previous issues, e.g. that `query_region` is not actually a cone search, but a box while a cone search is available. Adding a kwarg to fall back to the previous behaviour is still an API change, so while semantically would work, I'm looking for other options, too, maybe the we need a new function name for the current cone search-based implementation. (but I don't have any good ideas how to call it.)

emirkmo:
`query_cone_region`

`query_rectangle_region`

Is what I would look for in the docs (and first looked for).

Where either one or the other just calls `query_region` under the hood. That depends on whether backwards compatibility is more important or whether the function arguments making sense is more important.

The docstring for either can point to each other and mention the different endpoints used and the radius limitation.

bsipocz:
Yeah, let's wait for Ben's take and then we'll figure out something that brings in the least amount of API change.

keflavich:
If we do stick with the API change, we should include a deprecation warning that points users to the right way to do things.

bsipocz:
I'm starting to incline towards adding a kwarg to `query_region` that defaults to rectangular. Quite ugly, but that would be seamless, and no API change for all the previous use cases (I totally missed this server-side limitation on the query size, so the change doesn't feel like a simple backend refactoring any more to make the code more efficient and to fix a bug)

weaverba137:
Before I reply fully, I'm not entirely sure this is the thread we should continue the discussion on. For one, it's harder to find in the GitHub system since it's closed. Another open issue or PR might be easier for everyone to see.

emirkmo:
> Before I reply fully, I'm not entirely sure this is the thread we should continue the discussion on. For one, it's harder to find in the GitHub system since it's closed. Another open issue or PR might be easier for everyone to see.

I opened #2659

weaverba137:
Unless anyone objects, I am happy to move over to #2659.

bsipocz:
Yes, please move this convo to the issue linked above!

PR #2502, opened 2022-08-25
Recommend a better way to mark remote data tests in remote test modules

eerovaher:
Currently many test modules define a class that contains the test functions as methods. This has been convenient because marking the class with the `pytest.mark.remote_data` decorator applies the mark to all its methods, removing the need to mark each function individually. However, if all remote tests are in a separate module like the documentation already recommends then the [`pytestmark`](https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=pytestmark#globalvar-pytestmark) global variable could be used instead. The classes in remote test modules would not be needed and removing them would save one level of indentation.

This pull request updates the documentation to recommend `pytestmark`, and introduces it in the `atomic` sub-package as an example.

### DIFF
--- a/astroquery/atomic/tests/test_atomic_remote.py
+++ b/astroquery/atomic/tests/test_atomic_remote.py
@@ -6,8 +6,9 @@ from astropy.table import Table
 
 from ...atomic import AtomicLineList
 
+pytestmark = pytest.mark.remote_data
 
-@pytest.mark.remote_data
+
 def test_default_form_values():
     default_response = AtomicLineList._request(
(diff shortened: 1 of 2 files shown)


keflavich (approved):
lgtm.  We can slowly refactor other modules to follow this example in subsequent PRs

bsipocz (approved):
This looks good. My only reservation is that a few modules have a bit of a hit-and-miss for separating online and local tests, and using this approach may skip more tests than is absolutely necessary.  
The workaround for that is to be more watchful in review and spot when a test is being put in the incorrect test file.

PR #2504, opened 2022-08-26
converting input lists to np arrays

jaymedina:
`astroquery.mast.Observations.get_product_list` was mishandling lists because the assumed input was a numPy array. This led to incorrect requests being made to our API. This PR adds a conditional to catch for incoming lists to convert them into numPy arrays before handling. 

Before: 
--
image

After: 
--
image

### DIFF
--- a/astroquery/mast/observations.py
+++ b/astroquery/mast/observations.py
@@ -440,9 +440,11 @@ class ObservationsClass(MastQueryWithLogin):
             observations = np.array([observations])
         if isinstance(observations, Table):
             observations = observations['obsid']
+        if isinstance(observations, list):
+            observations = np.array(observations)
 
         observations = observations[observations != ""]
-        if len(observations) == 0:
+        if observations.size == 0:
             raise InvalidQueryError("Observation list is empty, no associated products.")
 
         service = self._caom_products
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,11 @@ hsa
 
 - New module to access ESA Herschel mission. [#2122]
 
+mast
+^^^
+
+- Fixed ``Observations.get_product_list`` to handle input lists of obsids. [#2504]
+
 
 Service fixes and enhancements
 ------------------------------
--- a/astroquery/mast/tests/test_mast.py
+++ b/astroquery/mast/tests/test_mast.py
@@ -441,6 +441,10 @@ def test_observations_get_product_list(patch_post):
     result = mast.Observations.get_product_list(observations[0:4])
     assert isinstance(result, Table)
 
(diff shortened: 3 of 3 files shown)


bsipocz (commented):
Please add a test and a changelog

astroquery/mast/observations.py
```diff
@@ -440,6 +440,8 @@ def get_product_list_async(self, observations):
             observations = np.array([observations])
         if isinstance(observations, Table):
             observations = observations['obsid']
+        if isinstance(observations, list):
+            observations = np.array(observations)
```
tomdonaldson, line 444: This looks good in that it will address the case where `observations` came in as a list.  I did notice a weird edge case however.

If observations comes in as an empty list, it gets converted to `array([], dtype=float64)`.  For that empty NDArray (and presumable empty NDArrays that weren't converted from lists), the next line has weird behavior:
observations = observations[observations != ""]
It issues a future warning, and returns an empty NDArray where `len(observations)==1`.  This makes the `len()==0` check fail, so there's a messy error later.

I'm not an expert on numpy array filtering, but if we're assuming that `observations` is a numpy array at this point, maybe we should do an explicit check for that after the if block, then maybe use `observations.size==0` instead of `len()==0`?

tomdonaldson: For the `observations.size==0`, the filtering could just be skipped I guess.

jaymedina: Thanks Tom! I agree - better to air on the safe side. I've added your two suggestions.

tomdonaldson: Thanks, @jaymedina !  In looking again, I think I oversimplified it a little.  I think there are other types of incoming `observations` that need to be covered (like Astropy Table or Row).  I would generally suggest using `instanceof` instead of `type() == ` to make sure inherited types are covered.   

I had been thinking that Astropy tables inherited from `np.ndarray`, but they don't, so if there's type checking here it needs to be a little messier, and worse, my suggestion of `observations.size` becomes bogus.  Sorry about that!!  I guess the bottom line is that since this method can accept multiple types for `observations`, we need to make sure that all those possibilities are handled cleanly.  The unit tests are probably a good place to ensure that handling.

jaymedina: Hi Tom, no worries, and yep I just realized that as well when trying to run my unit test and got an error for the Astropy Table and Row cases, so I decided to remove the assertion. I agree we can rely on the unit tests for these kinds of checks, so I added a new one for the input list case, and I think that should give us ample coverage. Thanks for your input!

bsipocz (approved):
All looks good, thanks!

PR #2525, opened 2022-09-12
Shorten `esa/jwst` and `gaia` tests

eerovaher:
Making a couple of small changes in the classes that are used for mocking server responses in `esa/jwst` and `gaia` tests made it possible to shorten the tests quite a lot. This makes the current tests simpler to understand and any future tests simpler to write. See the commit messages for details.


bsipocz (approved):
All looks good, thanks for the cleanup!

PR #2563, opened 2022-10-21
Test and refactor how `vizier` handles `query_region()` inputs with different units 

### LINKED ISSUE #2166: query_region_async in VizieR queries ignores some units
Looking at the code of `query_region_async` for VizieR, I noticed that some variables are not used. These are `o_unit` (the unit of the outer radius for queries including both the inner and the outer radius) and `h_unit` (the unit of the hight of a box for rectangular queries). In practice, the resulting queries are built only considering the unit of the inner radius (for cone queries) or the unit of the width of a box (for box queries). This clearly can lead to wrong requests if the units are different.

A minimal modification in the code should at least include a check that the units are the same; a slightly more advanced change would transform all quantities to a common unit.

eerovaher:
#2166 reported a bug in `vizier` about handling `query_region_async()` and `query_region()` inputs with different units. The code is actually already working like it is supposed to, but there were no unit tests that would verify this, and the fact that the code was misinterpreted suggested that it might benefit from being refactored.

Closes #2166.

### DIFF
--- a/astroquery/vizier/core.py
+++ b/astroquery/vizier/core.py
@@ -426,39 +426,18 @@ class VizierClass(BaseQuery):
             raise TypeError("Coordinates must be one of: string, astropy "
                             "coordinates, or table containing coordinates!")
 
-        # decide whether box or radius
         if radius is not None:
-            # is radius a disk or an annulus?
             if inner_radius is None:
-                radius = coord.Angle(radius)
-                unit, value = _parse_angle(radius)
-                key = "-c.r" + unit
-                center[key] = value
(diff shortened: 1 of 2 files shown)


keflavich (approved):
Looks like a reasonable cleanup & simplification to me.  I didn't realize you could nest `@parametrize` calls like that - good to know!

bsipocz (approved):
All looks good. It's borderline whether this needs a changelog or not. I'm somewhat thinking it's not necessary but could convince otherwise in a follow-up PR.

Thanks @eerovaher!

PR #2613, opened 2022-11-28
Adding HAPCut functionality to `astroquery.mast.Cutouts` tool

jaymedina:
This PR will expand `astroquery.mast`'s cutout functionality to support making Hubble Advanced Products (HAP) cutouts via MAST's [`HAPCut` API](https://mast.stsci.edu/hapcut/getting_started.html). 

! **Timeline: ASAB would like to see this merged before the end of Q1. Ideally before Dec 21st if possible.**

Changes include: 
- [x] New `HapcutClass` in `cutouts.py`
- [x] Unit testing for `HapcutClass` 
- [x] Updated documentation for `HapcutClass` functionality 
- [x] Updated release notes

## Example `download_cutouts` call with sample cutout shown: 

image

#### (default cutout size is 5 x 5 pixels): 

image

### DIFF
--- a/astroquery/mast/cutouts.py
+++ b/astroquery/mast/cutouts.py
@@ -201,7 +201,7 @@ class TesscutClass(MastQueryWithLogin):
         return Table(sector_dict)
 
     def download_cutouts(self, *, coordinates=None, size=5, sector=None, path=".", inflate=True,
-                         objectname=None, moving_target=False, mt_type=None):
+                         objectname=None, moving_target=False, mt_type=None, verbose=False):
         """
         Download cutout target pixel file(s) around the given coordinates with indicated size.
 
@@ -301,7 +301,7 @@ class TesscutClass(MastQueryWithLogin):
             localpath_table['Local Path'] = [zipfile_path]
             return localpath_table
 
-        print("Inflating...")
+        if verbose: print("Inflating...")
         # unzipping the zipfile
         zip_ref = zipfile.ZipFile(zipfile_path, 'r')
         cutout_files = zip_ref.namelist()
@@ -477,7 +477,8 @@ class ZcutClass(MastQueryWithLogin):
             warnings.warn("Coordinates are not in an available deep field survey.", NoResultsWarning)
         return survey_json
 
-    def download_cutouts(self, coordinates, *, size=5, survey=None, cutout_format="fits", path=".", inflate=True, **img_params):
+    def download_cutouts(self, coordinates, *, size=5, survey=None, cutout_format="fits", path=".", inflate=True,
+                         verbose=False, **img_params):
         """
         Download cutout FITS/image file(s) around the given coordinates with indicated size.
 
@@ -560,7 +561,7 @@ class ZcutClass(MastQueryWithLogin):
             localpath_table['Local Path'] = [zipfile_path]
             return localpath_table
 
-        print("Inflating...")
+        if verbose: print("Inflating...")
         # unzipping the zipfile
         zip_ref = zipfile.ZipFile(zipfile_path, 'r')
         cutout_files = zip_ref.namelist()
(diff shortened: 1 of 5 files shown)


jaymedina:
This question came up during one of our stand-ups: When will the next release of `astroquery` be?

docs/mast/mast.rst
```diff
@@ -1120,6 +1120,62 @@ To list the available deep field surveys at a particular location there is `~ast
...
+be returned as fits files (image files are not currently supported). This tool can be accessed in
+Astroquery by using the Hapcut class. Documentation for the supported HAPCut API can be found here:
+https://mast.stsci.edu/hapcut/
+
+
+Cutouts
+-------
+
+The `~astroquery.mast.HapcutClass.get_cutouts` function takes a coordinate and cutout size (in pixels or
+an angular quantity) and returns the cutout FITS file(s) as a list of ~astropy.io.fits.HDUList objects.
```
bsipocz, line 1137:
```suggestion
an angular quantity) and returns the cutout FITS file(s) as a list of `~astropy.io.fits.HDUList` objects.
```

docs/mast/mast.rst
```diff
@@ -1120,6 +1120,62 @@ To list the available deep field surveys at a particular location there is `~ast
    ['candels_gn_60mas', 'candels_gn_30mas', 'goods_north']
 
 
+HAPCut
```
bsipocz, line 1123: side note: the doctest are failing in an earlier example, so these has not been run during my testing

astroquery/mast/cutouts.py
```diff
@@ -637,3 +637,163 @@ def get_cutouts(self, coordinates, *, size=5, survey=None):
...
+
+        # Add size parameters to request
+        size_dict = _parse_cutout_size(size)
+        astrocut_request += f"&x={size_dict['x']}&y={size_dict['y']}&units={size_dict['units']}"
+
+        # Build the URL
+        astrocut_url = self._service_api_connection.REQUEST_URL + astrocut_request
+
+        # Set up the download path
+        path = os.path.join(path, '')
```
bsipocz, line 703: what's the purpose of this line?

astroquery/mast/cutouts.py
```diff
@@ -637,3 +637,163 @@ def get_cutouts(self, coordinates, *, size=5, survey=None):
...
+
+    def __init__(self):
+
+        super().__init__()
+
+        services = {"astrocut": {"path": "astrocut"}}
+
+        self._service_api_connection.set_service_params(services, "hapcut")
+
+    def download_cutouts(self, coordinates, *, size=5, path=".", inflate=True):
```
bsipocz, line 657: There seem to be a lot of duplicates in this method shared amongst the different cutout classes, I wonder maybe it's time to consolidate them into a base class, so any future maintenance becomes easier

astroquery/mast/cutouts.py
```diff
@@ -637,3 +637,163 @@ def get_cutouts(self, coordinates, *, size=5, survey=None):
...
+            with open(zipfile_path, 'r') as FLE:
+                response = json.load(FLE)
+            warnings.warn(response['msg'], NoResultsWarning)
+            return localpath_table
+
+        if not inflate:  # not unzipping
+            localpath_table['Local Path'] = [zipfile_path]
+            return localpath_table
+
+        print("Inflating...")
```
bsipocz, line 721: Don't use prints, us the logger or a verbose keyword instead. This should be changed in the other classes, too.

astroquery/mast/cutouts.py
```diff
@@ -637,3 +637,163 @@ def get_cutouts(self, coordinates, *, size=5, survey=None):
...
+        # so we can convert the dictionary into a JSON object
+        # in service_request_async(...)
+        param_dict["x"] = int(param_dict["x"])
+        param_dict["y"] = int(param_dict["y"])
+
+        # Adding RA and DEC to parameters dictionary
+        param_dict["ra"] = coordinates.ra.deg
+        param_dict["dec"] = coordinates.dec.deg
+
+        response = self._service_api_connection.service_request_async("astrocut",
```
bsipocz, line 772: extreme nitpick: fyi: we allow longer than 80char lines, no need to cut this into 3

astroquery/mast/cutouts.py
```diff
@@ -637,3 +637,163 @@ def get_cutouts(self, coordinates, *, size=5, survey=None):
...
+        response = self._service_api_connection.service_request_async("astrocut", 
+                                                                      param_dict, 
+                                                                      use_json=True)
+        response.raise_for_status()  # Raise any errors
+
+        try:
+            ZIPFILE = zipfile.ZipFile(BytesIO(response.content), 'r')
+        except zipfile.BadZipFile:
+            message = response.json()
+            warnings.warn(message['msg'], NoResultsWarning)
```
bsipocz, line 781: I think we should distinguish between an error in the response as opposed to a null response. I agree that it should not be an exception if there are no results, but if the response contains a genuine error message, we should raise that rather than warn about it.

astroquery/mast/cutouts.py
```diff
@@ -637,3 +637,163 @@ def get_cutouts(self, coordinates, *, size=5, survey=None):
...
+
+        response = self._service_api_connection.service_request_async("astrocut", 
+                                                                      param_dict, 
+                                                                      use_json=True)
+        response.raise_for_status()  # Raise any errors
+
+        try:
+            ZIPFILE = zipfile.ZipFile(BytesIO(response.content), 'r')
+        except zipfile.BadZipFile:
+            message = response.json()
```
bsipocz, line 780: Is there any reasons why response.json() rather than e.g. response.text is used?

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -1088,3 +1088,71 @@ def test_zcut_get_cutouts(self):
...
+    ###################
+    # HapcutClass tests #
+    ###################
+
+    def test_hapcut_download_cutouts(self, tmpdir):
+        
+        # Test 1: Simple API call with expected results
+        coord = SkyCoord(351.347812, 28.497808, unit="deg")
+
+        cutout_table = mast.Hapcut.download_cutouts(coordinates=coord, size=5, path=str(tmpdir))
```
bsipocz, line 1101: nitpick: ideally tests like these can be parametrized, so they would be independent and would not require repeated code

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -1088,3 +1088,71 @@ def test_zcut_get_cutouts(self):
...
+        assert isinstance(cutout_table, Table)
+        assert len(cutout_table) >= 1
+        for row in cutout_table:
+            assert os.path.isfile(row['Local Path'])
+            if 'fits' in os.path.basename(row['Local Path']):
+                assert fits.getdata(row['Local Path']).shape == (5, 5)
```
bsipocz, line 1107: could we check the shape for the other size inputs, too?

bsipocz (approved):
I have some big picture comments, which certainly points beyond this current PR.:

- I wonder at what point will it make sense to make a generic mast.Cutout class, that takes the coordinate and returns all available cutouts, or maybe also takes a missing name? Even if it's not exposed to the public API, I think we reached a point where having mast cutout base class would be beneficial as some of the methods share significant amount of code (e.g. enhancements and bugfixes, or any maintenance will be easier with a baseclass).


Otherwise I have minor comments, so would only like to see a few small things to be fixed.

docs/mast/mast.rst
```diff
@@ -1120,6 +1120,62 @@ To list the available deep field surveys at a particular location there is `~ast
...
+If the given coordinate appears in more than one product, a FITS file will be produced for each.
+
+.. doctest-remote-data::
+
+   >>> from astroquery.mast import Hapcut
+   >>> from astropy.coordinates import SkyCoord
+   ...
+   >>> cutout_coord = SkyCoord(351.347812, 28.497808, unit="deg")
+   >>> hdulist = Hapcut.get_cutouts(coordinates=cutout_coord, size=5)
+   >>> hdulist[0].info()    # doctest: +IGNORE_OUTPUT
```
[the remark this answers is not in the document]
jaymedina: The internal database that these calls access changes pretty regularly so I'll have to ignore the doctest.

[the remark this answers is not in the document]
bsipocz: but you're not testing the pixel data here, and I presume the dimension and format of the cutout should stay the same, would it not?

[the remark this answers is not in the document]
bsipocz: OK, thanks for showing the details. Agree on ignoring the output then.

astroquery/mast/cutouts.py
jaymedina: I don't think so, but I would defer to @ceb8 for that one. Using `response.text` requires you to index with an integer for the `msg` keyword value, so I'm guessing it's probably safer to index `response.json()` explicitly for `msg` rather than an implicit integer index of the `response.text` output.

bsipocz: OK, indeed makes sense. Thanks.

Maybe, can we add another assert, or even better add a `match=` in the `pytest.warn` line, so check we indeed parse out the expected message.

bsipocz: nope, keep the json, but add the match parameter to pytest.warns() for the tests, so it will only pass if the warning message also matching not just the type

astroquery/mast/tests/test_mast_remote.py
jaymedina: Thanks you just helped me catch a bug. Updated.

astroquery/mast/cutouts.py
jaymedina: Will do

astroquery/mast/cutouts.py
jaymedina: I made the distinction by way of an `if`/`else` statement. Let me know if this is what you were thinking of.

jaymedina:
> I wonder at what point will it make sense to make a generic mast.Cutout class, that takes the coordinate and returns all available cutouts, or maybe also takes a missing name? Even if it's not exposed to the public API, I think we reached a point where having mast cutout base class would be beneficial as some of the methods share significant amount of code (e.g. enhancements and bugfixes, or any maintenance will be easier with a baseclass).

This has actually been a topic of discussion recently, and we do have plans to consolidate all the cutouts classes into a single mission-agnostic cutout tool. I'm not sure what the timeline is will be on this, though. We have a few more things we want to update `astroquery.mast` with, including updating `TesscutClass` with support for a high level science product called TICA, as well as integrating `Cutouts` to the cloud, so consolidating all the classes into one will likely happen after that.

astroquery/mast/tests/test_mast_remote.py
jaymedina: Agreed, I think parametrization of some of these tests will be straight forward after consolidating all the classes into 1. One problem I see us running into is that some of the example inputs we use (mainly for the `coordinates` arg) only work in one tool and not the other, e.g. and RA/Dec of `102.7, 70.50` will work for TESSCut and not HAPCut. So I'll need to find a set of parameters that will work for all 3. But that's a bridge we can cross when we get to it.

bsipocz:
> This has actually been a topic of discussion recently ...

Sounds all great. It certainly points beyond this current PR, and any timeline will work that works for you.
Also re: cloud - as part of fornax (so mast is in the know), we're factoring out the cloud download logic from the mast module and upstream it into pyvo for a more generic audience, so it would be useful to communicate any plans for enhancements if it affects the bottom infrastructure.

astroquery/mast/tests/test_mast_remote.py
bsipocz: oh, I didn't mean to parametrize between classes, but for `size` within each of these tests, with a `pytest.mark.parametrize` (as practically that's the only thing changing within the test)

bsipocz:
Could you squash out some of the later debugging commits? I plan to come back and go through it again in the evening, but don't think there is any more things to address for merge.

jaymedina:
Yeah, I'm waiting for a science review on this from one group, once I hear back from them I'll squash this down so it's ready for merge.

jaymedina:
Squashed! @bsipocz

bsipocz:
Thanks! No need to squash to one, it's really more about keeping the incremental fixes at the later parts into more logical chunks. 
Either way, this is now good to go and I'll do it later today, but first will reuse a PR for an experiment for the gh command line client.

bsipocz:
Thank you @jaymedina!

PR #2660, opened 2023-02-09
IPAC/IRSA Moving Object Search Tool (MOST) interface

DinoBektesevic:
The PR adds support for MOST application at IPAC/IRSA. See the interface [here](https://irsa.ipac.caltech.edu/applications/MOST/).

There are some [instructions](https://irsa.ipac.caltech.edu/applications/MOST/instructions.html), but they do not always map to the exact key names and allowed values. There is an API for the service [here](https://irsa.ipac.caltech.edu/cgi-bin/MOST/nph-most?catalog=wise_merge&input_type=name_input&obj_name=Victoria&obs_begin=2014+05+01&obs_end=2014+05+30&output_mode=Brief) but because its output is the same as of the application itself the only functional difference is whether the search parameters are stringified into an URL fragment or not. 

Therefore I decided to query the application itself using a POST request. The returned response is parsed, and if HTML that HTML is parsed for the "Download *" links at the bottom of the page. These tables are then returned to the user. This makes "Regular" and "Full" output mode pretty much the same output mode, causing some potential confusion, but is the simplest solution.

I added documentation, could be improved. Same with the tests.

I wasn't sure what to log and and or how exactly to handle errors - I see there are custom made errors for different events but I was too lazy to peruse the docs to be honest. Happy to get any input on that. 

I also wasn't sure what all parameters should go into the `Config` class. So now some of them are in the default dict for the class and some are in `config` - happy to get advice on the recommended way forward.

There is a point to be made for unifying all of the same-meaning-but-different-named-parameters used by MOST (`input_type` but `output_mode`, `fits_region_files` for `create_tarballs` etc..) but that seemed like not the best idea - happy to hear otherwise.

EDIT: I see now I forgot to add `list_catalogs` and `get_images` as methods. Also, what should I name the main `query` method - I was not sure which example in the API Reference was the closest match?


DinoBektesevic:
I can't seem to reproduce [the error](https://github.com/astropy/astroquery/actions/runs/4559659088/jobs/8043843364?pr=2660#step:5:521) locally. 

Error:

```
UNEXPECTED EXCEPTION: ConnectionError(MaxRetryError("HTTPSConnectionPool(host='irsa.ipac.caltech.edu', port=443): Max retries exceeded with url: /cgi-bin/MOST/nph-most (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f04900aad90>: Failed to establish a new connection: An attempt was made to connect to the internet by a test that was not marked `remote_data`. The requested host was: 134.4.54.110'))"))
```

On the lines

```
353 .. doctest-remote-data::
354 
355     >>> from astroquery.ipac.irsa.most import Most
356     >>> Most.query_object(output_mode="Brief",
```

in `most.core.py`

astroquery/ipac/irsa/most/core.py
```diff
@@ -0,0 +1,1017 @@
+"""
+MOST
```
bsipocz, line 2: This narrative docs belongs to the rst file. We only auto generate the API docs at the bottom of it.
(moving this out should also fix the misparsing of most in the index page list)

astroquery/ipac/irsa/most/core.py
```diff
@@ -0,0 +1,1017 @@
...
+
+
+class MOSTClass(BaseQuery):
+    URL = conf.server
+    TIMEOUT = conf.timeout
+
+    def __init__(self):
+        super().__init__()
+
+    def _validate_name_input_type(self, params):
```
bsipocz, line 470: ~~these methods a little bit overkill as they are being used only once and they do a one liner, but now that we have it, they could stay.~~

OK, I take it back, I see that the mpc or manual inputs are a bit more complicated. I leave the previous comment here though in case someone else falls into the same review train of thought :)

astroquery/ipac/irsa/most/core.py
```diff
@@ -0,0 +1,1017 @@
...
+        # I think I saw somewhere that some password prompt will pop up for
+        # catalogs listed as '--- Internal use only:' but there are seemingly
+        # no limits to queries there.
+        if "--- Internal use only:" in catalogs:
+            catalogs.remove("--- Internal use only:")
+        return catalogs
+
+    def get_images(
+            self,
+            catalog="wise_merge",
```
bsipocz, line 675: mandate all optional kwargs to be kwarg only, here and in all the other methods, too (at minimum in the public ones).

astroquery/ipac/irsa/most/core.py
```diff
@@ -0,0 +1,1017 @@
...
+            arg_perihelion=None,
+            ascend_node=None,
+            semimajor_axis=None,
+            mean_anomaly=None,
+            perih_dist=None,
+            perih_time=None,
+            get_query_payload=False,
+            save=False,
+            savedir='',
+    ):
```
bsipocz, line 697: I really don't like black's love of whitespace and this way of 20+ lines instead of ~4 for the signature. Can we put some sense back in it, into ~100 char long lines instead?

astroquery/ipac/irsa/most/core.py
```diff
@@ -0,0 +1,1017 @@
...
+            semimajor_axis=semimajor_axis,
+            mean_anomaly=mean_anomaly,
+            perih_dist=perih_dist,
+            perih_time=perih_time,
+            get_query_payload=get_query_payload,
+            output_mode="Regular",
+            with_tarballs=True,
+        )
+
+        request = self._request("GET", queryres["fits_tarball"],
```
bsipocz, line 820: super nitpick as this level of consistency doesn't matter, but I believe we tend to call this `response` instead of `request`.

astroquery/ipac/irsa/tests/test_most.py
bsipocz, line 1: all the tests are mocked here, there should be a counterpart test file `test_most_remote.py` that actually queries the server.

(some remote coverage is already present in the narrative docs, but a bit more formalised tests would be great, too)

docs/ipac/irsa/most/most.rst
```diff
@@ -0,0 +1,4 @@
+.. automodule:: astroquery.ipac.irsa.most.core
```
bsipocz, line 1: narrative should come here instead of this.

bsipocz (commented):
Overall this PR is very close to be ready.

Leaving some big picture comments, they are mostly logistics and nitpicks. I'll do a run on the docs once it's been moved around.

astroquery/ipac/irsa/most/core.py
```diff
@@ -416,7 +416,7 @@
     'https://irsa.ipac.caltech.edu/workspace/TMP_X69utS_13312/MOST/pid15792/fitsimage_A850RA.tar.gz'
     >>> matched["region_tarball"]  # doctest: +IGNORE_OUTPUT
     'https://irsa.ipac.caltech.edu/workspace/TMP_X69utS_13312/MOST/pid15792/ds9region_A850RA.tar'
-    >>> matched["results"].columns
+    >>> matched["results"].columns  # doctest: +IGNORE_OUTPUT
```
bsipocz, line 419: This doesn't trick the linter, so I would remove the commit, but then the linter won't run on the rst files either.

(otherwise the doctest infra is not sensitive to whitespace, so you can cut up the output into two lines)

astroquery/ipac/irsa/most/__init__.py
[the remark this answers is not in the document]
DinoBektesevic: I didn't know what to do about `Conf` so I merged them. Was that expected?

astroquery/ipac/irsa/most/core.py
DinoBektesevic: Yes, the methods are broken out for consistency and clarity more than for necessity. Too many deeply nested `if`s.

astroquery/ipac/irsa/most/core.py
DinoBektesevic: The way this reads to me is as if you want me to make optional keyword arguments just keyword arguments - in which case every invocation of the method will require the user to provide them and that sounds wrong? Alternative understanding is to user `**kwargs` but that sounds wrong too? Could you clarify what you meant?

docs/ipac/irsa/most/most.rst
```diff
@@ -0,0 +1,449 @@
...
+mjd_obs        Modified Julian Date (days) of the midpoint of the observation.
+ra_obs         Ephemeris of the object's right ascension (deg, J2000).
+dec_obs        Ephemeris of the object's declination (deg, J2000).
+sun_dist       Heliocentric distance of the object (AU).
+geo_dist       Geocentric distance of the object (AU).
+dist_ctr       Projected distance  from the center of the image (deg).
+phase          Sun-Object-earth angle (deg).
+vmag           Estimate of visual magnitude (from Horizons).
+image_url      Links to download or view the data.
+postcard_url   ??? usually ``null``
```
DinoBektesevic, line 100: I wanted to leave a comment to point out that I added some most basic error checks to the query - errors are re-raised but when no results are found I return `None`. Could not make out the standard expected behavior from other classes. 

Additionally, I still don't know what a `postcard_url` is - usually it's just `null`. This should be clarified before merge I'd think.

astroquery/ipac/irsa/tests/test_most.py
```diff
@@ -0,0 +1,249 @@
...
+
+    assert "results" in response
+    assert "metadata" in response
+    assert "region" in response
+    assert "fits_tarball" not in response
+    assert "region_tarball" not in response
+
+    results = Table.read(data_path("most_regular_results.tbl"), format="ipac")
+    metadata = Table.read(data_path("most_imgframes_matched_final_table.tbl"), format="ipac")
+    url = REGULAR_BASE_URL + "ds9region/ds9_orbit_path.reg"
```
bsipocz, line 180: Super nitpick on naming things: maybe call these expected, e.g. expected_result, and the `response` above as `result`, too? As it's not the response, but the final result, e.g. the parsed query respose.

astroquery/ipac/irsa/tests/test_most.py
```diff
@@ -0,0 +1,249 @@
...
+
+import io
+import os
+import pytest
+
+from astropy.io import votable
+from astropy.table import Table
+from astroquery.utils.mocks import MockResponse
+from astroquery.ipac.irsa import Most
+from astropy.utils.diff import report_diff_values
```
bsipocz, line 11: hmm, interesting find, I wasn't aware of this one.

astroquery/ipac/irsa/tests/test_most_remote.py
```diff
@@ -0,0 +1,129 @@
...
+    assert "region" in response
+    assert "fits_tarball" not in response
+    assert "region_tarball" not in response
+
+    # check content is as expected for the columns we don't expect to change
+    # This excludes URLS, they contain PIDs. The rest can be compared, but the
+    # problem is that report_diff_values doesn't sort the values and they are
+    # not returned in same order nor header width. So we have to compare
+    # manually
+    results = Table.read(data_path("most_regular_results.tbl"), format="ipac")
```
bsipocz, line 47: again, call it expected, etc from the non remote tests comments

astroquery/ipac/irsa/tests/test_most_remote.py
```diff
@@ -0,0 +1,129 @@
...
+    ]
+
+    cats = Most.list_catalogs()
+    assert expected == cats
+
+
+@pytest.mark.remote_data
+def test_no_results():
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
```
bsipocz, line 95: nope, this should be pytest.warns. You can drop the `match` part, but test with the warning type. Here, and below, too.

```suggestion
    with pytest.warns(NoResultsWarning, match='Number of Matched Image Frames   = 0'):
       
```

docs/ipac/irsa/most/most.rst
```diff
@@ -0,0 +1,449 @@
...
+    The URLs returned by the results table, however, point to the NASA/IPAC
+    Infrared Science Archive, which means that the URLs to the images themselves
+    will remain valid even after the Most URLs expire.
+
+
+Reference/API
+=============
+
+.. automodule:: astroquery.ipac.irsa.most
+   :members:
```
bsipocz, line 449: This is my suspect for the weird looking docs. We do use automodapi. that provides a much nicer API reference, but given this is pulled into the main IRSA namespace, it's already been API docs-ed there. So I would say, drop the reference here, and we'll figure out whether to link it again from the other namespace, or even include this page in there?

```suggestion

```

bsipocz (commented):
A lot of nitpicks, with a few genuine fixes to be made, mostly around the tests, but all of them are very straightforward.

And this needs a changelog entry. 

Also, something looks weird with the rendered API docs in the most.rst page, but it all renders well in the one layer up namespace.

astroquery/ipac/irsa/most.py
```diff
@@ -0,0 +1,559 @@
...
+        Raises
+        ------
+        ValueError
+            If the input does not have the minimum required parameters set to
+            an at least truthy value.
+        """
+        obj_type = params.get("obj_type", False)
+        if not obj_type:
+            raise ValueError("When input type is 'mpc_input' key 'obj_type' is required.")
+        if obj_type not in ("Asteroid", "Comet"):
```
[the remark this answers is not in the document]
DinoBektesevic: For future reference

```
================================================================================
 
Toolkit version: N0065
 
SPICE(SPKINSUFFDATA) --
 
Insufficient ephemeris data has been loaded to compute the state of -99999
relative to 0 (SOLAR SYSTEM BARYCENTER) at the ephemeris epoch 2010 JAN 07
00:01:06.184.
 
A traceback follows.  The name of the highest level module is first.
spkez_c --> SPKEZ --> SPKACS --> SPKAPS --> SPKLTC --> SPKGEO
 
================================================================================
```

are the kinds of errors that pop up if they are not first-letter-capitalized. The upstream is basically just a form whose payload is directly delivered to SPICE toolkit and their internal search engine. We can fix this, as well as rename some of these parameters to be more consistent in the `qparams` map (like why arg_perihelion but perih_dist and perih_time? Shouldn't it all be short or all of it long. Same with the input_type and output_mode - an arbitrary change that makes it, in my humble opinion, much harder to remember how to use the thing. Most of the `obj_` prefixes could be dropped from the parameter's name too. If this is getting passed onward, something to think about.

astroquery/ipac/irsa/most.py
```diff
@@ -0,0 +1,559 @@
...
+        if not params.get("body_designation", False):
+            params["body_designation"] = "Test"+params["obj_type"]
+
+    def _validate_input(self, params):
+        """
+        Validate the minimum required set of parameters, for a given input
+        type, are at least truthy.
+
+        These include the keys `catalog`, `input_type`, `output_mode` and
+        `ephem_step` in addition to keys required by the specified input type.
```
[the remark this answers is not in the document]
DinoBektesevic: I don't know how I missed so many of these, it must be because they're private methods so docs never yelled back at me. Thanks!

astroquery/ipac/irsa/most.py
```diff
@@ -0,0 +1,559 @@
...
+        if obj_type == "Asteroid":
+            if not params.get("semimajor_axis", False):
+                raise ValueError("When obj_type is 'Asteroid', 'semimajor_axis' is required.")
+        elif obj_type == "Comet":
+            if not params.get("perih_dist", False):
+                raise ValueError("When obj_type is 'Comet', 'perih_dist' is required.")
+
+        # This seemingly can be whatever
+        if not params.get("body_designation", False):
+            params["body_designation"] = "Test"+params["obj_type"]
```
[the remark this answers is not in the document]
bsipocz:
```
body_designation is an identifying string less than 32 characters the user chooses, required for orbital element input (input_type = manual_input).
```

astroquery/ipac/irsa/most.py
```diff
@@ -0,0 +1,559 @@
...
+        html = BeautifulSoup(response.content, "html5lib")
+        catalog_dropdown_options = html.find("select").find_all("option")
+
+        catalogs = [tag.string for tag in catalog_dropdown_options]
+        # I think I saw somewhere that some password prompt will pop up for
+        # catalogs listed as '--- Internal use only:' but there are seemingly
+        # no limits to queries there.
+        if "--- Internal use only:" in catalogs:
+            catalogs.remove("--- Internal use only:")
+        return catalogs
```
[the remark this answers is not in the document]
bsipocz:
```
The Internal-Use-only datasets are free to search in MOST.  The way it is supposed to work is that the images will not be accessible.
```

astroquery/ipac/irsa/most.py
```diff
@@ -0,0 +1,559 @@
...
+            containing at least ``results``, ``metadata`` and ``region`` keys,
+            and optionally ``fits_tarball`` and ``region_tarball`` keys. When
+            in ``"Brief"`` or ``"Gator"`` an `~astropy.table.Table` object and
+            in ``"VOTable"`` an `~astropy.io.votable.tree.VOTableFile`. See
+            module help for more details on the content of these tables.
+        """
+        # This is a map between the keyword names used by the MOST cgi-bin
+        # service and their more user-friendly names. For example,
+        # input_type -> input_mode or fits_region_files --> with tarballs
+        qparams = {
```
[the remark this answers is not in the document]
DinoBektesevic: I kept tripping over `input_mode` is actually `input_type` and then the `ouput_mode` is correct, not `output_type` and that bothering me. Also `fits_region_files` is not instructive and giving a `"on"` and `""` as strings seems very unpythonic, bool is better.. It'd be nice to simplify some of these names too, but I didn't want to steer too far away from the interface.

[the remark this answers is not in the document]
bsipocz: Oh, but then you can do those in a separate few liners, no need to do the full dictionary approach. (the dict very much hides these differences anyway, I didn't spot the difference between `type` and `mode` only the `with_tarball`.

astroquery/ipac/irsa/most.py
```diff
@@ -0,0 +1,559 @@
...
+            "semimajor_axis": semimajor_axis,
+            "mean_anomaly": mean_anomaly,
+            "perih_dist": perih_dist,
+            "perih_time": perih_time,
+        }
+
+        if get_query_payload:
+            return qparams
+
+        self._validate_input(qparams)
```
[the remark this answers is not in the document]
DinoBektesevic: I can unravel the checks if you want, there'd be a few nested ifs but trimming out required parameters perhaps would make it more readable. It's unfortunate but the online interface just punts the parameters straight onwards into other services and not all of them have as helpful error message as SPICE error messages are.
Your choice. 

```

Object Name Error:

User Input String=

HORIZONS name lookup failed

Workspace URL: https://irsa.ipac.caltech.edu/workspace/TMP_UOMGHF_19078/MOST/pid19078 
```

astroquery/ipac/irsa/most.py
```diff
@@ -0,0 +1,559 @@
...
+        response = self._request("POST", self.URL,
+                                 data=qparams, timeout=self.TIMEOUT)
+
+        # service unreachable or some other reason
+        if not response.ok:
+            raise response.raise_for_status()
```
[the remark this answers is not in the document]
DinoBektesevic: Funny. Surely, though, you mean it without the `raise` call :)

astroquery/ipac/irsa/most.py
```diff
@@ -0,0 +1,559 @@
...
+            raise InvalidQueryError(response.text)
+
+        # presume that response is HTML to simplify conditions
+        if "Number of Matched Image Frames   = 0" in response.text:
+            warnings.warn("Number of Matched Image Frames   = 0", NoResultsWarning)
+            return None
+
+        if qparams["output_mode"] in ("Brief", "Gator"):
+            return Table.read(response.text, format="ipac")
+        elif qparams["output_mode"] == "VOTable":
```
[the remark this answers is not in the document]
bsipocz: then drop Full and use votable. Maybe not even expose this to the users? 
(and then parse the votable and return it as an astropy Table)

E.g. is there a use case to use either Brief or Gator?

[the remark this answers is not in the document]
DinoBektesevic: I don't know, given that all these different output formats, contents and types exist I would assume they do? I'm not a power user of the interface and really only rely on finding the ids and locations of the images from it.

astroquery/ipac/irsa/tests/test_most.py
DinoBektesevic: Secret wonders hide in the depths of various utils modules in this wide bright world.

astroquery/ipac/irsa/most.py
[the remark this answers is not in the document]
DinoBektesevic: I kind of need the dict for the `self._request("POST", self.URL, data=qparams, timeout=self.TIMEOUT)` call couple of lines later? Unless there's a different way of invoking a `self._request` that I didn't correctly understand from the manual. Or is there a clever way of creating a dict where variable names and values become keys and values?

docs/ipac/irsa/most/most.rst
DinoBektesevic: I think the weirdness is all due to the fact that MOST doesn't behave like a simple extension of the core general catalog query system implemented by the IRSA module, so it's difficult so it's difficult to reconcile these differences if they occupy the same namespace. The use of MOST as a service is not simple nor straight-forward because what it returns changes depending on what the query is. This puts the onus of understanding what happens in the back-end on the user. Some of this is hidden by the verification system, forcing user to be capitalization aware for example, some of it by the subtle different mapping between argument names in the `query_region` method and online form. Certainly including this into IRSA would at least double the length of that page.

Up to you. I have no hard preferences about it but it's a tough question. Removed as requested for now.

astroquery/ipac/irsa/most.py
```diff
@@ -0,0 +1,559 @@
...
+        ValueError
+            If the input does not have the minimum required parameters set to
+            an at least truthy value.
+        """
+        if params.get("catalog", None) is None:
+            raise ValueError("Which catalog is being queried is always required.")
+
+        input_type = params.get("input_type", None)
+        if input_type is None:
+            raise ValueError("Input type is always required.")
```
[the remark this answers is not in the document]
DinoBektesevic: Sometimes, when they are the parameters that go into SPICE Toolkit, because SPICE has nice error messages. Other times you see unhelpful messages such as: https://github.com/astropy/astroquery/pull/2660#discussion_r1231377611

bsipocz (approved):
I've added one more commit that makes all remote doctests pass on my laptop. We'll likely need to add a few more `IGNORE_OUTPUTS` (currently only added it for those that print out the temporary workspace url, that changes for each query).
If tests keep failing on the cron CI, but for now it all looks fine.

bsipocz:
Thank you so much @DinoBektesevic, and I'm sorry for dragging it along for so long.

PR #2663, opened 2023-02-10
Restore rectangle search

### LINKED ISSUE #2659: Re-enable rectangular search for SDSS query_region
#2477 Removed rectangular search from `query_region` in `SDSS` and converted it to a cone search. However, a drawback was not significantly discussed:

- The rectangular search can use a much larger "radius" i.e. field size. The limit for a cone search is 3 arcmin while the limit for a rectangular search can be as large as 12 arcmin (or more).

With the sparse fields we usually query for extragalactic transients, this does not usually get close to row-size limits. 

The changes in #2477 were made without Depreciation warnings as well, which I understand with the pre-release model.. however it would have been nice to at least preserve the old endpoint in  function such as `query_rectangle_region` or in a parameter such as `rectangular: bool = False` (where setting it to True uses the old `query_region`) so that backwards compatibility can be maintained.

Anyway this is a Feature Request to add this back in, as the old version still works (can be tested by pinning `astroquery` to e.g., `v0.4.6`, but making the change to the endpoint being HTTPS, a recent fix from #2654):

```Python
from astroquery import sdss
sdss.conf.skyserver_baseurl = sdss.conf.skyserver_baseurl.replace("http://","https://")
working_sdss_table = sdss.SDSS.query_region(<pars>,radius=<large radius multiple times over the MAX_CROSSID_RADIUS>) 
```

In fact in SDSS rectangular search there is only a suggestion, not a hard limit, on "radius" of 0.2 degrees, which seems to be not enforced by radius but by rows returned. See e.g.,: https://skyserver.sdss.org/dr17/SearchTools/rect

weaverba137:
This PR closes #2659. Rectangle searches are passed as a SQL query to `query_sql()`.

- [X] Fix any existing broken tests.
- [x] Enable tests with rectangular search.
- [x] Test calling `query_region()` with either `radius` or `width`; not neither, not both.
- [x] Decide what defaults should be: *e.g.* what *should* happen if neither `radius` nor `width` are specified?
- [x] Decide on a definition of "rectangle": `cos(dec)` or not.
- [x] Verify that documentation is clear.
- [x] Change log, etc.

Other notes:

* I rearranged some of the documentation sections, even in functions unrelated to this change, because they did not strictly follow numpydoc standards.
* But it turns out that breaks `async_to_sync`, so I'll undo that for the purposes of this PR.

### DIFF
--- a/astroquery/sdss/core.py
+++ b/astroquery/sdss/core.py
@@ -339,10 +339,10 @@ class SDSSClass(BaseQuery):
 
         if width is not None:
             if isinstance(width, Angle):
-                width = width.to_value(u.arcmin)
+                width = width.to_value(u.degree)
             else:
                 try:
-                    width = Angle(width).to_value(u.arcmin)
+                    width = Angle(width).to_value(u.degree)
                 except ValueError:
                     raise TypeError("width should be either Quantity or "
                                     "convertible to float.")
@@ -350,10 +350,10 @@ class SDSSClass(BaseQuery):
                 height = width
             else:
                 if isinstance(height, Angle):
-                    height = height.to_value(u.arcmin)
+                    height = height.to_value(u.degree)
                 else:
                     try:
-                        height = Angle(height).to_value(u.arcmin)
+                        height = Angle(height).to_value(u.degree)
                     except ValueError:
                         raise TypeError("height should be either Quantity or "
                                         "convertible to float.")
@@ -378,16 +378,8 @@ class SDSSClass(BaseQuery):
             for n, target in enumerate(coordinates):
                 # Query for a rectangle
                 target = commons.parse_coordinates(target).transform_to('fk5')
+                rectangles.append(self._rectangle_sql(target.ra.degree, target.dec.degree, width, height=height))
 
-                ra = target.ra.degree
-                dec = target.dec.degree
-                dra = Angle(width).to('degree').value / 2.0
-                ddec = Angle(height).to('degree').value / 2.0
-                rectangles.append('((p.ra BETWEEN {0:g} AND {1:g}) '
-                                  'AND (p.dec BETWEEN {2:g} AND {3:g}))'.format(ra - dra,
-                                                                                ra + dra,
-                                                                                dec - ddec,
-                                                                                dec + ddec))
             rect = ' OR '.join(rectangles)
             if 'WHERE' in sql_query:
                 sql_query += f' AND ({rect})'
--- a/astroquery/sdss/core.py
+++ b/astroquery/sdss/core.py
@@ -66,7 +66,7 @@ class SDSSClass(BaseQuery):
 
         This query returns the nearest `primary object`_.
 
-        Note that there is a server-side limit of 3 arcmin on `radius`.
+        Note that there is a server-side limit of 3 arcmin on ``radius``.
 
         .. _`primary object`: https://www.sdss.org/dr17/help/glossary/#surveyprimary
 
@@ -129,10 +129,10 @@ class SDSSClass(BaseQuery):
         Raises
         ------
         TypeError
-            If the `radius` keyword could not be parsed as an angle.
+            If the ``radius`` keyword could not be parsed as an angle.
         ValueError
-            If the `radius` exceeds 3 arcmin, or if the sizes of
-            `coordinates` and `obj_names` do not match.
+            If the ``radius`` exceeds 3 arcmin, or if the sizes of
+            ``coordinates`` and ``obj_names`` do not match.
 
         Returns
         -------
@@ -208,16 +208,16 @@ class SDSSClass(BaseQuery):
                            spectro=False, field_help=False, get_query_payload=False,
                            data_release=conf.default_release, cache=True):
         """
-        Used to query a region around given coordinates. Either `radius` or
+        Used to query a region around given coordinates. Either ``radius`` or
         `width` must be specified.
 
-        When called with keyword `radius`, a radial or "cone" search is
+        When called with keyword ``radius``, a radial or "cone" search is
         performed, centered on each of the given coordinates. In this mode, internally,
         this function is equivalent to the object cross-ID (`query_crossid`),
         with slightly different parameters.  Note that in this mode there is a server-side
-        limit of 3 arcmin on `radius`.
+        limit of 3 arcmin on ``radius``.
 
-        When called with keyword `width`, and optionally a different `height`,
+        When called with keyword ``width``, and optionally a different ``height``,
         a rectangular search is performed, centered on each of the given
         coordinates. In this mode, internally, this function is equivalent to
         a general SQL query (`query_sql`).
@@ -286,6 +286,15 @@ class SDSSClass(BaseQuery):
         cache : bool, optional
             If ``True`` use the request caching mechanism.
 
+        Raises
+        ------
+        TypeError
+            If the ``radius``, ``width`` or ``height`` keywords could not be parsed as an angle.
+        ValueError
+            If both ``radius`` and ``width`` are set (or neither),
+            or if the ``radius`` exceeds 3 arcmin,
+            or if the sizes of ``coordinates`` and ``obj_names`` do not match.
+
         Examples
         --------
         >>> from astroquery.sdss import SDSS
@@ -307,17 +316,6 @@ class SDSSClass(BaseQuery):
             The result of the query as a `~astropy.table.Table` object.
 
         """
-        # Move Raises section here, since async_to_sync does not appear to like it.
-        #
-        # Raises
-        # ------
-        # TypeError
-        #     If the `radius`, `width` or `height` keywords could not be parsed as an angle.
-        # ValueError
-        #     If both `radius` and `width are set (or neither),
-        #     or if the `radius` exceeds 3 arcmin,
-        #     or if the sizes of `coordinates` and `obj_names` do not match.
-
         # Allow field_help requests to pass without requiring a radius or width.
         if field_help and radius is None and width is None:
             radius = 2.0 * u.arcsec
@@ -1279,7 +1277,7 @@ class SDSSClass(BaseQuery):
 
     def _rectangle_sql(self, ra, dec, width, height=None, cosdec=False):
         """
-        Generate SQL for a rectangular query centered on `ra`, `dec`.
+        Generate SQL for a rectangular query centered on ``ra``, ``dec``.
 
         This assumes that RA is defined on the range ``[0, 360)``, and Dec on
         ``[-90, 90]``.
@@ -1293,9 +1291,11 @@ class SDSSClass(BaseQuery):
         width : float
             Width of rectangle in degrees.
         height : float, optional
-            Height of rectangle in degrees. If not specified, `width` is used.
+            Height of rectangle in degrees. If not specified, ``width`` is used.
         cosdec : bool, optional
             If ``True`` apply ``cos(dec)`` correction to the rectangle.
+            Otherwise, rectangles become increasingly triangle-like
+            near the poles.
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


weaverba137:
The `async_to_sync` functionality appears to assume that `Returns` is the last section of the docstring, but this violates the numpydoc standard. I'll eventually get around to restoring the `Returns`, but it would be more robust for `async_to_sync` to find and replace the `Returns` section, rather than just assuming it is at the end.

keflavich:
A fix to that incorrect assumption would be welcome!

weaverba137:
@keflavich, good, but I think I will keep that as a separate issue for now.

weaverba137:
@keflavich, I need some help on this one. I can no longer determine why `async_to_sync` is not producing compilable doc strings, even after I've restored the doc string sections as closely as possible to the state prior to this PR (as well as the last tag).

keflavich:
I see https://readthedocs.org/projects/astroquery/builds/19731041/ is the problem.

I'm not sure, but my first guess is that you've changed something like
```
"""
this
"""
```
to
```
"""this
"""
```
which has broken the `dedent` code?

I can dig a bit more later if that's not the soln

astroquery/sdss/core.py
```diff
@@ -193,16 +204,27 @@ def query_crossid_async(self, coordinates, *, radius=5. * u.arcsec, timeout=TIME
...
                            fields=None, photoobj_fields=None, specobj_fields=None, obj_names=None,
                            spectro=False, field_help=False, get_query_payload=False,
                            data_release=conf.default_release, cache=True):
-        """
-        Used to query a circular region (a "cone search") around given coordinates.
-
-        This function is equivalent to the object cross-ID (`query_crossid`),
-        with slightly different parameters. It returns all objects within the
-        search radius; this could potentially include duplicate observations
+        """Used to query a region around given coordinates. Either `radius` or
```
keflavich, line 212: try 
```suggestion
        """
        Used to query a region around given coordinates. Either `radius` or
```

weaverba137: That helped, but now it seems to be having problems with lines like:
```
Note that there is a server-side limit of 3 arcmin on `radius`.
```
which is the standard way to refer to arguments to a function.  Instead it is being interpreted with the `py:obj` role.

keflavich:
>
> That helped, but now it seems to be having problems with lines like:
>
> Note that there is a server-side limit of 3 arcmin on `radius`.
>
> which is the standard way to refer to arguments to a function. Instead it
> is being interpreted with the py:obj role.
>

try double instead of single backticks?

> --
Adam

astroquery/sdss/core.py
weaverba137: PS, [The Numpy Style Guide](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) uses single backticks to refer to parameters.

keflavich: We must have (I think unintentionally?) assigned single-backticks to be objects?  I don't think this was an intentional choice.  See the opening note on https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html.

weaverba137: Right, I was seeing other evidence of that. Probably not worth trying to solve in this PR, so I'll just use double-backticks for now.

keflavich: If you can figure out where we set that, let me know - this is something that has always confused and frustrated me too.

weaverba137: https://github.com/astropy/sphinx-astropy/blob/main/sphinx_astropy/conf/v1.py#L80

weaverba137: Documentation is compiling successfully now, & I created astropy/sphinx-astropy#58 to ask about why `default_role` is set.

weaverba137:
@keflavich, *et al*., I still need to do some updates to the remote data tests, but I think it is time to move this out of draft status and start thinking about the unchecked boxes above:

* What should be the default for `query_region()` if neither `radius` nor `width` are specified? Currently it raises an exception.
* What is the definition of "rectangle". Currently, I implemented an improved version of the original rectangle from the previous tag (now handles RA wrap-around!). It does not currently implement `cos(dec)` corrections.
* Is the documentation sufficiently clear?

astroquery/sdss/core.py
```diff
@@ -277,16 +317,83 @@ def query_region_async(self, coordinates, *, radius=2. * u.arcsec, timeout=TIMEO
...
-                                                          get_query_payload=True,
-                                                          data_release=data_release)
+        # Allow field_help requests to pass without requiring a radius or width.
+        if field_help and radius is None and width is None:
+            radius = 2.0 * u.arcsec
+
+        if radius is None and width is None:
+            raise ValueError("Either radius or width must be selected!")
+        if radius is not None and width is not None:
+            raise ValueError("Either radius or width must be selected!")
```
keflavich, line 327:
```suggestion
            raise ValueError("Either radius or width must be specified, not both.")
```

astroquery/sdss/core.py
```diff
@@ -277,16 +317,83 @@ def query_region_async(self, coordinates, *, radius=2. * u.arcsec, timeout=TIMEO
...
-                                                          region=True,
-                                                          field_help=field_help,
-                                                          get_query_payload=True,
-                                                          data_release=data_release)
+        # Allow field_help requests to pass without requiring a radius or width.
+        if field_help and radius is None and width is None:
+            radius = 2.0 * u.arcsec
+
+        if radius is None and width is None:
+            raise ValueError("Either radius or width must be selected!")
```
keflavich, line 325:
```suggestion
            raise ValueError("Either radius or width must be specified.")
```

astroquery/sdss/core.py
```diff
@@ -277,16 +317,83 @@ def query_region_async(self, coordinates, *, radius=2. * u.arcsec, timeout=TIMEO
...
+
+        if width is not None:
+            if isinstance(width, Angle):
+                width = width.to_value(u.degree)
+            else:
+                try:
+                    width = Angle(width).to_value(u.degree)
+                except ValueError:
+                    raise TypeError("width should be either Quantity or "
+                                    "convertible to float.")
```
keflavich, line 349: I think this achieves the same thing more tersely:
```suggestion
            width = u.Quantity(width, u.degree).value
```

astroquery/sdss/core.py
```diff
@@ -277,16 +317,83 @@ def query_region_async(self, coordinates, *, radius=2. * u.arcsec, timeout=TIMEO
...
+                height = width
+            else:
+                if isinstance(height, Angle):
+                    height = height.to_value(u.degree)
+                else:
+                    try:
+                        height = Angle(height).to_value(u.degree)
+                    except ValueError:
+                        raise TypeError("height should be either Quantity or "
+                                        "convertible to float.")
```
keflavich, line 360:
```suggestion
                height = u.Quantity(height, u.degree).value
```

astroquery/sdss/core.py
```diff
@@ -277,16 +317,83 @@ def query_region_async(self, coordinates, *, radius=2. * u.arcsec, timeout=TIMEO
...
+                sql_query = sql_query.replace('ON p.objID = x.objID ', '').replace(' ORDER BY x.up_id', '')
+            else:
+                sql_query = sql_query.replace(' ON p.objID = x.objID ORDER BY x.up_id', '')
+
+            if (not isinstance(coordinates, list) and not isinstance(coordinates, Column)
+                    and not (isinstance(coordinates, commons.CoordClasses) and not coordinates.isscalar)):
+                coordinates = [coordinates]
+
+            rectangles = list()
+            for n, target in enumerate(coordinates):
```
keflavich, line 381:
```suggestion
            for target in coordinates:
```

keflavich (changes requested):
> What should be the default for query_region() if neither radius nor width are specified? Currently it raises an exception.

This is the right behavior

> What is the definition of "rectangle". Currently, I implemented an improved version of the original rectangle from the previous tag (now handles RA wrap-around!). It does not currently implement cos(dec) corrections.

The current implementation is fine, but ...

> Is the documentation sufficiently clear?

Yes, but include a note that the rectangle coordinates represent (ra +/- width, dec +/- height) as opposed to (ra +/- width/cos(dec), dec +/- height)


Otherwise, looks good.  I put in some minor code readability suggestions - if they do not achieve the desired behavior, though, feel free to reject.

astroquery/sdss/core.py
weaverba137: Done.

astroquery/sdss/core.py
weaverba137: Done.

astroquery/sdss/core.py
weaverba137: Done.

astroquery/sdss/core.py
weaverba137: OK, that's a nice trick. Done. Although for the record, two possible types of exception are raised: `width = u.Quantity('foo', u.degree).value` raises `TypeError`, while `width = u.Quantity('2 things', u.degree).value` raises `ValueError`.

astroquery/sdss/core.py
weaverba137: Done, but see comment above.

weaverba137:
@keflavich, I've implemented your suggestions and also run remote tests to verify those are working.  The final step is updating the documentation.

weaverba137:
@keflavich, *et al.*, I've added further details to the documentation and the change log.  For some reason Codecov is saying the total coverage decreased, even though I increased the coverage of sdss/core.py to 100%.  The largest decrease appears to be in cds/core.py, but I don't know why that is.

docs/sdss/sdss.rst
```diff
@@ -17,25 +17,57 @@ images, are also included in DR17.  Users may select alternate DR's.
...
 
-This example shows how to perform an object cross-ID with SDSS. We'll start
-with the position of a source found in another survey, and search within a 5
-arcsecond radius for optical counterparts in SDSS. Note use of the keyword
-argument spectro, which requires matches to have spectroscopy, not just
-photometry:
+This example shows how to perform an individual object cross-ID with SDSS.
+We'll start with the position of a source found in another survey, and search
+within a 5 arcsecond radius (a "cone search") for optical counterparts in SDSS.
+Note use of the keyword argument spectro, which requires matches to have
```
keflavich, line 23:
```suggestion
Note use of the keyword argument ``spectro``, which requires matches to have
```

docs/sdss/sdss.rst
```diff
@@ -17,25 +17,57 @@ images, are also included in DR17.  Users may select alternate DR's.
...
+rectangular region centered on a coordinate or each coordinate in a list.
+This mode is invoked with the ``width`` keyword, which is the width in
+right ascension. Optionally, the ``height`` keyword can be used to specify
+a different range of declination.  With these parameters,
+`~astroquery.sdss.SDSSClass.query_region` constructs a rectangle in RA, dec
+that does *not* correct for the geometry at high declination, also known as
+the :math:`\cos \delta` correction.  At high declination, these rectangles
+would appear much more like trapezoids. However, this is the more intuitive
+interpretation of "this range of RA, that range of Dec" that many people use.
+Finally though, the constructed rectangles *do* account for RA wrap-around,
```
keflavich, line 64:
```suggestion
Finally, though, the constructed rectangles *do* account for RA wrap-around,
```

keflavich (approved):
I like the clarification text about the shape of the selection.  I have two very minor suggestions but I think this is good to merge.  I'm not really worried about the coverage decrease because it can't be related, but if anyone else has ideas, voice 'em

docs/sdss/sdss.rst
weaverba137: Good suggestion, done.

docs/sdss/sdss.rst
weaverba137: Good suggestion, done.

weaverba137:
OK, I've committed the last two suggestions.

weaverba137:
Reminder: this is waiting for final review.

keflavich:
@bsipocz I'm good with this, but I'm wary of hitting 'merge' without your eyes on it.

bsipocz (approved):
All looks good, thanks!

PR #2668, opened 2023-02-13
Expanding `astroquery.mast.Tesscut` to support TICA

jaymedina:
Following the expanded functionality of [TESSCut](https://mast.stsci.edu/tesscut/) for TICA high-level science products, this PR will serve to update `TesscutClass` to reflect this new functionality for users who prefer making TESSCut API requests through `astroquery.mast`. 

**Timeline: We would like to get this merged in by COB February 23, if possible.**

#### Functionality updates:

1. New product call for `get_sectors`:
image

2. New product call for `download_cutouts`: 
image

3. New product call for `get_cutouts`:
image

#### Moving targets cutouts cannot be made from TICA products, so I've put checks in all 3 functions to catch and correct when TICA is fed while moving targets is `True`:
image

### DIFF
--- a/astroquery/mast/cutouts.py
+++ b/astroquery/mast/cutouts.py
@@ -106,7 +106,8 @@ class TesscutClass(MastQueryWithLogin):
                     }
         self._service_api_connection.set_service_params(services, "tesscut")
 
-    def get_sectors(self, *, coordinates=None, radius=0*u.deg, objectname=None, moving_target=False, mt_type=None):
+    def get_sectors(self, *, coordinates=None, radius=0*u.deg, product='SPOC', objectname=None,
+                    moving_target=False, mt_type=None):
         """
         Get a list of the TESS data sectors whose footprints intersect
         with the given search area.
@@ -126,6 +127,14 @@ class TesscutClass(MastQueryWithLogin):
             `astropy.units` may also be used.
 
             NOTE: If moving_target is supplied, this argument is ignored.
+        product : str
+            Default is 'SPOC'.
+            The product whose sectors will be returned. Options are: 'SPOC' or 'TICA', for the Science Processing
+            Operations Center (SPOC) products, and the TESS Image CAlibration (TICA) high-level science products,
+            respectively. TICA products will usually be available for the latest sectors sooner than their SPOC
+            counterparts, but are not available for sectors 1-26.
+
+            NOTE: TICA is currently not available for moving targets.
         objectname : str, optional
             The target around which to search, by name (objectname="M104")
             or TIC ID (objectname="TIC 141914082"). If moving_target is True, input must be the name or ID
@@ -152,6 +161,11 @@ class TesscutClass(MastQueryWithLogin):
 
         if moving_target:
 
+            # The Moving Targets service is currently only available for SPOC
+            if product.upper() != "SPOC":
+                product = "SPOC"
+                warnings.warn("Only SPOC is available for moving targets queries. Defaulting to SPOC.", InputWarning)
+
             # Check that objectname has been passed in and coordinates
             # is not
             if coordinates:
@@ -163,7 +177,7 @@ class TesscutClass(MastQueryWithLogin):
                                         "`JPL ephemerides service <https://ssd.jpl.nasa.gov/horizons.cgi>`__) "
                                         "of a moving target such as an asteroid or comet.")
 
-            params = {"obj_id": objectname}
+            params = {"product": product.upper(), "obj_id": objectname}
 
             # Add optional parameter is present
             if mt_type:
@@ -179,9 +193,14 @@ class TesscutClass(MastQueryWithLogin):
             # If radius is just a number we assume degrees
             radius = Angle(radius, u.deg)
 
+            # Making sure input product is either SPOC or TICA
+            if product.upper() not in ['TICA', 'SPOC']:
+                raise InvalidQueryError("Input product must either be SPOC or TICA.")
+
             params = {"ra": coordinates.ra.deg,
                       "dec": coordinates.dec.deg,
-                      "radius": radius.deg}
+                      "radius": radius.deg,
+                      "product": product.upper()}
 
             response = self._service_api_connection.service_request_async("sector", params)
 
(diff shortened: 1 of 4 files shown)


astroquery/mast/tests/test_mast_remote.py
```diff
@@ -759,42 +759,36 @@ def test_catalogs_download_hsc_spectra(self, tmpdir):
...
-
-        with pytest.warns(NoResultsWarning):
-            coord = SkyCoord(90, -66.5, unit="deg")
-            sector_table = mast.Tesscut.get_sectors(coordinates=coord,
-                                                    radius=0)
-            assert isinstance(sector_table, Table)
-            assert len(sector_table) == 0
-
-        coord = SkyCoord(90, -66.5, unit="deg")
-        with pytest.warns(NoResultsWarning):
```
jaymedina, line 783: This test was the same as the one from above (L775-780) except with the assertions un-indented, so it looked like a duplicate.

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -759,42 +759,36 @@ def test_catalogs_download_hsc_spectra(self, tmpdir):
...
+
+        name = "tess" if product == "spoc" else product
+        assert f"{name}-s00" in sector_table['sectorName'][0]
+
         assert sector_table['sector'][0] > 0
         assert sector_table['camera'][0] > 0
         assert sector_table['ccd'][0] > 0
 
-        # Moving target functionality testing
+    def test_tesscut_get_sectors_mt(self):
```
jaymedina, line 788: I decided to move all the moving targets testing to a separate function. Moving targets is not supported by TICA, so it didn't make sense to have these calls under the `test_tesscut_get_sectors` function, since that now iterates between `product=` TICA and SPOC.

astroquery/mast/cutouts.py
```diff
@@ -152,6 +161,11 @@ def get_sectors(self, *, coordinates=None, radius=0*u.deg, objectname=None, movi
 
         if moving_target:
 
+            # The Moving Targets service is currently only available for SPOC
+            if product.upper() != "SPOC":
+                product = "SPOC"
+                warnings.warn("Only SPOC is available for moving targets queries. Defaulting to SPOC.", InputWarning)
```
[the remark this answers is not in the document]
keflavich: I mean, since it should be unlikely for a user to end up here accidentally, we are better off telling them they've gone wrong and won't get the expected result, rather than changing their input options and returning a (probably unexpected) result

astroquery/mast/cutouts.py
```diff
@@ -204,8 +223,8 @@ def get_sectors(self, *, coordinates=None, radius=0*u.deg, objectname=None, movi
             warnings.warn("Coordinates are not in any TESS sector.", NoResultsWarning)
         return Table(sector_dict)
 
-    def download_cutouts(self, *, coordinates=None, size=5, sector=None, path=".", inflate=True,
-                         objectname=None, moving_target=False, mt_type=None, verbose=False):
+    def download_cutouts(self, *, coordinates=None, size=5, sector=None, product='SPOC', path=".",
+                         inflate=True, objectname=None, moving_target=False, mt_type=None, verbose=False):
```
keflavich, line 227: `inflate` is not described in the docstring

astroquery/mast/cutouts.py
```diff
@@ -379,6 +425,13 @@ def get_cutouts(self, *, coordinates=None, size=5, sector=None,
 
         if moving_target:
 
+            # The Moving Targets service is currently only available for SPOC
+            if product.upper() != "SPOC":
+                product = "SPOC"
+                warnings.warn("Only SPOC is available for moving targets queries. Defaulting to SPOC.", InputWarning)
```
keflavich, line 431: ditto above, and since this code (and the docstring) is repeated, is there any way these could be refactored such that this checking only needs to be done once in the code?  That may be beyond the scope of this PR, but copy-pasted code & docs rings alarm bells for me that we could write a single function to handle this functionality instead of adding it to several functions.

docs/mast/mast.rst
```diff
@@ -927,6 +938,22 @@ and returns a target pixel file, with format described
      1  PIXELS        1 BinTableHDU    280   1196R x 12C   [D, E, J, 25J, 25E, 25E, 25E, 25E, J, E, E, 38A]
      2  APERTURE      1 ImageHDU        81   (5, 5)   int32
 
+For users with time-sensitive targets who would like cutouts from the latest observations, 
+we recommend requesting for the TICA product. Using the same target from the example above, 
+let's request for the TICA cutouts:
```
keflavich, line 943:
```suggestion
this example shows a request for the TICA cutouts:
```

docs/mast/mast.rst
```diff
@@ -927,6 +938,22 @@ and returns a target pixel file, with format described
...
+
+.. doctest-remote-data::
+
+   >>> from astroquery.mast import Tesscut
+   >>> from astropy.coordinates import SkyCoord
+   ...
+   >>> cutout_coord = SkyCoord(107.18696, -70.50919, unit="deg")
+   >>> hdulist = Tesscut.get_cutouts(coordinates=cutout_coord, size=5, product='tica')
```
keflavich, line 951: I'd suggest noting in the narrative docs above that `size=5` means 5 pixels, and perhaps even mention how big a pixel (typically?) is

keflavich: If this is specified above (I haven't read the full docs tonight), perhaps just remove `size=5`, since it is the default value.

keflavich (changes requested):
I have a few minor suggestions, mostly about docs and looking for places to simplify the code.  I don't see any show-stoppers though.

astroquery/mast/cutouts.py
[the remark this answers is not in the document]
jaymedina: Fair point, yes I can make this an error instead

astroquery/mast/cutouts.py
jaymedina: There are a few parts in `cutouts.py` with duplicate code so I'm in favor of leaving it how it is for now and opening up a new PR for this refactoring and addressing all of the duplicates that way

astroquery/mast/cutouts.py
jaymedina: Inflate is in the docstring, starting at L235

docs/mast/mast.rst
jaymedina: The unit is specified a couple examples above, so I've removed size=5 and added that the default is 5 pixels in the little blurb where it's specified.

astroquery/mast/cutouts.py
keflavich: Ahhh, I see, it was _already_ in the docstring but wasn't in the parameters.  Got it!

jaymedina:
I've added more unit tests to cover `cutouts.py` but somehow the diff coverage didn't budge and is still at -0.01%. Will this keep my PR from getting merged or is this good to go?

ceb8 (approved):
Looks good to me!

@keflavich  Can you verify that your requested changes have been made?

keflavich (approved):

ceb8:
Thanks @jaymedina!

PR #2675, opened 2023-02-27
Updated pip command for repository install

jaymedina:
Updated the pip command for installing the bleeding edge version of `astroquery`. The command in place was returning an error due to `pip` not finding `astropy_helpers` even after manually installing it.


ceb8 (approved):
Looks good to me.

bsipocz (approved):
While we're at it, I'm pushing another commit to update the source install instructions, too.

Issue #2447, opened 2022-06-19
TESScut timeouts for larger cuts


TeHanHunter:
I have been using TESScut to cut 90*90 FFIs and it worked fine until recently. Now it is stuck at downloading and then gives a timeout error.

```
from astroquery.mast import Tesscut
ra = '60' 
dec = '60'
coord = SkyCoord(ra, dec, unit="deg")
sector_table = Tesscut.get_sectors(coordinates=coord)
hdulist = Tesscut.get_cutouts(coordinates=coord, size=90)
```

Any help is appreciated!

bsipocz:
cc @jaymedina

TeHanHunter:
It seems that my pc does not work. The same script works on a cluster. I will find out why.

StuartLittlefair:
@TeHanHunter did you ever find out why? I am finding the same issue; I can download the cutouts fine if I use the astrocut API directly using `requests`...

TeHanHunter:
Hi, I think the problem persists for me. Especially when the number of
sectors is large. I was able to download sectors individually to avoid the
issue.

jaymedina:
Hi all, what is the error you're receiving? Is it a `503 Service Temporarily Unavailable` error?

StuartLittlefair:
No, it's a timeout error "timeout of 600 exceeded".

The thing is, using the API with requests takes only about 1 minute...

jaymedina:
OK I'll look into this, thanks!

jaymedina:
Any `astroquery.mast` requests that hit the upper limit of 600 seconds get hit with a timeout error. You can see the upper limit by calling the `ServicesAPI()` class under `services.py` like so: 

```
from astroquery.mast import services

services.ServiceAPI().TIMEOUT

>>> 600
```

You can technically override this by reassigning the `TIMEOUT` parameter to whatever you want, like so:

```
Tesscut._service_api_connection.TIMEOUT = 5000
```

After overriding this parameter I was able to get the 90x90 cutouts HDU list for a single sector for coordinates (0, 89) after a runtime of ~689 seconds:

```
Tesscut._service_api_connection.TIMEOUT = 5000

coords = '0 89'
start = time.time()
cutouts = Tesscut.get_cutouts(coordinates=coords, size=90, sector=19)

print(f'Time lapsed: {str(time.time()-start)}')

>>> Time lapsed: 689.3902568817139
```

[...]

PR #2693, opened 2023-03-29
Cutouts limit

jaymedina:
Addresses some points from the conversation here https://github.com/astropy/astroquery/issues/2447 . 

Changes: 
- [ ] Add a warning message for cutout sizes where at least 1 dimension goes beyond the recommended size limit of 30 pixels. 
- [ ] Add a `timeout` argument to `get_cutouts` that allows the user to modify the request processing time limit so that at timeout error doesn't occur at 600s.


astroquery/mast/cutouts.py
```diff
@@ -70,21 +71,37 @@ def _parse_cutout_size(size):
                       InputWarning)
 
     # Getting x and y out of the size
+
     if np.isscalar(size[0]):
         x = size[1]
         y = size[0]
         units = "px"
+        limit_reached = True if ((x >= 30) or (y >= 30)) else False
```
bsipocz, line 79: Does this 30 have to be hard-wired? (is this a server side limit?)

astroquery/mast/cutouts.py
```diff
@@ -70,21 +71,37 @@ def _parse_cutout_size(size):
...
+        # to degrees.
+        upper_limit = (30 * 21*u.arcsec).to(u.deg)
+        limit_reached = True if ((x >= upper_limit) or (y >= upper_limit)) else False
     else:
         raise InvalidQueryError("Cutout size must be in pixels or angular quantity.")
 
+    if limit_reached:
+        warnings.warn("You have selected a large cutout size that may result in a timeout error. We suggest limiting"\
+                      " the size of your requested cutout, or changing the request timeout limit from its" \
+                      " default 600 seconds to something higher, using the timeout argument.", InputWarning)
```
bsipocz, line 103: no need for the line breaks as it's all in a pair of parens
```suggestion
        warnings.warn("You have selected a large cutout size that may result in a timeout error. We suggest limiting"
                      " the size of your requested cutout, or changing the request timeout limit from its" 
                      " default 600 seconds to something higher, using the timeout argument.", InputWarning)
```

bsipocz (commented):
I wonder whether the `InputWarning` is the correct one here, or would rather need a more astroquery/astropy warning type? @keflavich, do you have a preference?

Also, this needs a changelog.

astroquery/mast/cutouts.py
bsipocz: Would you mind adding this description to the narrative docs, as a `Note`, too?

jaymedina: Will do

eerovaher, line 79: The following is equivalent:
```suggestion
        limit_reached = x >= 30 or y >= 30
```
But based on the pull request description the following should be used:
```suggestion
        limit_reached = x > 30 or y > 30
```

astroquery/mast/cutouts.py
```diff
@@ -70,21 +71,37 @@ def _parse_cutout_size(size):
...
+
     elif size[0].unit.physical_type == 'angle':
         x = size[1].to(u.deg).value
         y = size[0].to(u.deg).value
         units = "d"
+
+        # Based on the literature, TESS resolution is approx. 21 arcseconds per pixel.
+        # We will convert the recommended upper limit for a dimension from pixels
+        # to degrees.
+        upper_limit = (30 * 21*u.arcsec).to(u.deg)
```
eerovaher, line 95: Why is the conversion to degrees needed?
```suggestion
        upper_limit = 30 * 21*u.arcsec
```

astroquery/mast/cutouts.py
```diff
@@ -70,21 +71,37 @@ def _parse_cutout_size(size):
...
+        limit_reached = True if ((x >= upper_limit) or (y >= upper_limit)) else False
     else:
         raise InvalidQueryError("Cutout size must be in pixels or angular quantity.")
 
+    if limit_reached:
+        warnings.warn("You have selected a large cutout size that may result in a timeout error. We suggest limiting"
+                      " the size of your requested cutout, or changing the request timeout limit from its" 
+                      " default 600 seconds to something higher, using the timeout argument.", InputWarning)
```
eerovaher, line 103: The user will receive this warning even if they have followed the advice and did use the `timeout` argument.

astroquery/mast/cutouts.py
```diff
@@ -414,6 +431,10 @@ def get_cutouts(self, *, coordinates=None, size=5, product='SPOC', sector=None,
         response : A list of `~astropy.io.fits.HDUList` objects.
         """
 
+        if timeout:
+            log.info(f'Request timeout upper limit is being changed to {timeout} seconds.', InputWarning)
+            self._service_api_connection.TIMEOUT = timeout
```
eerovaher, line 436: That specifying a function parameter permanently changes the value of a constant is very unexpected. It would be better to make the timeout configurable using the `astropy` configuration system, see https://docs.astropy.org/en/stable/config/index.html#usage-tips

astroquery/mast/cutouts.py
jaymedina: Because the code needs to represent the upper limit in terms of degrees for the cases where users input their cutout size in degrees rather than pixels.

jaymedina:
I believe I've addressed everything now so this is ready for a final review @bsipocz @eerovaher . Once I get this approved, I'll do a squash so it's ready for merge.

astroquery/mast/cutouts.py
```diff
@@ -589,12 +633,24 @@ def download_cutouts(self, coordinates, *, size=5, survey=None, cutout_format="f
...
         -------
         response : `~astropy.table.Table`
             Cutout file(s) for given coordinates
         """
+
+        # Modify TIMEOUT attribute if necessary (usually this is modified for large requests)
+        if timeout_add:
+            self._service_api_connection.TIMEOUT = self._service_api_connection.TIMEOUT + timeout_add
+            log.info(f"Request timeout upper limit is being changed to {self._service_api_connection.TIMEOUT}"
+                     " seconds.")
```
eerovaher, line 652: I don't think this is a good approach because it permanently modifies the timeout, which is very unexpected. There already exists a configuration parameter for the timeout: https://github.com/astropy/astroquery/blob/770949eda7883fef82c3be714b1f618cdce7d05c/astroquery/mast/__init__.py#L23-L25
It should be simple to use a different value for one function call without modifying it permanently, see https://docs.astropy.org/en/stable/config/index.html#usage-tips

astroquery/mast/cutouts.py
```diff
@@ -60,31 +72,50 @@ def _parse_cutout_size(size):
...
+
         if len(size) == 1:
             size = np.repeat(size, 2)
 
+        # Based on the literature, TESS resolution is approx. 21 arcseconds per pixel.
+        # We will convert the recommended upper limit for a dimension from pixels
+        # to the unit being passed.
+        unit = size[0].unit
+        upper_limit = (30 * 21*u.arcsec).to(unit).value
+        limit_reached = size[0].value > upper_limit or size[1].value > upper_limit
```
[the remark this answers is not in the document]
jaymedina: The input `size` isn't always going to be in arcseconds, though. That's why `upper_limit` was created, so that the 30 pixel upper limit gets computed in whatever unit the user fed their cutout size with.

astroquery/mast/cutouts.py
```diff
@@ -408,14 +441,25 @@ def get_cutouts(self, *, coordinates=None, size=5, product='SPOC', sector=None,
             first majorbody is tried and then smallbody if a matching majorbody is not found.
 
             NOTE: If moving_target is supplied, this argument is ignored.
+        timeout_add : int or float, optional
+            The amount (in seconds) by which the request processing time upper limit will be changed.
```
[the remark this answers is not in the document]
jaymedina: Yes, I've changed the `_change_timeout` method now to `_get_default_timeout` so that the `TIMEOUT` attribute gets written back to the default if `timeout` is called.

astroquery/mast/cutouts.py
```diff
@@ -60,31 +72,50 @@ def _parse_cutout_size(size):
...
         size = np.atleast_1d(size)
+
         if len(size) == 1:
             size = np.repeat(size, 2)
 
+        # Based on the literature, TESS resolution is approx. 21 arcseconds per pixel.
+        # We will convert the recommended upper limit for a dimension from pixels
+        # to the unit being passed.
+        unit = size[0].unit
+        upper_limit = (30 * 21*u.arcsec).to(unit).value
```
bsipocz, line 87: now that there is the `mission` kwarg, too, maybe check for it here, so you only use the TESS hard-wired 21 arcsec pixel scale for TESS?
Also, I think this would not work for pixel units, even though the docstring above says pixel or angular units are both acceptable. ==> overall it would be worth to do parametrized testing using all three options.

astroquery/mast/cutouts.py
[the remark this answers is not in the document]
bsipocz: being faster here also shouldn't be such a huge deal, hopefully this piece of code only runs once per query, and then after all we're querying a webservice, which would certainly dominate the overall speed.

astroquery/mast/cutouts.py
```diff
@@ -106,8 +137,14 @@ def __init__(self):
                     }
         self._service_api_connection.set_service_params(services, "tesscut")
 
+    def _get_default_timeout(self):
+        """ Gets the default request timeout limit. """
+
+        return self._service_api_connection.TIMEOUT
```
bsipocz, line 143: why is this method necessary? (I see in the history that there has been a bit of a back and forth, so maybe it's not needed any more?

astroquery/mast/cutouts.py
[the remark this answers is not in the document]
eerovaher: I agree performance isn't critical here. The suggested version is preferable mainly because it produces the same output with much less code. Being faster is just a nice bonus.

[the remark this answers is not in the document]
bsipocz: The code is buggy as is, so I suggest first ensuring all acceptable input units are included in the test suite. Then I suppose there we'll need to have a both the upper limit definitions, but could also have your `any()` simplification.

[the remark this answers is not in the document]
eerovaher:
```suggestion
        with u.set_enabled_equivalencies(u.pixel_scale(21 * u.arcsec / u.pixel)):
            limit_reached = (size > 30 * u.pixel).any()
```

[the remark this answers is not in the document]
bsipocz: Nice, I totally forgot about this equivalency.

bsipocz (changes requested):
I see test failures, so the tests indeed need to be revisited:

```
FAILED astroquery/mast/tests/test_mast_remote.py::TestMast::test_zcut_download_cutouts - UnboundLocalError: local variable 'limit_reached' referenced before assignment
FAILED astroquery/mast/tests/test_mast_remote.py::TestMast::test_zcut_get_cutouts - UnboundLocalError: local variable 'limit_reached' referenced before assignment
FAILED astroquery/mast/tests/test_mast_remote.py::TestMast::test_hapcut_download_cutouts - UnboundLocalError: local variable 'limit_reached' referenced before assignment
FAILED astroquery/mast/tests/test_mast_remote.py::TestMast::test_hapcut_get_cutouts - UnboundLocalError: local variable 'limit_reached' referenced before assignment
```

astroquery/mast/cutouts.py
jaymedina: I included this for the case where a user might switch between `get_cutouts` and `download_cutouts` but does not anticipate that the `timeout` setting stayed the same. Having the `timeout` arg in each individual method implies that it would only change `timeout` for a call with that method and not any of the others, so I figured it would make sense to keep the changes contained by resetting that arg each time.

astroquery/mast/cutouts.py
jaymedina: The `limit_reached` arg will be ignored if `mission` is anything but TESS, and the `upper_limit` check for pixel units is a few lines further up in this method.

jaymedina:
Just squashed and rebased, let me know if there's anything else that needs to be addressed before merge.

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -964,6 +964,11 @@ def test_tesscut_get_cutouts(self, product):
         assert len(cutout_hdus_list) >= 1
         assert isinstance(cutout_hdus_list[0], fits.HDUList)
 
+        # Check that an INFO message is returned when timeout is adjusted
+        mast.Tesscut.get_cutouts(product=product, coordinates=coord, size=5, timeout=1000)
```
bsipocz, line 968: Maybe assert that the `TIMEOUT` attribute has been properly set back to the default one.

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -1016,6 +1021,15 @@ def test_tesscut_get_cutouts_mt(self):
                                      moving_target=True)
             assert error_tica_mt in str(error_msg.value)
 
+    @pytest.mark.xfail(raises=InputWarning)
```
bsipocz, line 1024: This is not the correct approach, xfails are more for cases of known buggy behaviour, or expected server side issues, etc.
Instead, please use pytest.warns context manager, there are plenty of usage examples for it in astroquery.

bsipocz (changes requested):
I'm afraid some more iterations for the tests are needed to cover the missing (and not working) input case, as well as to fix checking for the raised warning. But overall the PR is pretty close to being ready.

astroquery/mast/tests/test_mast_remote.py
jaymedina: The reason I had to do it this way is because running with pytest.warns will allow the query to run fully, which will take several minutes and also end up failing the test with the timeout error. This is for the queries that exceed the recommended timeout limit of 600 seconds because the cutout size is too large and the `timeout` arg hasn't been changed.

astroquery/mast/cutouts.py
```diff
@@ -56,35 +68,58 @@ def _parse_cutout_size(size):
         either pixels or degrees.
     """
 
+    # This local variable will change to True if input cutout size exceeds recommended limits for TESS
+    limit_reached = False
+
     # Making size into an array [ny, nx]
     if np.isscalar(size):
```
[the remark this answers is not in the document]
jaymedina: Just made some new changes that tackle 2d inputs both unitless (which are assumed to be in pixels, following the tradition of the code) and with units: 

image

image

docs/mast/mast.rst
```diff
@@ -1042,6 +1042,40 @@ and because the TICA products are not available for sectors 1-26, we request cut
...
+
+.. doctest-remote-data::
+
+   >>> import astropy.units as u
+   >>> from astroquery.mast import Tesscut
+   >>> from astropy.coordinates import SkyCoord
+   ...
+   >>> cutout_coord = SkyCoord(107.18696, -70.50919, unit="deg")
+   >>> hdulist = Tesscut.get_cutouts(coordinates=cutout_coord, size=0.2*u.deg, timeout=2600)
+   INFO: Request timeout upper limit is being changed to 2600 seconds. [astroquery.mast.cutouts]
```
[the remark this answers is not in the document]
jaymedina: I'll remove this example but I'd like to keep the example that returns the `LargeQueryWarning` when `timeout` hasnt been changed and cutout size is too large, just so the user has an idea of what to expect. I'll be skipping that example in the doctest so it doesn't extend the runtime.

astroquery/mast/cutouts.py
```diff
@@ -56,35 +68,67 @@ def _parse_cutout_size(size):
...
+    limit_reached = False
+
+    # Checking 2d size inputs for the recommended cutout size
+    if isinstance(size, list) & (mission == 'TESS'):
+        if len(size) == 2:
+            if np.isscalar(size[0]):
+                size = [size[0] * u.pixel, size[1] * u.pixel]
+
+            with u.set_enabled_equivalencies(u.pixel_scale(21 * u.arcsec / u.pixel)):
+                limit_reached = (size * size[0].unit > 30 * u.pixel).any()
```
jaymedina, line 81: For `set_enabled_equivalencies`, the entire list needs to be assigned a unit rather than individual elements, in order for a conversion to take place. Otherwise this line results in an error.

image

astroquery/mast/cutouts.py
jaymedina: The code has moved around since this last comment so I'm not sure what you're referring to, but I made a call to `_get_default_timeout` several lines below L414.

astroquery/mast/cutouts.py
```diff
@@ -56,35 +68,67 @@ def _parse_cutout_size(size):
         either pixels or degrees.
     """
 
+    # This local variable will change to True if input cutout size exceeds recommended limits for TESS
+    limit_reached = False
+
+    # Checking 2d size inputs for the recommended cutout size
+    if not isinstance(size, (int, float, u.Quantity)) & (mission == 'TESS'):
```
jaymedina, line 75: My goal here was to catch for any array-like data structures, so lists tuples or np arrays. Unfortunately, I can't call for them explicitly because `u.Quantity` objects are arrays themselves, so I went with _excluding_ anything that is a single int/float/u.Quantity value, which will implicitly pick up data structures. 

image

jaymedina: Needed another parenthesis around the `not isinstance(size, (int, float, u.Quantity))` - this has been patched.

astroquery/mast/cutouts.py
```diff
@@ -56,35 +68,67 @@ def _parse_cutout_size(size):
...
     """
 
+    # This local variable will change to True if input cutout size exceeds recommended limits for TESS
+    limit_reached = False
+
+    # Checking 2d size inputs for the recommended cutout size
+    if isinstance(size, list) & (mission == 'TESS'):
+        if len(size) == 2:
+            if np.isscalar(size[0]):
+                size = [size[0] * u.pixel, size[1] * u.pixel]
```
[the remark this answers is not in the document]
bsipocz: yes, I would think you don't need the `size[0].unit` in your example and then it would work (you already have that in the screenshot below)

astroquery/mast/cutouts.py
bsipocz: But it doesn't provide the default timeout, but gets the attribute at all time. If you need the value from the attribute, do call it directly in the code (this is why I don't see the reason for this method), or if you really need the default, do reach out to it directly from the config.

Maybe it's just a naming issue? Though I still think the method is not necessary as it basically just calls one line that you could call directly when needed.

bsipocz (commented):
two more points to come back to for a tiny bit of a cleanup, but otherwise this will be good to go.

jaymedina:
@bsipocz I've removed the `_get_default_timeout` method and am now calling the default timeout directly from `Conf()`.

I'm not sure I understood this comment here, are you suggesting to remove the `not isinstance(u.Quantity)` conditional in the lines above your comment? 
> here you could put the list into a quantity and then do the check.

> `size2 = u.Quantity(size2)`

> So, maybe distinguish in the isinstance conditional between Quantity and not yet quantity inputs?

bsipocz:
> I'm not sure I understood this comment here, are you suggesting to remove the not isinstance(u.Quantity) conditional in > the lines above your comment?

>>   here you could put the list into a quantity and then do the check.
>> size2 = u.Quantity(size2)
>> So, maybe distinguish in the isinstance conditional between Quantity and not yet quantity inputs?

I'll do a suggestion commit when next to a computer (travelling atm). Basically I meant to suggest chopping up the conditional to distinguish between vector quantities vs a list of scalar quantities, so the case you show in the screenshot would work.

bsipocz:
@snbianco - Please advise here, so you want to keep this PR/have capacity to take over and rewarm or shall we close it?

Thanks!

snbianco:
I can pick this up! I'm not sure when I'll get to it, but these seem like worthwhile changes. I don't know if I have permission to push to this PR though, so I'll probably need to open a new one.

bsipocz:
yeap, picking up changes from this PR into a new one sounds all reasonable. I'll keep this one open until it's happening, but once it's done I'll go ahead and close it without merge.

Please keep the changes from here in the first commit and credit it to Jenny!

PR #2696, opened 2023-03-31
Revert kwarg-only function arguments for splatalogue

keflavich:
Splatalogue was modified to accept only keyword-specified arguments, but `Splatalogue.query_lines()` is not valid.  Currently, all of the docs have invalid examples as a result.  This simple modification fixes the problem.  The standard usage is, and should be:
```python
Splatalogue.query_lines(min_frequency, max_frequency)
```
the alternative, which is presently being enforced, should be optional:
```python
Splatalogue.query_lines(min_frequency=min_frequency, max_frequency=max_frequency)
```

There's another optional approach that is supported either way:
```python
Splatalogue.query_lines(band='alma3')
```


keflavich:
oh... ok, so this approach doesn't work.  Hrm.

bsipocz:
Hmm, the docs still has the doctest-skip-all, thus we haven't notice this before.

Why do you feel strongly against requiring being explicit? E.g. in practice all the parameters are optional as you either need the min/max or the bandname, so just assuming that if two are given they would be min/max is not exactly great.

So, what about updating the docs instead, and enabling remote testing on it?

keflavich:
The band name should be regarded as optional, imo, and the min/max frequency should not.  They are only _technically_ optional, in the literal sense of a query can be executed without raising an immediate error if you give a band name.  You cannot execute successful queries without providing either a tight pair of frequencies or a band name plus a chemical name.

This is a module I'm intimately familiar with, and I believe `Splatalogue.query_lines(1*u.GHz, 2*u.GHz)` is both the most common and most appropriate pattern.  I'd be much happier deprecating/blocking the 'band' query and making `min_frequency, max_frequency` required non-keyword arguments than leaving the module as is.  `query_lines(minfreq, maxfreq)` is much closer in spirit to `query_object(name)` or `query_region(coordinates)` than it is to a generic `query(**kwargs)` function.

I did some searching for use cases on github, too, and all uses I saw (that were not astroquery clones or my own test cases....) use positional arguments.

https://github.com/search?p=4&q=splatalogue.query_lines&type=Code

That said, I do think we should update the docs to use explicit keywords in all examples.

keflavich:
I don't understand the readthedocs failure.  I'm not sure if it's related or not; the traceback makes no sense.

keflavich:
The error messages were nonsense, but were caused by doubling up sphinx block directives

keflavich:
Maybe it was because I put them on consecutive lines?  The message never told me what I was doing was wrong specifically, but it caused errors in every _other_ module.

bsipocz:
> This is a module I'm intimately familiar with, and I believe Splatalogue.query_lines(1*u.GHz, 2*u.GHz) is both the most common and most appropriate pattern. I'd be much happier deprecating/blocking the 'band' query and making min_frequency, max_frequency required non-keyword arguments than leaving the module as is. query_lines(minfreq, maxfreq) is much closer in spirit to query_object(name) or query_region(coordinates) than it is to a generic query(**kwargs) function.

OK. I still don't like it from the pure API design POV, but could live with the two positional arguments. I would not think making them mandatory would make things better, the band kward feels very useful.

keflavich:
The main change in this PR is this one:
https://github.com/astropy/astroquery/blob/86ec38f558596e97148a7fef859d006913d3bedb/astroquery/splatalogue/core.py#L422

so I kept the arguments as optional positional arguments.

I also made regex_str:
https://github.com/astropy/astroquery/blob/86ec38f558596e97148a7fef859d006913d3bedb/astroquery/splatalogue/core.py#L79
a positional argument, since the intent of that function is to search.

Otherwise, this is all documentation fixes.  It's ready for review.

astroquery/splatalogue/load_species_table.py
```diff
@@ -66,10 +66,12 @@ def species_lookuptable(*, filename='splat-species.json', recache=False):
     # check to see if the file exists; if not, we run the
     # scraping routine
     if recache or not os.path.isfile(file_cache):
-        species = get_json_species_ids(filename)
+        species = get_json_species_ids(outfile=filename)
```
bsipocz, line 69: there are some patchy test coverage around this load/build methods, and even `__main__`, so would be nice to clean up that, too (can be done separately, I've just spotted it now).

astroquery/splatalogue/build_species_table.py
```diff
@@ -67,6 +67,7 @@ def get_json_species_ids(*, outfile='splat-species.json', base_url=conf.base_url
             species[kid['class'][0]][kid['value']] = kid.text
 
     with open(data_path(outfile), 'w') as f:
+        assert isinstance(species, dict)
```
bsipocz, line 70: why the assert in the code? This might be skipped in practice, etc.

But would be nice to have test coverage

astroquery/splatalogue/core.py
```diff
@@ -76,14 +76,14 @@ def set_default_options(self, **kwargs):
         """
         self.data.update(self._parse_kwargs(**kwargs))
 
-    def get_species_ids(self, *, restr=None, reflags=0, recache=False):
+    def get_species_ids(self, regex_str=None, *, reflags=0, recache=False):
```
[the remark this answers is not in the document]
bsipocz: and if rename happens, you may want to use a name that's more informative. Maybe `species_regex`?

bsipocz (changes requested):
OK, some more comments. The main blocker is the kwarg rename, it should be done with a deprecation decorator.

And please add a changelog about the rename, too.

astroquery/splatalogue/core.py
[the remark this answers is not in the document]
keflavich: OK, I'm willing to abandon that.  `restr` wasn't great, it was just short.  It... kind of was never intended to be a `kwarg`, just an internal required positional variable.

astroquery/splatalogue/load_species_table.py
keflavich: yes, agreed - I already expanded test coverage a bit, I think?  but, some stuff is still not caught

astroquery/splatalogue/build_species_table.py
keflavich: can remove this.  I added it as part of debugging, doesn't really belong.

astroquery/splatalogue/core.py
```diff
@@ -76,14 +76,14 @@ def set_default_options(self, **kwargs):
...
         """
         Get a dictionary of "species" IDs, where species refers to the molecule
         name, mass, and chemical composition.
 
         Parameters
         ----------
-        restr : str
+        regex_str : str
             String to compile into an re, if specified.   Searches table for
```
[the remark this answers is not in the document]
keflavich: it's not an implementation detail exactly, but yes, this could be rephrased.  It's important that the user know that the string is intended as a python `re` regular expression, since `re` languages differ between every implementation.

astroquery/splatalogue/core.py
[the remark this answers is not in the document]
bsipocz: haha, good reminder that internal stuff should be named well, too. Or at least the stuff that ends up in signatures and docstrings.
I'm happy to push commits to this, but it will need to wait until at least the evening.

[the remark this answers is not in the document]
keflavich: I renamed the variable, but I still prefer to keep it as a positional argument.  I added the deprecation warning.

docs/splatalogue/splatalogue.rst
```diff
@@ -1,5 +1,3 @@
-.. doctest-skip-all
```
bsipocz, line 1: Interesting why the examples work in this without remote-data? Would it make sense to make it opt-out to use the shipped json?

docs/splatalogue/splatalogue.rst
```diff
@@ -35,51 +29,53 @@ To search within this list for a particular species, you can use regular express
...
+   105
+   >>> just_CO = Splatalogue.get_species_ids(species_regex=' CO ') # note the spaces
    >>> len(just_CO)
    4
-   >>> just_CO  # includes different vibrationally excited states
-   {u'02812 CO v = 0 - Carbon Monoxide': u'204',
-    u'02813 CO v = 1 - Carbon Monoxide': u'990',
-    u'02814 CO v = 2 - Carbon Monoxide': u'991',
-    u'02815 CO v = 3 - Carbon Monoxide': u'1343'}
+   >>> assert just_CO == {'02812 CO v = 0 - Carbon Monoxide': '204',
```
bsipocz, line 38: why the asserts in the documentation?

bsipocz (commented):
More comments regarding the documentation, basically the asserts should go, but otherwise things are looking good.

docs/splatalogue/splatalogue.rst
keflavich: I should've left myself a note about this... iirc, there was a problem with the non-assert version. I'll remove and see what happens.

docs/splatalogue/splatalogue.rst
keflavich: Many of them are labeled remote-data; only a couple work without it.

docs/splatalogue/splatalogue.rst
keflavich: Looks like tests pass. I bet it was some local configuration error.

keflavich:
@bsipocz any idea what's going on with the dev deps build?  Looks like pyvo and codecov don't get along?

bsipocz:
> any idea what's going on with the dev deps build? Looks like pyvo and codecov don't get along?

Ahh, so codecov actually bite us. cc @pllim 
I'll fix this separately.

bsipocz (approved):
When CI is all clean, this is good to go.

PR #2712, opened 2023-04-24
ALMA: auth handling to login to Keycloak

at88mph:
Added auth handling to login to Keycloak.

Fixes #2670

### DIFF
--- a/astroquery/alma/core.py
+++ b/astroquery/alma/core.py
@@ -212,11 +212,14 @@ def _gen_sql(payload):
     return sql + where
 
 
-#
-# Authentication session information for connecting to an OIDC instance.  Assumes an OIDC system like Keycloak
-# with a preconfigured client called "oidc".
-#
 class AlmaAuth(BaseQuery):
+    """Authentication session information for passing credentials to an OIDC instance
+
+    Assumes an OIDC system like Keycloak with a preconfigured client app called "oidc" to validate against.
+    This does not use Tokens in the traditional OIDC sense, but rather uses the Keycloak specific endpoint
+    to validate a username and password.  Passwords are then kept in a Python keyring.
+    """
+
     _CLIENT_ID = 'oidc'
     _GRANT_TYPE = 'password'
     _INVALID_PASSWORD_MESSAGE = 'Invalid user credentials'
@@ -229,7 +232,12 @@ class AlmaAuth(BaseQuery):
         self._auth_hosts = auth_urls
         self._auth_host = None
 
-    def set_auth_hosts(self, auth_hosts):
+    @property
+    def auth_hosts(self):
+        return self._auth_hosts
+
+    @auth_hosts.setter
+    def auth_hosts(self, auth_hosts):
         """
         Set the available hosts to check for login endpoints.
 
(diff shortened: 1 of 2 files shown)


keflavich:
@andamian could you provide the review for this one?  It looks good to me but I haven't tested and I know nothing about the ALMA login protocol.

astroquery/alma/core.py
```diff
@@ -212,6 +212,91 @@ def _gen_sql(payload):
     return sql + where
 
 
+#
+# Authentication session information for connecting to an OIDC instance.  Assumes an OIDC system like Keycloak
+# with a preconfigured client called "oidc".
```
andamian, line 217: In spite of the name and the description, this class only checks that the user/password is valid with any of the Keycloak systems deployed in a list of URLs. Nothing else from the key cloak functionality (such as tokens) are further used in the scheme (but maybe it should?). 

This confused me a great deal as I was trying to figure out what kind of magic Keycloak employs to facilitate authentication on the services so I suggest adding the above paragraph or something to that extend to make it clear and prevent future confusion.

astroquery/alma/core.py
```diff
@@ -212,6 +212,91 @@ def _gen_sql(payload):
...
+    _REALM_ENDPOINT = '/auth/realms/ALMA'
+    _LOGIN_ENDPOINT = f'{_REALM_ENDPOINT}/protocol/openid-connect/token'
+    _VERIFY_WELL_KNOWN_ENDPOINT = f'{_REALM_ENDPOINT}/.well-known/openid-configuration'
+
+    def __init__(self):
+        super().__init__()
+        self._auth_hosts = auth_urls
+        self._auth_host = None
+
+    def set_auth_hosts(self, auth_hosts):
```
andamian, line 232: I would make this a property setter and fail on `None`. This way it gets invoked in the ctor too.

astroquery/alma/core.py
```diff
@@ -212,6 +212,91 @@ def _gen_sql(payload):
...
+        auth_hosts : array
+            Available hosts name.  Checking each one until one returns a 200 for
+            the well-known endpoint.
+        """
+        self._auth_hosts = auth_hosts
+
+    @property
+    def host(self):
+        if self._auth_host is None:
+            for auth_url in self._auth_hosts:
```
andamian, line 247: Currently this returns a generic `TypeError: 'NoneType' object is not iterable` if `self._auth_hosts` is not set properly. See my comment above.

astroquery/alma/core.py
```diff
@@ -212,6 +212,91 @@ def _gen_sql(payload):
...
+        json_auth = login_response.json()
+
+        if 'error' in json_auth:
+            log.debug(f'{json_auth}')
+            error_message = json_auth['error_description']
+            if self._INVALID_PASSWORD_MESSAGE not in error_message:
+                raise LoginError("Could not log in to ALMA authorization portal: "
+                                 f"{self.host} Message from server: {error_message}")
+            else:
+                log.error(error_message)
```
andamian, line 292: Missing `LoginError`. I think it should always raise an exception unless it succeeds.

astroquery/alma/core.py
```diff
@@ -212,6 +212,91 @@ def _gen_sql(payload):
...
+        Authenticate to one of the configured hosts.
+
+        Parameters
+        ----------
+        username : str
+            The username to authenticate with
+        password : str
+            The user's password
+        """
+        data = {
```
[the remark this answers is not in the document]
at88mph: That might be true.  If the `self._auth_host` is set, that just means that `self.get_valid_host()` was called at some point, not that the current user is logged in.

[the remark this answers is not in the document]
andamian: Right. Please ignore my comment.

andamian (approved):

andamian:
@bsipocz - FYI: this is ready from our perspective.

keflavich (approved):

bsipocz:
Thank you @at88mph and @andamian! I see two, genuine-looking alma failures, but both are unrelated and are present on `main`, too.

PR #2729, opened 2023-05-05
Fixing CI hiccups from `astroquery.mast` unit test 

jaymedina:
### Summary

The dev dependency test was failing due to an error in the computation of the offset between the resolved object and its sky coordinates from the literature. It was infinitesimally small, something like 10**-15, so I rounded it to the neares 10**-10.

### Task List
- [x] Rounded value to nearest 10th decimal place to fix systematic error in offset computation (`astroquery/mast/test_mast.py`)


jaymedina:
I fixed the dev dependency error but I'm actually not sure why the readthedocs is failing. The error message at the top is suggesting there could be a dependency conflict in astroquery? @keflavich https://readthedocs.org/projects/astroquery/builds/20444648/

PR #2856, opened 2023-10-19
Feature simbad query tap

ManonMarchand:
Hello, 

## What this PR does

Add `Simbad.query tap` functionality to the simbad module. This is called like this: 

```python
Simbad.query_tap("SELECT top 5 from ref")
```

This feature comes with helpers methods to explore SIMBAD's relational schema. 
- `Simbad.list_tables()` returns the public tables of SIMBAD
- `Simbad.list_columns(*args, keyword: str = None)` returns the columns corresponding to the entry table(s) and with matches for the given keyword
- `Simbad.list_linked_tables(table: str)` returns the non obvious (i.e. the columns don't have the same names) possible joins

Properties to connect to the tap endpoint:
- `Simbad.simbad_mirrors` is a set of the two simbad's mirrors
- `Simbad.mirror` takes the value of `conf.server` at instantiation but can be switched later
- `Simbad.tap` is created from the current `Simbad.mirror`

And the private method
- `_adql_parameter`
that will be called when building adql queries from user inputs (to silently escape invalid characters and reserved sql vocabulary)

## Documentation-wise

I took the liberty to add a graphviz graph of simbad's schema since sphinx-graphviz is already in the dependencies of sphinx-automodapi but we can remove it if needed.

The documentation on queries are now organized in a rough 
- everything about objects
- everything about bibliography
- queries that mix both
This might need discussion too ?

That was also a lot of text and I'm not a native-english speaker, if someone has the time to point mistakes/unclear parts I'd love feedback.

## Next steps

This PR only adds new methods. The rework to solve #1468 will be based on `Simbad.query_tap` and the two private functions introduced here.


ManonMarchand:
On the coverage, some lines are marked as missed but are tested in `test_simbad_remote.py`, is it normal?

bsipocz:
Thank you so much! I'll try to get a timely review for this.

To answer the questions above:

> On the coverage, some lines are marked as missed but are tested in test_simbad_remote.py, is it normal?

yes, do ignore the number here, its main purpose is to check whether there is a significant drop in coverage and we are aware that the actual percentage is most often larger when one includes the remote tests (I tend to run those numbers for significantly big PRs like this one as part of the review).

Re pyvo 1.4.2: that version is a bit too new to mandate as a minimum required version. Maybe you could do a try/except import and later a version dependent conditional?

ManonMarchand:
I added the try/except clause for DALOverflowWarning :slightly_smiling_face:

astroquery/simbad/core.py
```diff
@@ -936,6 +1006,243 @@ def query_objectids_async(self, object_name, *, cache=True,
...
+        """
+        query = ("SELECT from_table, from_column, target_table, target_column"
+                 " FROM TAP_SCHEMA.key_columns JOIN TAP_SCHEMA.keys USING (key_id)"
+                 f" WHERE (from_table = '{_adql_parameter(table)}')"
+                 f" OR (target_table = '{_adql_parameter(table)}')")
+        if get_adql:
+            return query
+        return self.query_tap(query)
+
+    def query_tap(self, query: str, maxrec=10000, uploads=None):
```
bsipocz, line 1150: make optional kwargs kwarg only. Here, and in all the other methods. Also, we don't really do any type hinting, but it would not hurt to leave it in there.

```suggestion
    def query_tap(self, query: str, *, maxrec=10000, uploads=None):
```

astroquery/simbad/core.py
```diff
@@ -260,7 +295,7 @@ def _request(self, *args, **kwargs):
 
 
 @async_to_sync
-class SimbadClass(SimbadBaseQuery):
+class SimbadClass(SimbadBaseQuery, BaseVOQuery):
```
bsipocz, line 298: The base classes should have a reversed order here.

astroquery/simbad/core.py
```diff
@@ -300,6 +335,41 @@ class SimbadClass(SimbadBaseQuery):
     def __init__(self):
         super().__init__()
         self._VOTABLE_FIELDS = self._VOTABLE_FIELDS.copy()
+        self._mirror = conf.server
+        self._tap = None
+
+    @property
+    def simbad_mirrors(self):
+        """Set of the two Simbad mirrors domains."""
+        return {'simbad.cds.unistra.fr', 'simbad.harvard.edu'}
```
bsipocz, line 344: We have this in the config already. If it all being moved into class properties then I suppose it should be cleanup from the config. (Though keeping the URLs in the config, rather than in the core.py is more consistent, but no strong preference)

astroquery/simbad/core.py
```diff
@@ -936,6 +1006,243 @@ def query_objectids_async(self, object_name, *, cache=True,
 
         return response
 
+    def tables(self, get_adql=False):
```
bsipocz, line 1009: In general I'm more in favour of using more descriptive method names. We don't have a strong precedent yet, but `list_tables` or `get_tables` may work.

astroquery/simbad/core.py
```diff
@@ -936,6 +1006,243 @@ def query_objectids_async(self, object_name, *, cache=True,
...
+        `~astropy.table.table.Table`
+        """
+        query = ("SELECT table_name, description"
+                 " FROM TAP_SCHEMA.tables"
+                 " WHERE schema_name = 'public'")
+        if get_adql:
+            return query
+        return self.query_tap(query)
+
+    def columns(self, *tables: str, get_adql=False):
```
bsipocz, line 1028: Same, add a verb for the method name to better describe it, `list_columns` maybe?

astroquery/simbad/core.py
```diff
@@ -936,6 +1006,243 @@ def query_objectids_async(self, object_name, *, cache=True,
...
+        if len(tables) == 1:
+            query += f" AND table_name = '{tables[0]}'"
+        elif len(tables) > 1:
+            query += f" AND table_name IN ({str(tables)[1:-1]})"
+        query += " ORDER BY table_name, principal DESC, column_name"
+        if get_adql:
+            return query
+        return self.query_tap(query)
+
+    def find_columns_by_keyword(self, keyword: str, get_adql=False):
```
bsipocz, line 1067: maybe this can be merged into the previous method? by adding the optional `keyword` kwarg to it?

astroquery/simbad/tests/test_simbad_remote.py
```diff
@@ -252,3 +257,45 @@ def test_multi_vo_fields(self):
...
+        result = Simbad.query_tap("SELECT * FROM TAP_UPLOAD.numbers "
+                                  "JOIN TAP_UPLOAD.letters USING(letters)",
+                                  uploads={"numbers": table_numbers, "letters": table_letters})
+        expect = "letters numbers\n------- -------\n      a       1\n      b       2\n      c       3"
+        assert expect == str(result)
+        # Test query_tap raised errors
+        # DALOverflowWarning exists since pyvo 1.4.2
+        with pytest.raises(DALOverflowWarning, match="Partial result set *"):
+            truncated_result = Simbad.query_tap("SELECT * from basic", maxrec=2)
+            assert len(truncated_result) == 2
```
bsipocz, line 277: This doesn't work for old pyvo

docs/simbad/query_tap.rst
```diff
@@ -0,0 +1,274 @@
...
+  category, specific redshifts, and the results should have an NGC name in their
+  list of names.*/
+  WHERE (otype != 'star..') AND (rvz_redshift < 1) AND (id LIKE 'NGC%')
+  /*The result is then sorted so that the top 5 selected corresponds to
+  the objects cited by the largest number of papers.*/
+  ORDER BY nbref DESC
+
+This ADQL query can be called with `~astroquery.simbad.SimbadClass.query_tap`: 
+
+.. a bit long due to the ordering
```
bsipocz, line 34: can we make it faster, e.g. the ordering is not really needed here if it makes it much slower?

docs/simbad/query_tap.rst
```diff
@@ -0,0 +1,274 @@
...
+            basic rvz_radvel_prec ...                                           
+            basic        rvz_type ...                                           
+    mesVelocities          origin ...                                  meta.note
+
+Example TAP queries
+^^^^^^^^^^^^^^^^^^^
+
+This section lists more complex queries by looking at use cases from former astroquery issues.
+
+`Getting all bibcodes containing a certain type of measurement for a given object. <https://github.com/astropy/astroquery/issues/243>`__
```
bsipocz, line 180: No need to link to github issues, remove the link from the heading here and below, too

bsipocz (commented):
I made a couple of comments while travelling, will come back for another fine combing before merging, but honestly I don't think anything else critical will be found.

astroquery/simbad/core.py
ManonMarchand: Ah yes, this is actually because I was unable to get the other server from there (see the question here https://github.com/astropy/astroquery/issues/2291#issuecomment-1755445731). How do we access the second value of server from the config?

astroquery/simbad/core.py
ManonMarchand: Much clearer, thanks a lot for the suggestion. I chose to use `list_tables`, `list_columns` and `list_linked_tables`

astroquery/simbad/core.py
ManonMarchand: Ah yes, good idea. This is implemented in the last commit from today :)

docs/simbad/query_tap.rst
ManonMarchand: I could remove it, but I wanted that first query to present every TAP step (select, from, where, order by). The doctest output would also need to be skipped because the output will change very often. Is it a big issue that this one is skipped?

bsipocz: no, in this case it's OK to skip it then

ManonMarchand: Actually, timing it with 

```python
import timeit
timeit.timeit("""Simbad.query_tap("SELECT TOP 5 basic.ra, basic.dec, main_id, nbref FROM basic JOIN ident ON basic.oid = ident.oidref WHERE ((otype != 'Cl*..') AND (rvz_redshift < 1) AND (id LIKE 'NGC%')) ORDER BY nbref DESC")""", number=1, setup="from astroquery.simbad import Simbad")
```

Gives 1.04s on my machine. It's slower than the average query (~0.15s) but not dramatic.

bsipocz: Oh, that's not slow at all, at least we have much bigger problems to clear out. 

As we don't run the remote test for PRs, I would indeed suggest to change the skip to an ignore output, so we would have a chance of picking up errors if/when it stops working.

astroquery/simbad/core.py
ManonMarchand: I saw that but the error that is raised does not help much someone who does not know beforehand which servers are available

```python
from astroquery.alma import conf
conf.archive_url = "anything"
```

returns

```
TypeError: Provided value for configuration item archive_url not valid: the value "anything" is unacceptable.
```

While with this way of using properties, we'd have 

```python
Simbad.mirror = "anything"
```

which returns

```
ValueError: 'anything' does not correspond to a Simbad mirror, the two existing ones are {'simbad.harvard.edu', 'simbad.cds.unistra.fr'}.
```

this can help a bit more (for example I'm always forgetting that it it simbad.cds.unistra.fr and not simbad.unistra.fr)

(and reading the two versions now makes me think that it might be better with a property name like Simbad.url or Simbad.server. What do you think?)

pllim: I *think* there is a way to access the list. I mean, the config has to have it somewhere for validation. I'll dig later after dropping kid at school.

astroquery/simbad/core.py
ManonMarchand: Done :)

docs/simbad/query_tap.rst
ManonMarchand: Done :)

ManonMarchand:
Oups, I rebased too far sorry. I have no idea how to fix that

astroquery/simbad/tests/test_simbad_remote.py
ManonMarchand: I think it's fixed now.

astroquery/simbad/core.py
bsipocz: > I saw that but the error that is raised does not help much someone who does not know beforehand which servers are available

Thus there is the list added in alma outside of the config. But of course something nicer, directly from the config would be the best.

bsipocz:
> Oups, I was squashing my commits and rebased too far sorry. I have no idea how to fix that

I suspect you rebased on an outdated commit, rather than the freshly fetched `main` branch from the astropy/upstream remote.

ManonMarchand:
Will it be an issue when merging?

bsipocz:
> Will it be an issue when merging?

No, I'll do a rebase before merge ;)

astroquery/simbad/core.py
pllim: It is a little convoluted but the info is there. Here is an example from core lib `astropy`:

```python
>>> from astropy.utils.iers import conf
>>> c = conf.__class__.__dict__["iers_degraded_accuracy"]  # Grabs the ConfigItem instance
>>> c.cfgtype  # Valid options encoded in the return string
'option(error, warn, ignore)'
```

Here is how it is defined at a high level: https://github.com/astropy/astropy/blob/1ae26ae0fcf5b8ffe0563d73be7787e9c835bf0e/astropy/utils/iers/iers.py#L190-L195

```python
class Conf(_config.ConfigNamespace):
    iers_degraded_accuracy = _config.ConfigItem(
        ["error", "warn", "ignore"],
        "IERS behavior if the range of available IERS data does not "
        "cover the times when converting time scales, potentially leading "
        "to degraded accuracy.",
    )
```

The implementation predates me and a lot of the internals are handed off to the external configobj package that astropy bundles. So don't ask me why it has to be like this. Hope this helps!

ManonMarchand: It does! Thanks a lot.

astroquery/simbad/core.py
```diff
@@ -335,42 +336,51 @@ class SimbadClass(BaseVOQuery, SimbadBaseQuery):
...
-    def mirror(self, server: str):
+    @server.setter
+    def server(self, server: str):
         f"""Allows to switch server between Simbad mirrors.
 
         Parameters
         ----------
         server : str
-            It should be one of {self.simbad_mirrors}.
+            It should be one of {conf.servers_list}.
```
bsipocz, line 354: `servers_list` is not in conf, you can say the full path of `astroquery.simbad.servers_list` or just leave it as `servers_list`.

astroquery/simbad/core.py
bsipocz: yes, I think this alma solution sounds reasonable, and way more readable and understandable than the suggested workaround. I suspect the whole config system has a lot of historical debt as being an old and never really touched up part of the package.

bsipocz:
> EDIT : and I'll stop changing stuff so that you can review, sorry again.

No worries, do add as many changes as you wish, you can always add the review label, or ping me when done and I'll come back then.

astroquery/simbad/core.py
pllim: Yes, that is definitely cleaner than accessing config in weird places. 👏

docs/simbad/query_tap.rst
```diff
@@ -0,0 +1,322 @@
+`~astroquery.simbad.SimbadClass.query_tap` (for Table Access Protocol) is the one
+query to rule them all. It allows to interrogate all the information in SIMBAD with the
```
keflavich, line 2:
```suggestion
query to rule them all. It allows one to access all the information in SIMBAD with the
```
Minor English, but "Allow to <verb>" isn't an accepted construction.  Also, "interrogate" carries connotations that mean it's not so commonly used; it's technically acceptable, but I'd recommend "access" or "query" instead.

docs/simbad/query_tap.rst
```diff
@@ -0,0 +1,322 @@
...
+           deg                deg                       
+         float64            float64        object  int32
+    ------------------ ------------------ -------- -----
+    10.684708333333333 41.268750000000004    M  31 12412
+    13.158333333333333 -72.80027777777778 NAME SMC 10875
+       187.70593076725 12.391123246083334    M  87  7040
+    148.96845833333333  69.67970277777778    M  82  5769
+        23.46206906218 30.660175111980003    M  33  5737
+
+And voilà, we get the 5 NGC objects that are the most cited in literature, are not cluster
```
keflavich, line 54:
```suggestion
And voilà, we get the 5 NGC objects that are the most cited in literature, are not clusters
```

docs/simbad/query_tap.rst
```diff
@@ -0,0 +1,322 @@
...
+    mesDiameter      oidref        basic           oid
+
+The output indicates that the ``mesDiameter`` table can be linked to ``basic`` with the following
+join statement: ``[...] mesDiameter JOIN basic ON mesDiameter.oidref = basic.oid [...]``.
+
+
+.. graphviz:: simbad-er.gv
+    :layout: neato
+    :caption: A quick view of SIMBAD's tables. Hover the links to see the linked columns.
+    :alt: This interactive graph summarizes the information that can be obtained with `~astroquery.simbad.SimbadClass.list_tables` and `~astroquery.simbad.SimbadClass.list_linked_tables`.
```
keflavich, line 127: are the links working?  I don't see hovers.  But this renders nicely!
image

keflavich: is it possible to increase the font size though?  that might be a good idea, to match the overall fontsize of the document.  It's OK if that's not possible.

docs/simbad/query_tap.rst
```diff
@@ -0,0 +1,322 @@
...
+           ref   last_page     INTEGER ...               meta.bib.page
+           ref    nbobject     INTEGER ...                 meta.number
+           ref      oidbib      BIGINT ...        meta.record;meta.bib
+           ref        page     INTEGER ...               meta.bib.page
+           ref       title        CLOB ...                  meta.title
+           ref      volume     INTEGER ...             meta.bib.volume
+           ref        year    SMALLINT ...          meta.note;meta.bib
+
+`~astroquery.simbad.SimbadClass.list_columns` can also be called with a keyword argument.
+This returns columns from any table for witch the  given keyword is either in the table name,
```
keflavich, line 159:
```suggestion
This returns columns from any table for witch the given keyword is either in the table name,
```

docs/simbad/simbad.rst
```diff
@@ -16,15 +16,15 @@ formulated. If successful all the queries will return the results in a
 
 A warning about big queries
 ---------------------------
-The SIMBAD database has limited querying capacity.  If you spam the server with
-queries, you may be temporary blacklisted.  The rate limit may vary, but you
-should not submit more than ~5-10 queries per second.
 
-If you want to perform large queries, we suggest using vectorized queries
-when possible.  You can pass `~astroquery.simbad.SimbadClass.query_region`
+The SIMBAD database is largely used and had to limit the rate of incoming queries.
```
keflavich, line 20:
```suggestion
The SIMBAD database is widely used and has to limit the rate of incoming queries.
```

keflavich (approved):
I have a couple nitpicky English comments, but this looks great and the documentation is super helpful.

docs/simbad/query_tap.rst
ManonMarchand: Thanks for the review and all the English tips! The links work for me only when I use a local server to see the docs, there might be a bit of javascript to make it work in the sphinx-graphviz library, I'm not sure. I hope it will work too when it's live in astroquery's documentation. Otherwise it'll be a static graph and we'll have to change the legend. I'm looking for a way to change the font size :)

ManonMarchand: A peek: 
Peek 18-10-2023 15-45

ManonMarchand: The fontsize is changed... but in a bit of a ugly way. It has to be specified in each node in the DOT file and [it only works with pixels](https://graphviz.org/docs/attrs/fontsize/ "https://graphviz.org/docs/attrs/fontsize/"), not 'em'. I chose 16px as it seems to be the standard size of most browsers but it won't adapt to the readers' browsers parameters if they changed it.

For the hovering labels, I don't know. I tried a fresh python environment to generate the docs and it still worked. I noticed the tooltips only appear for me when I explicitly select the browser's window by clicking on it... Otherwise it might be an upstream bug with sphinx-graphviz?

Also, should I add in the comments of the `query_tap.rst` file that there is a script that can generate the file anew in case Simbad's schema changes? https://github.com/ManonMarchand/simbad-ER-diagram/blob/main/simbad-and-graphviz.py

ManonMarchand:
All done for Adam comments

astroquery/simbad/core.py
```diff
@@ -300,6 +336,50 @@ class SimbadClass(SimbadBaseQuery):
...
+        """
+        if server in conf.servers_list:
+            self._server = server
+        else:
+            raise ValueError(f"'{server}' does not correspond to a Simbad server, "
+                             f"the two existing ones are {conf.servers_list}.")
+
+    @property
+    def tap(self):
+        """A ``~pyvo.dal.tap.TAPService`` service for Simbad."""
```
bsipocz, line 364: we should be able to intersphinx link to pyvo, the trick is to link to the namespace its API is documented and not to the namespace the source code is
```suggestion
        """A `~pyvo.dal.TAPService` service for Simbad."""
```

astroquery/simbad/core.py
```diff
@@ -300,6 +336,50 @@ class SimbadClass(SimbadBaseQuery):
...
+    @property
+    @lru_cache(1)
+    def hardlimit(self):
+        """The maximum number of lines for Simbad's output.
+
+        This property is cached to avoid calls to simbad's capability
+        webpage each time the getter is called.
+        """
+        # replace stack of property and lru_cache by functools.cache_property when
+        # astroquery drops python 3.7 support
```
bsipocz, line 381: ahh, reminder that we should indeed drop support.

astroquery/simbad/core.py
```diff
@@ -936,6 +1016,256 @@ def query_objectids_async(self, object_name, *, cache=True,
...
+        """The names and descriptions of the tables in SIMBAD.
+
+        Parameters
+        ----------
+        get_adql : bool, optional
+            Returns the ADQL string instead of querying SIMBAD.
+
+        Returns
+        -------
+        `~astropy.table.table.Table`
```
bsipocz, line 1029: interesting that this link works, I would expect a sphinx warning and therefore a red CI status for this
```suggestion
        `~astropy.table.Table`
```

astroquery/simbad/core.py
```diff
@@ -936,6 +1016,256 @@ def query_objectids_async(self, object_name, *, cache=True,
...
+        looks for columns in the selected Simbad tables that contain the
+        given keyword. The keyword search is not case-sensitive.
+
+        Parameters
+        ----------
+        *tables : str, optional
+            Add tables names as strings to restrict to these tables columns.
+        keyword : str, optional
+            A keyword to look for in column names, table names, or descriptions.
+        get_adql : bool, optional
```
bsipocz, line 1053: we may need to call this `get_query_payload` to be consistent with the rest of the astroquery API, though your name is more precise and makes sense. 

What do you think @keflavich, is it time to cleanup what `get_query_payload` does and return the sql/adql where it makes more sense than returning the GET/POST parameters?

astroquery/simbad/core.py
```diff
@@ -936,6 +1016,256 @@ def query_objectids_async(self, object_name, *, cache=True,
...
+             basic       otype  VARCHAR ...                  src.class
+        """
+        query = ("SELECT table_name, column_name, datatype, description, unit, ucd"
+                 " FROM TAP_SCHEMA.columns"
+                 " WHERE table_name NOT LIKE 'TAP_SCHEMA.%'")
+        # select the tables
+        if len(tables) == 1:
+            query += f" AND table_name = '{tables[0]}'"
+        elif len(tables) > 1:
+            query += f" AND table_name IN ({str(tables)[1:-1]})"
```
bsipocz, line 1099: I don't really get this logic without some more background info, how many `tables` are expected here? Why drop the first and last ones? Could you maybe add a one-line comment or reference above the line?

astroquery/simbad/core.py
bsipocz: there are other occurances, those should be fixed, too.

bsipocz (commented):
Overall looks good, I have a couple of minor comments and questions.

and There are some unrelated commmits picked-up, and small doctest failures, so could you please rebase and fix the doctest?

astroquery/simbad/core.py
ManonMarchand: Should I keep the comment to help when the support will be dropped, or is it noise?

astroquery/simbad/core.py
ManonMarchand: Thanks for bringing this, I forgot to raise my concerns about this argument.

I was also thinking about a `get_raw_votable` (or `get_votable`) option. The use cases I see are:
- for debug purposes when the votable does not parse correctly. This can occur because astropy's parser is more strict than the one of TOPCat that we use for our tests (see #2916)
- to use votable sugar (examples are MIVOT annotations that will appear with https://github.com/astropy/pyvo/pull/497, or a simple SAMP broadcast with `astropy.samp`, it works well, I tried.)

But I did not want to fall in the boolean trap cause these two arguments (`get_adql\get_query_payload` and `get_votable`) would be mutually exclusive.

Then my idea was to maybe deviate from the API in place and introduce an `output_format` argument that'd accept a `Literal(['astropy_table', 'adql_string', 'astropy_votable'])` and would default to `astropy_table`. 

What are your thoughts?

astroquery/simbad/core.py
ManonMarchand: Hum it's useless. I was removing the parenthesis around the string representation of the tuple and then I was adding them around again.

ManonMarchand: And for the other question, tables is a tuple with zero or more items.

```python
Simbad.list_columns()
```

will return all columns of all tables while 

```python
Simbad.list_columns("ref")
```

will returns columns of the `ref` table only and 

```python
Simbad.list_columns("cat", "ident", "ids")
```

will return columns of `cat`, `ids` and `ident`

astroquery/simbad/core.py
ManonMarchand: Done with https://github.com/astropy/astroquery/pull/2856/commits/d2b2bc4b8325e713a40d3150735d30d42e22760f :slightly_smiling_face:

ManonMarchand:
I removed the `_adql_name` method because it didn't prove useful in the next steps of astroquery.simbad's refactoring.

astroquery/simbad/core.py
bsipocz: please do keep comments like these. Whenever we drop support I usually go around grep for things like `3.7`, etc, so will more likely to find this for cleanup because of the comment.

astroquery/simbad/core.py
bsipocz: > I was also thinking about a get_raw_votable (or get_votable) option.

Yes, those would be great and I vaguely recall that it came up multiple times in the past and there should be issues for it already either here or in pyvo.

> Then my idea was to maybe deviate from the API in place and introduce an output_format argument that'd accept ['astropy_table', 'adql_string', 'astropy_votable', 'raw_votable'] and would default to astropy_table. This could be extended too.
What are your thoughts?

lets factor out this discussion into a separate ticket and pull into the discussion other interested parties (I think Adrian will definitely be interested, maybe the ESA folks, too, even though they don't use pyvo (yet)). I might use a different arg name that makes is clearer that it's a debug option, and may not support all 4 options you listed, but overall it looks useful to have more tools for easy troubleshooting purposes.

bsipocz (approved):
Thank you @ManonMarchand, this should go in now.

PR #2919, opened 2024-01-07
Update core.py

AnkitHMishra:
Changed the wl to w in line 91 and 92


keflavich:
This is now redundant with #2918 - I'm closing this in favor of that one, which includes a test.  Thanks for the fix @AnkitHMishra !

AnkitHMishra:
I have tried to change the core.py lines from low_wl to low_w and upp_wl to
upp_w but the problem still persists.

AnkitHMishra:
Yes it works now. Thanks for your help

PR #2931, opened 2024-01-25
EXP: check if upstream patch in astropy fixes devdeps

neutrinoceros:
I think https://github.com/astropy/astropy/pull/15948 should resolve errors currently seen here in devdeps. Let's put that to the test.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,7 +28,7 @@ setenv =
 deps =
     devdeps: numpy>=0.0.dev0
     devdeps: matplotlib>=0.0.dev0
-    devdeps: astropy>=0.0.dev0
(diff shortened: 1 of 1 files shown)


neutrinoceros:
ah, of course my branch needs to be built from source, and by default this happens in isolation against numpy 1.x, which then crashes pytest on collection...

neutrinoceros:
I couldn't make it work with tox but I was able to run tests locally with every dependency compiled against numpy 2 and, compared with the [reference run](https://github.com/astropy/astroquery/actions/runs/7651124450/job/20848349480?pr=2929), I got rid of all errors (`ValueError: Input parameters lon and lat cannot be broadcast`) which was my goal with https://github.com/astropy/astropy/pull/15948

PR #2975, opened 2024-03-28
Update documention for the field upload_resource in the functions launch_job and launch_job_async [ci skip]

cosmoJFH:
Following the discussion in the issue [2529 ](https://github.com/astropy/astroquery/issues/2529), we would like to update the documentation for the field upload_resource in the functions launch_job and launch_job_async

cc @esdc-esac-esa-int


keflavich (approved):
This is a very minor documentation change, so we can go ahead with it.

bsipocz (approved):
Thanks! The RTD failure is unrelated and has been fixed in `main` since, so we should merge this now.

PR #2986, opened 2024-04-09
JPL Horizons: Test refraction with and extra_precision

mkelley:
Added column definitions to address #2977 .  In addition, reviewed other possible column return values and added 
 a few that may be missing (see https://ssd.jpl.nasa.gov/horizons/manual.html#output ).


bsipocz (approved):
Thank you @mkelley!

PR #3012, opened 2024-05-25
Gaia: delete any reference to data_structure "Combined" in the function load_data and in the documentation [ci skip]

cosmoJFH:
In a near future, the gaia datalink service is going to generate ancillary products for data_structure as "Individual" or "Raw", so the option "Combined" will not be available anymore.

We would like to delete any reference to this data_structure in the method  _load\_data_ as well as in the general documentation.

cc @esdc-esac-esa-int 

Jira ticket: GAIAMNGT-1704


bsipocz (approved):
Thank you. I'm on the fence of this needs a changelog, after all we don't do any validation just passing it directly to the server.

So for now I leave it as is, and revisit at release time.

PR #3069, opened 2024-07-17
Bugfix in get_cloud_uri_list, added test case

snbianco:
This is to fix a small bug in the last PR that was merged for MAST. 

I also fixed a `KeyError` so that error messages give necessary information, and I added a test case for downloading files from the cloud so that we catch similar bugs in the future.


bsipocz (approved):
Thanks!

PR #3070, opened 2024-07-19
BUG: fix an incoming incompatibility with Python 3.15 (`locale.getdefaultlocale` is deprecated in Python 3.12)

neutrinoceros:
See
https://docs.python.org/3/library/locale.html#locale.getdefaultlocale
and
https://docs.python.org/3/library/locale.html#locale.getencoding

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -136,6 +136,9 @@ Infrastructure, Utility and Other Changes and Additions
 
 - Versions of PyVO <1.5 are no longer supported. [#3002]
 
+- Fix an incoming incompatibility with Python 3.15 (`locale.getdefaultlocale`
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Good question. I think it originated in `astropy-helpers`, and [`package-template` has this to say](https://github.com/astropy/package-template/blob/36ed48db74c9a0bfcf214e14a1a3fb1fdaae0c6c/docs/ape17.rst#L48) about it
> Next you should remove the ah_bootstrap.py file:

so maybe it should just be removed here instead of fixed ?

bsipocz (commented):
Lets leave the helpers alone for now. The issue has been patched before and seems to be sufficient for CI, and anyway, it will all be resolved once I rework the whole packaging in one go.

PR #3071, opened 2024-07-19
DEP: refactored out runtime dependency on setuptools

neutrinoceros:
Context:
- `pkg_resources`, which is part of `setuptools`, is de facto a runtime dependency of `astroquery`
- it is not declared as such: I suspect the reason this has not been a problem (yet) in CI is that some test dependency (likely pytest-doctestplus via pytest-astropy) pulls it in. I'm working on eliminating this dependency from the stack, see https://github.com/scientific-python/pytest-doctestplus/pull/258
- `pkg_resources` is also deprecated as an API (see https://setuptools.pypa.io/en/latest/pkg_resources.html)

I followed the upstream [migration guide](https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename) here.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -136,6 +136,8 @@ Infrastructure, Utility and Other Changes and Additions
 
 - Versions of PyVO <1.5 are no longer supported. [#3002]
 
+- Dropped ``setuptools`` as a runtime dependency. [#3071]
+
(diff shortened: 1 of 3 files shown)


ceb8:
Thanks @neutrinoceros !

PR #3079, opened 2024-08-02
Astrometry.net: reduce number of calls when polling for job status

### LINKED ISSUE #3078: Astrometry.net: can query for job id once, rather than each time
Hi,

Minor request: in the Astrometry.net module, while polling for a submission to get solved, it looks like you're hitting two URLs each time - one to get the job id from the submission id, and then one to get the job status.  I think you only need to get the job id once.  So this could split into two loops, one where you poll until you get the job id, and then keep the job id fixed and poll for the job status.

eg don't do this request each time through the loop
https://github.com/astropy/astroquery/blob/main/astroquery/astrometry_net/core.py#L250

Thanks,
dustin

dstndstn:
Only query for job id once, rather than each time through the job-status polling loop

closes #3078

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,8 @@ astrometry_net
 
 - Remove photutils from Astroquery astrometry.net [#3067]
 
+- Reduce the number of API calls when polling for job status [#3078]
+
 alma
 ^^^^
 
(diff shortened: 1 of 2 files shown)


dstndstn:
For #3078

keflavich (approved):
lgtm

bsipocz (approved):
Thanks @dstndstn!

PR #3081, opened 2024-08-09
Reduced download volume for ESASky JWST tests

imbasimba:
As discovered in https://github.com/astropy/astroquery/issues/3043, JWST has released hundreds of new images in the area we previously used to test the JWST download functionality. I've now changed the tests to download JWST images from an area that currently only has a few observations.


bsipocz (approved):
Changes here looks good. Thanks!

I'll however remove the changelog entry as a rule of thumb we only do those for user facing changes.

Also, I see a few remote test failures, but all of those are unrelated to these specific changes, and thus are better suited for a follow-up PR.

PR #3096, opened 2024-09-20
Get list of unique products

snbianco:
Added implementation, a test, and documentation for a new function (`Observations.get_unique_product_list`). Given observations, the function returns a list of associated data products, each with a unique data URI. This is essentially a wrapper to `Observations.get_product_list` that removes duplicate rows before returning.

Because `get_product_list` is async and returns a response that needs to be parsed into a Table, I was unable to add this functionality as a parameter to the original function (since we need the result Table to remove duplicates). It's possible that the `Mast.Caom.Products` service could filter out duplicates at some point, but I think this is the best solution for now.

### DIFF
--- a/astroquery/mast/observations.py
+++ b/astroquery/mast/observations.py
@@ -457,6 +457,8 @@ class ObservationsClass(MastQueryWithLogin):
         Note that obsid is NOT the same as obs_id, and inputting obs_id values will result in
         an error. See column documentation `here <https://masttest.stsci.edu/api/v0/_productsfields.html>`__.
 
+        To return unique data products, use ``Observations.get_unique_product_list``.
+
         Parameters
         ----------
         observations : str or `~astropy.table.Row` or list/Table of same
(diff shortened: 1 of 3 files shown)


snbianco:
@dr-rodriguez @astro-jimig

astroquery/mast/observations.py
```diff
@@ -959,8 +961,32 @@ def _remove_duplicate_products(self, data_products):
...
 
+        Returns
+        -------
+        unique_products : `~astropy.table.Table`
+            Table containing products with unique dataURIs.
+        """
+        products = self.get_product_list(observations)
+        unique_products = self._remove_duplicate_products(products)
+        if len(unique_products) < len(products):
+            log.info("To return all products, use `Observations.get_product_list`")
```
astrojimig, line 989: This log message could be slightly more informative if you include the numbers, for example "{len(products) - len(unique_products)} duplicate products removed; To return all products, use `Observations.get_product_list`"

astrojimig: Ah never mind, the other log message captures that info! You're way ahead of me :)

astrojimig:
This looks good! Thanks @snbianco !

bsipocz (approved):
Looks good, thanks! I would suggest a follow-up docs cleanup PR that would make the links better.

PR #3127, opened 2024-11-07
Panstarrs Bugfix

snbianco:
I mentioned in my last MR that we were having an intermittent issue with Panstarrs queries. This is a workaround for robustness so that query functions do not fail when responses have unexpected data. To get the name of the column, we now use the `column_name` key if it exists.

I also added SKIP directives to the code snippets in the data README so that extra files are not left behind when running the test suite.


snbianco:
@dr-rodriguez @astrojimig

PR #3149, opened 2024-12-13
fix: query objects

ManonMarchand:
Hi astroquery maintainers :slightly_smiling_face: 

This PR fixes a rare corner case where the `query_objects` method would not work if the people added the votable field `ident` beforehand.

## MRE:

```python
simbad = Simbad()
simbad.add_votable_fields("ident")
result = simbad.query_objects(['HD 1'])
```
It's solved by adding the possibility to give aliases to joined tables and using and alias for one of the `ident` tables in this specific case. I don't think an alias is needed anywhere else. Here the need for one of the ident to be added with a  LEFT JOIN and the one from add_votable_fields should be added with a normal JOIN is very specific to this method.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -149,6 +149,9 @@ simbad
   A helper method was added ``astroquery.simbad.utils.CriteriaTranslator`` to
   translate between the sim-script syntax and the TAP/ADQL syntax. [#2954]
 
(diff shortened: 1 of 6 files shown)


keflavich (approved):

bsipocz (approved):
Thanks!

PR #3164, opened 2024-12-23
Handle non-ICRS coordinates in query functions

snbianco:


astroquery/mast/utils.py
```diff
@@ -121,6 +121,32 @@ def resolve_object(objectname):
     return coordinates
 
 
+def parse_coordinates(coordinates):
```
bsipocz, line 124: I wonder if this should be a kwarg argument for `commons.parse_coordinates` instead? e.g. to convert right at the end before returning the current output

bsipocz (commented):
I wonder if this should live in the commons function, or you could use what we already have for simbad (_parse_coordinate_and_convert_to_icrs)? 
(or also move the simbad one to commons? cc @ManonMarchand )

_parse_coordinate_and_convert_to_icrs

ManonMarchand:
Oh that methods only parses coordinates if I detect that they are written as a sim-script string. This is an old way of writing coordinates that only existed in the former Simbad interface (hence this is in Simbad's utils and not in commons). 
In the other cases I use parse_coordinates from commons and then convert the SkyCoord it returns to icrs 🙂

astroquery/mast/utils.py
snbianco: Do you mean a parameter like `return_frame` or something to that effect that you could pass in a value like 'ICRS' and get the coordinate back in whatever frame was given?

snbianco:
@bsipocz Pinging this PR again in case it was lost in the shuffle somewhere! I'll be offline from 3/26 - 4/3, but this isn't high priority, so no worries if you don't get around to it before then.

astroquery/utils/commons.py
```diff
@@ -40,7 +40,7 @@
 ASTROPY_LT_7_1_1 = not minversion('astropy', '7.1.1')
 
 
-def parse_coordinates(coordinates):
+def parse_coordinates(coordinates, return_frame=None):
```
bsipocz: make this mandatory kwarg please
```suggestion
def parse_coordinates(coordinates, *, return_frame=None):
```

astroquery/utils/commons.py
```diff
@@ -93,6 +97,14 @@ def parse_coordinates(coordinates):
             c = SkyCoord(coordinates)
     else:
         raise TypeError("Argument cannot be parsed as a coordinate")
+
+    # Convert to requested frame, if given
+    if return_frame and c.frame.name != return_frame.lower():
+        try:
+            c = c.transform_to(return_frame)
+        except ValueError:
+            warnings.warn(f"Failed to transform coordinates to requested frame: {return_frame}.", InputWarning)
```
bsipocz: I feel should let the exception propagate, after all this will only be triggered when return frame is explicitly asked for, thus a problem should be surfaced up to the user/module level

CHANGES.rst
```diff
@@ -239,6 +239,8 @@ Infrastructure, Utility and Other Changes and Additions
 - Removed usage of the astropy TestRunner, therefore the unadvertised
   ``astroquery.test()`` functionality. [#3215]
 
+- ``return_frame`` parameter in ``utils.commons.parse_coordinates`` returns coordinates in the specified frame. [#3164]
```
bsipocz: please rebase to remove the merge commit and to squash out the changelog iterations, and to put this into the 0.4.11 section. Thanks!

snbianco:
Updated and rebased!

CHANGES.rst
```diff
@@ -14,6 +14,8 @@ gaia
 
 - Deprecated ``band`` from ``load_data`` as it has no effect on upstream
   response any more. [#3278]
+- New datalink retrieve types EPOCH_PHOTOMETRY_CROWDED_FIELD, EPOCH_ASTROMETRY_BRIGHT, XP_MEAN_SPECTRUM_GRAVLENS,
```
bsipocz: make sure these are not included in the rebased version

bsipocz (commented):
Unfortunately with the most recent mast merges this is now conflicting again. Could you please do a rebase again and make sure that the only diff in the changelog is the line for this PR?

Code looks good, so I'll merge once the rebase is done and it passes the remote tests.

Thanks.

snbianco:
Rebased! For some reason, the last line in `CHANGES.rst` is still showing up in the diff. I'm not sure how to fix this since there doesn't seem to be any difference between the two versions.

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -416,12 +416,6 @@ def test_mast_service_request(self):
         # Is result limited to ten rows
         assert len(result) == 10
 
-        # Are the GALEX observations in the results table
-        assert "GALEX" in result['obs_collection']
-
-        # Are the two GALEX observations with obs_id 6374399093149532160 in the results table
-        assert len(result[np.where(result["obs_id"] == "6374399093149532160")]) == 2
-
```
snbianco, line 424: Removing these assertions as they occasionally break when the number of results increases.

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -416,12 +416,6 @@ def test_mast_service_request(self):
         # Is result limited to ten rows
         assert len(result) == 10
 
-        # Are the GALEX observations in the results table
-        assert "GALEX" in result['obs_collection']
```
bsipocz, line 420: Why did you remove the GALEX assert?

snbianco: Because we're only fetching one page and not the entire list of results, it's possible that added observations will push the GALEX observations onto the next page and this test will break. My other comment was meant to apply to both these assertions, but I think I only selected one of them, my bad!

bsipocz: OK fair enough. Thanks for the reasoning!

PR #3175, opened 2025-01-14
feat: add query_hierarchy

ManonMarchand:
Hello astroquery maintainers :slightly_smiling_face: !

In the former API, we had two fields `membership` and `link_bibcode` that would hint whether the object had parents. I did not replace those in the rewrite from #2954 as I didn't find them easy to use, and they did not have a direct equivalent in the relational schema.

This PR, adds a new query method `query_hierarchy` to replace them. The behavior is based on what people can see on the webpages in the hierarchy section. You can either ask for parents, children, or siblings of the object.

Ex on NGC4038's page:
image

The two fields from before (`membership` and `link_bibcode`) are only returned if the users ask for them with the argument `detailed_hierarchy`. This is toggled off by default as it has a tendency of duplicating the results **a lot** (see the example in the new documentation section).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -176,6 +176,9 @@ simbad
 - fixed ``query_objects`` that would not work in combination with the additional field
   ``ident`` [#3149]
 
+- added ``query_hierarchy``: a new method that allows to get the parents, children, or
+  siblings of an object [#3175]
+
 skyview
 ^^^^^^^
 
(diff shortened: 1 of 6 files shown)


keflavich (approved):
This looks great!  I read through the code and docs, and it looks solid. This is a neat addition!

ManonMarchand:
Thanks! I amended my commit to add `query_hierarchy` to the list of methods affected by `add_votable_fields` and to the list of methods with the criteria argument in the docs

bsipocz (approved):
Thanks @ManonMarchand!

PR #3194, opened 2025-01-29
Make the API more flexible (allow to ommit `vizier:` when crossmatching with a vizier catalog)

ManonMarchand:
While trying to make the error message clearer as suggested in comment https://github.com/astropy/astroquery/issues/3191#issuecomment-2621412051 I figured that we can also add the `vizier:` if the user did not write it. We already test if the table is available anyway, so we have the information.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,12 @@ ipac.nexsci.nasa_exoplanet_archive
 
 - Fixed InvalidTableError for DI_STARS_EXEP and TD tables. [#3189]
 
+xmatch
+^^^^^^
+
+- the API is more flexible: you can now ommit the ``vizier:`` before the catalog name
+  when crossmatching with a vizier table [#3194]
+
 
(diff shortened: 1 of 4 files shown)


ManonMarchand:
Arg windows and their strange return lines. I'm checking how to fix the test.

keflavich (approved):
Looks great.

The documentation actually does include `vizier:`, but I just glazed right over it when reading the docs.

We could do with some more examples to show, e.g., cone search with Vizier catalogs, but that can go in a different PR (just reminding you here since you noted it in #3191)

ManonMarchand:
Yes, I plan a PR on documentation in the next months, but it's a bit longer to write than fixing the error message/ making this small quality of life improvement. Let's keep https://github.com/astropy/astroquery/issues/3191 open :slightly_smiling_face:

bsipocz (approved):
Thank you!

PR #3216, opened 2025-02-20
Euclid: new euclid module in astroquery

cosmoJFH:
Dear Astroquery team,

This PR aims to integrate a new module for the ESA Euclid Archive. We have developed a fully functional module that gives access to different environments. Please let us know your thoughts and we will iterate on this to merge it. Thanks for your support!

Kind regards,

cc @esdc-esac-esa-int

Jira: EUCLIDPCR-1863


docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+
+To load only table names metadata (TAP+ capability):
+
+.. doctest-remote-data::
+
+  >>> from astroquery.esa.euclid import Euclid
+  >>> tables = Euclid.load_tables(only_names=True, include_shared_tables=True)
+  INFO: Retrieving tables... [astroquery.utils.tap.core]
+  INFO: Parsing tables... [astroquery.utils.tap.core]
+  INFO: Done. [astroquery.utils.tap.core]
```
keflavich, line 159: Is there a way to turn this verbose text off?  If so, please include instructions here about how to do so; users frequently ask for help suppressing messages.

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+
+This query performs a cone search centered at the specified ra/dec coordinates with the provided radius argument.
+
+.. doctest-remote-data::
+
+  >>> #example cone search for source NGC6505
+  >>> from astroquery.esa.euclid import Euclid
+  >>> coord = SkyCoord("17h51m07.4s +65d31m50.8s", frame='icrs')
+  >>> radius = u.Quantity(0.5, u.deg)
+  >>> job = Euclid.cone_search(coordinate=coord, radius=radius, table_name="sedm.mosaic_product", ra_column_name="ra", dec_column_name="dec", columns="*", async_job=True)
```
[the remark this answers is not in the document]
keflavich: also, this is clearly `query_region_async`, since `job.get_results()` must be called afterward.

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+   102158889 2024-10-26T13:10:32.453 267.3807789 65.4983    EUC_MER_BGSUB-MOSAIC-NIR-H_TILE102158889-ED035A_20241024T212936.705156Z_00.00.fits    /euclid/repository_idr/iqr1/Q1_R1/MER/102158889/NISP    /data/euclid_q1/Q1_R1/MER/102158889/NISP       NIR_H 0.16895922479034217
+
+
+1.3. Query object
+^^^^^^^^^^^^^^^^^
+
+This query searches for all the objects contained in an arbitrary rectangular projection of the sky.
+
+WARNING: This method implements the ADQL BOX function that is deprecated in the latest version of the standard
+(ADQL 2.1,  see: https://ivoa.net/documents/ADQL/20231107/PR-ADQL-2.1-20231107.html#tth_sEc4.2.9).
```
keflavich, line 283: I'm just confused about this.  Why do they deprecate the box function? did they replace it with something comparable?

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+
+The following example searches for all the sources contained in an squared region of side = 0.1 degrees around an specific point in ra/dec coordinates. The results are sorted by distance (``dist``) in ascending order.
```
keflavich, line 285: I would also call this search _query_region`.  An object should be an object name, not a coordinate.

I'd definitely prefer that we combine the current `query_object` and `cone_search` into a single `query_region` to match the existing API.  Then, there are two shapes users can provide: the rectangular box shape (width & height) and a circular shape (cone).  Vizier does this: https://github.com/astropy/astroquery/blob/819280102e2de055ae4680aa151290dc1ef8e4fc/astroquery/vizier/core.py#L409-L413

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+    0.03960602180308949            4826.7998046875                       --                         --                        --               --             7826.669921875 ...                                       --                                           --       --             1     0.6376287937164307            --       1
+   0.039606556762811496            4826.7998046875                       --                         --                        --               --             7826.669921875 ...                                       --                                           --       --             0   0.012528697960078716            --       1
+    0.03962541836711639            4826.7998046875                       --                         --                        --               --             7826.669921875 ...                                       --                                           --       --             0   0.003399776993319392            --       1
+  Length = 2000 rows
+
+
+Synchronous queries like this one return a limited number of rows -> 2000
+
+The previous query can be executed as an asynchronous version:
```
keflavich, line 320: So how does one exceed the 2000-row limit with async queries?  The example below also stops at 2000 rows.

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+   0.039606556762811496            4826.7998046875                       --                         --                        --               --             7826.669921875 ...                                       --                                           --       --             0   0.012528697960078716            --       1
+    0.03962541836711639            4826.7998046875                       --                         --                        --               --             7826.669921875 ...                                       --                                           --       --             0   0.003399776993319392            --       1
+  Length = 2000 rows
+
+
+1.4. Synchronous query
+^^^^^^^^^^^^^^^^^^^^^^
+
+The results of a synchronous query are stored at the user side (i.e., they are not saved in the server). These queries
+must be used when the amount of data to be retrieved (number of rows) is small, otherwise, a timeout error can be raised
```
keflavich, line 356:
```suggestion
can only be used when the amount of data to be retrieved (number of rows) is small, otherwise, a timeout error can be raised
```

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+(an error created because the execution time of the query exceeds time execution limit). The output of the synchronous
```
keflavich, line 357: We can assume that astroquery users know what a timeout is, I think.  At least, we don't clarify this in any other modules
```suggestion
. The output of the synchronous
```

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+queries is limited to 2000 rows. If you need more than that, you must use asynchronous queries. The results of
+asynchronous queries can be saved in memory (default) or in a file but is also stored on the server/archive so you can
+access the results by logging in on the archive website too.
+
+Query without saving results in a file:
+
+.. Skipping authentication requiring examples
+.. doctest-skip::
+
+  >>> query content: getting some galaxies from the mer catalogue
```
keflavich, line 367: something is wrong with this line - it's not python code.  Was this meant to be commented?

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+  Phase: COMPLETED
+  Owner: None
+  Output file: 1
+
+Note: to obtain the current location, type:
+
+.. doctest-skip::
+
+  >>> import os
+  >>> print(os.getcwd())
```
keflavich, line 449: This is documentation of python, not astroquery; it is not necessary
```suggestion

```

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Asynchronous queries save all results at server side which means they take up the user's file quota on the archive.
+These queries can be accessed at any time on the archive website. For anonymous users, results are kept for three days.
+For authenticated users the asynchronous results are kept at server side forever (until the user decides to remove them).
+Make sure to delete the results you don't need anymore every once in a while to make sure you don't reach the archive
+user quota. When that happens all future jobs/queries will start failing. You can delete the jobs on the archive website
+(and soon through python too).
+
+Queries retrieved results can be stored locally in memory (by default) or in a file.
```
keflavich, line 462: I'm not sure what this means; could you revise please?

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+          b   3.0   4.0
+          c   5.0   6.0
+
+Note: to obtain the current location, type:
+
+.. doctest-skip::
+
+  >>> import os
+  >>> print(os.getcwd())
+  /Current/directory/path
```
keflavich, line 551: not needed
```suggestion

```

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+To remove asynchronous
+
+.. Pseudo code with ``job_id`` list ellipsis, skipping test
+.. doctest-skip::
+
+  >>> from astroquery.esa.jwst import Jwst
+  >>> job = Jwst.remove_jobs(["job_id_1", "job_id_2", ...])
+
+
+1.6. Synchronous query on an 'on-the-fly' uploaded table
```
keflavich, line 522: what does 'on-the-fly' mean here?  Is this supposed to be related to the environments above?

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+To download the products for a given EUCLID observation_id (observations) or tile_index (mosaics):
+
+.. Skipping authentication requiring examples
+.. doctest-skip::
+
+  >>> #downloading all products for observation id: 102018211
+  >>> mos_id = 1399
+  >>> path = Euclid.get_obs_products(id=mos_id, product_type='mosaic', filter="VIS", output_file=f"{output_folder}/products_{mos_id}.fits", verbose=True)
+
+For big files the download may require a long time.
```
keflavich, line 712: This is hopefully obvious to users; maybe add something about what a typical Euclid image size is?

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+  >>> full_qualified_table_name = 'user_<your_login_name>.table_test_from_file'
+  >>> query = 'select * from ' + full_qualified_table_name
+  >>> job = Euclid.launch_job(query=query)
+  >>> results = job.get_results()
+
+
+2.5.3. Uploading table from an astropy Table
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A in memory PyTable (See https://wiki.python.org/moin/PyTables) can be uploaded to the user private area.
```
keflavich, line 865: I get a 503 error going to that link.  The example below uses astropy tables.  Can you clarify what's going on here?

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+  Retrieving table 'user_joe.table'
+  Parsing table 'user_joe.table'...
+  Done.
+  Table 'user_joe.table' updated.
+
+
+2.8. Cross match
+^^^^^^^^^^^^^^^^
+
+It is possible to run a geometric cross-match between the rA/dec coordinates of two tables
```
keflavich, line 1000:
```suggestion
It is possible to run a geometric cross-match between the RA/Dec coordinates of two tables
```

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+using the crossmatch function provided by the archive. In order to do so the user must be
+logged in. This is required because the cross match operation will generate a join table
+in the user private area. That table contains the identifiers of both tables and the separation,
+in degrees, between rA/dec coordinates of each source in the first table and its associated
```
keflavich, line 1004:
```suggestion
in degrees, between RA/Dec coordinates of each source in the first table and its associated
```

cosmoJFH: Updated to ra/dec

docs/esa/euclid/euclid.rst
cosmoJFH: Updated to ra/dec

docs/esa/euclid/euclid.rst
cosmoJFH: The link works fine for me:

image

docs/esa/euclid/euclid.rst
cosmoJFH: The size depends on the kind of image. I mean it is not easy to give an aprox. value.

docs/esa/euclid/euclid.rst
cosmoJFH: This is explained in the following sentence:

"A votable can be uploaded to the server in order to be used in a query."

docs/esa/euclid/euclid.rst
cosmoJFH: Sorry but making this changes it is not an decision that we could not do in a near future, since these names have been selected by the people that belong to Euclid consortium.

docs/esa/euclid/euclid.rst
cosmoJFH: We only tried to give a warning related to the a change in the ADQL standard. That means that this could be changed in the future.

docs/esa/euclid/euclid.rst
cosmoJFH: Those lines are printed by the logger. After reading https://docs.astropy.org/en/stable/logging.html

we though that the log message could be changed by setting the log level:

from astropy import log
log.setLevel('ERROR')
from astroquery.esa.euclid import Euclid
....

but that did not change the log messages. So, maybe we did not understand the information or the previous commands do not property work.

On the other hand we checked that the command

log.disable_color()

produced the expected change in the color of the log message.

docs/esa/euclid/euclid.rst
```diff
@@ -0,0 +1,1215 @@
...
+  >>> Euclid.get_status_messages()
+
+This method will retrieve the same warning messages shown in EUCLID Science Archive with information about service degradation.
+
+
+---------------------------
+0. Environments
+---------------------------
+
+The Euclid Science Archive has several environments serving different purposes for the Euclid Consortium members.
```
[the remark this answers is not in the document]
cosmoJFH: We have updated the title to 

0. Euclid science archive systems

Each "environment" is associated to a different Euclid archive. This term is commonly used in the project.

docs/esa/euclid/euclid.rst
bsipocz, line 1: I see a lot of test failures in this file, mostly undefined variables, but also indent errors. Please revise them.

(you need to run the tests with remote-data enabled, CI doesn't do that for default:
`pytest docs/esa/euclid/euclid.rst -R` is the command.

bsipocz (changes requested):
Thank you for this module. It's exciting to see that we'll have multiple ways to access Euclid data from day one (well, that's my plan to release this by DR1)

First round of review. I was focusing on the code itself, and identified a lot of potential for cleanps and some places where the API could be made more consistent with the rest of the package. I don't think any of these items require significant time to fix.

The bigger item though is to add tests that are querying the server. At the moment it's only the examples in the documentation doing it. So please add a separate test file with tests that are running against the actual remote service.

And lastly, many of the examples in the documentation are failing the tests. Those will need to be fixed. So far as I see most of these issues are code level ones, e.g. a missing import or variable definition rather than doctest result related. 

What I haven't done in this round: I didn't read the documentation, nor I tried to use the module, but I feel it's better to submit the existing review items sooner rather than later, and Adam already did a read through of the documentation.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+"""
+===============
+Euclid TAP plus
+===============
+@author: Juan Carlos Segovia (original code from Gaia) / John Hoar (Euclid adaptation)
+@contact: juan.carlos.segovia@sciops.esa.int / jhoar@sciops.esa.int
+European Space Astronomy Centre (ESAC)
+European Space Agency (ESA)
+Created on 30 jun. 2016
+Euclid adaptation: 31 January 2022
```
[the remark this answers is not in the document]
cosmoJFH: The redundant information has been removed as you suggested.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+
+
+class EuclidClass(TapPlus):
+    REQUESTED_OBSERVATION_ID = "Missing required argument: 'observation_id'"
+    REQUESTED_TILE_ID = "Missing required argument: 'tile_id'"
+    REQUESTED_OBSERVATION_ID_AND_TILE_ID = "Incompatible: 'observation_id' and 'tile_id'. Use only one."
+    REQUESTED_PRODUCT_TYPE = "Missing required argument: 'product_type'"
+    REQUESTED_GENERIC = "Missing required argument"
+    REQUESTED_RADIUS = "Radius cannot be greater than 30 arcminutes"
+    EUCLID_MESSAGES = "notification?action=GetNotifications"
```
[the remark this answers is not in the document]
cosmoJFH: The fields have been renamed:

```
    ERROR_MSG_REQUESTED_OBSERVATION_ID = "Missing required argument: 'observation_id'"
    ERROR_MSG_REQUESTED_TILE_ID = "Missing required argument: 'tile_id'"
    ERROR_MSG_REQUESTED_OBSERVATION_ID_AND_TILE_ID = "Incompatible: 'observation_id' and 'tile_id'. Use only one."
    ERROR_MSG_REQUESTED_PRODUCT_TYPE = "Missing required argument: 'product_type'"
    ERROR_MSG_REQUESTED_GENERIC = "Missing required argument"
    ERROR_MSG_REQUESTED_RADIUS = "Radius cannot be greater than 30 arcminutes"
```

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+                                          datalink_context="datalink",
+                                          verbose=verbose,
+                                          use_names_over_ids=conf.USE_NAMES_OVER_IDS)
+        else:
+            self.__euclidcutout = cutout_handler
+
+        if show_server_messages:
+            self.get_status_messages()
+
+    def load_tables(self, *, only_names=False, include_shared_tables=False, verbose=False):
```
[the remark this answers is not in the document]
cosmoJFH: I fully agree. Done.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+        verbose : bool, optional, default 'False'
+            flag to display information about the process
+
+        Returns
+        -------
+        A list of table objects
+        """
+        return super().load_tables(only_names=only_names, include_shared_tables=include_shared_tables, verbose=verbose)
+
+    def load_table(self, table, *, verbose=False):
```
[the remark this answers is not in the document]
cosmoJFH: I fully agree. Done.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+            raise ValueError(
+                str(msg) + " must be either a string or astropy.coordinates")
+        if isinstance(value, str):
+            c = commons.parse_coordinates(value)
+            return c
+        else:
+            return value
+
+    @staticmethod
+    def __handle_http_error(error):
```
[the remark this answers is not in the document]
cosmoJFH: The static method is used by other methods in the EuclidClass class, so we would rather keep it.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+                tar_ref.extractall(path=output_dir)
+        elif zipfile.is_zipfile(output_file_full_path):
+            with zipfile.ZipFile(output_file_full_path, 'r') as zip_ref:
+                zip_ref.extractall(output_dir)
+        elif not EuclidClass.is_gz_file(output_file_full_path):
+            # single file: return it
+            files.append(output_file_full_path)
+            return files
+
+    def get_obs_products(self, *, id=None, schema="sedm", product_type=None, product_subtype="STK", filter="VIS",
```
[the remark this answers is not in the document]
cosmoJFH: The method was renamed to `get_observation_products`

docs/esa/euclid/euclid.rst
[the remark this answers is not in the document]
bsipocz: I think we need to drop the duplication of methods and use a async_job kwarg instead like in some of the modules. There is an issue about it already.

So, I'm a strong 👎 on `query_region_async`, but indeed we need to call it query_region instead of `cone_search` and rename the rest of the cone_search methods in the other modules.

Whether `query_region` or `cone_search` is the better term is debatable, but for historical reasons we're stuck with `query_region` and thus it would be great to make the API consistent.

astroquery/esa/euclid/core.py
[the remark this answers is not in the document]
bsipocz: yes, but do we really need to handle the HTTP errors? They are quite standard, so just just wonder why not just raise it as without overriding. 

(but it's not a big deal if this stays)

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+        self.__extract_file(output_file_full_path=output_file_full_path, output_dir=output_dir, files=files)
+        if files:
+            return files
+
+        self.__check_file_number(output_dir=output_dir, output_file_name=os.path.basename(output_file_full_path),
+                                 output_file_full_path=output_file_full_path, files=files)
+
+        return files
+
+    def __get_tile_catalogue_list(self, *, tile_index=None, product_type=None, verbose=False):
```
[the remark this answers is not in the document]
cosmoJFH: We changed the signature of the method, so the arguments are mandatory,

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+                f"mer_segmentation_map.product_type = '{product_type}';")
+
+        if query is None:
+            raise ValueError(f"Invalid product type {product_type}.")
+
+        job = super().launch_job(query=query, output_format='votable_plain', verbose=verbose,
+                                 format_with_results_compressed=('votable_gzip',))
+        return job.get_results()
+
+    def get_product_list(self, *, observation_id=None, tile_index=None, product_type=None, verbose=False):
```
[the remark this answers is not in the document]
cosmoJFH: We changed the signature of the method to

`def get_product_list(self, *, observation_id=None, tile_index=None, product_type, verbose=False):`

so only the `product_type` is mandatory. Note that user can use observation_id or tile_index.

docs/esa/euclid/euclid.rst
keflavich: Huh, this might indicate an upstream error?  Or maybe somehow the wrong logger is being used?

@bsipocz do you know more about this?

(I haven't tested)

docs/esa/euclid/euclid.rst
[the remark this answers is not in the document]
keflavich: fwiw, I'm very firmly on `query_region` >> `cone_search` because `query_region` can do non-cone-searches.  

Otherwise, I'm fine with using a kwarg rather than a function, but the synchronous version must be accessible via this function and should be the default.  The asynchronous version should be an option.

docs/esa/euclid/euclid.rst
keflavich: sure sure, I'm not suggesting any change is needed here, I just don't understand why `box` would be deprecated.  This is a conversation for another thread, but I'd be interested if you know.  Don't consider this question part of the code review.

docs/esa/euclid/euclid.rst
keflavich: The astroquery API is designed to be uniform across modules as much as possible.  The `query_region` method in all other modules performs either a cone search or a box search.  Therefore, what you have described here is a `query_region` method.

`query_object` is a method that takes an object name and passes it through a name resolver, then performs a region search around that.  It's fine if that is the intent here, but then a name resolver should be included and the default parameter should be an object name.

This is a uniform requirement across all modules that use spatial information and is defined in the astroquery paper.

docs/esa/euclid/euclid.rst
keflavich: OK, I understand that explanation, but the words 'on-the-fly' do not seem to apply to this explanation.

docs/esa/euclid/euclid.rst
keflavich: Maybe we can state what the smallest Euclid image is?  Presumably that's one detector footprint?  But, OK, if there's no easy default, let's leave it be.

docs/esa/euclid/euclid.rst
keflavich: Ah, indeed, something was wrong with my browser.

Nevertheless, my comment was actually about the table type used here.  You use astropy tables, not PyTables, so why link to PyTables?

docs/esa/euclid/euclid.rst
keflavich: Following up: I think this is part of a larger issue, https://github.com/astropy/astroquery/issues/3219, so we won't let it hold back merging this one.

docs/esa/euclid/euclid.rst
cosmoJFH: We think the the errors were fixed.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+                                          table_edit_context="TableTool",
+                                          data_context="data",
+                                          datalink_context="datalink",
+                                          connhandler=tap_plus_conn_handler,
+                                          verbose=verbose,
+                                          client_id='ASTROQUERY',
+                                          use_names_over_ids=conf.USE_NAMES_OVER_IDS)
+
+        if datalink_handler is None:
+            self.__eucliddata = TapPlus(url=url_server,
```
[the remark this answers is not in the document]
cosmoJFH: I fully agree with you. Done.

astroquery/esa/euclid/core.py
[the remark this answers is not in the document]
cosmoJFH: We removed the method __handle_http_error

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+        except HTTPError as err:
+            message = EuclidClass.__handle_http_error(err)
+            if message is None:
+                log.error('Query failed: %s, %s' % (query, str(err)))
+            else:
+                log.error(f'Query failed: {query}: \n {message}')
+        except Exception as exx:
+            log.error('Query failed: %s, %s' % (query, str(exx)))
+
+    def load_async_job(self, jobid=None, name=None, verbose=False):
```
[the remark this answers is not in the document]
cosmoJFH: The method was removed.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+        except HTTPError as err:
+            message = EuclidClass.__handle_http_error(err)
+            if message is None:
+                log.error('Loading async job failed: %s, %s' % (jobid, str(err)))
+            else:
+                log.error(f'Loading async job {jobid} failed: \n {message}')
+        except Exception as exx:
+            log.error('Loading async job failed: %s, %s' % (jobid, str(exx)))
+
+    def search_async_jobs(self, *, jobfilter=None, verbose=False):
```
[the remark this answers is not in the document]
cosmoJFH: The method was removed.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+        except HTTPError as err:
+            message = EuclidClass.__handle_http_error(err)
+            if message is None:
+                log.error('Loading jobs with filter failed: %s' % (str(err)))
+            else:
+                log.error(f'Loading jobs with filter failed: \n {message}')
+        except Exception as exx:
+            log.error('Loading jobs with filter failed: %s' % (str(exx)))
+
+    def list_async_jobs(self, *, verbose=False):
```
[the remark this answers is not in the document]
cosmoJFH: The method was removed.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+        columns: list, optional, default ()
+            if empty, all columns will be selected
+        verbose : bool, optional, default 'False'
+            flag to display information about the process
+
+        Returns
+        -------
+        The job results (astropy.table)
+        """
+        return self.__query_object(coordinate, radius=radius, width=width, height=height, async_job=False,
```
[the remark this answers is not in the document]
cosmoJFH: The method was removed.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+                                   verbose=verbose, columns=columns)
+
+    def query_object_async(self, coordinate, *, radius=None, width=None, height=None, verbose=False, columns=()):
```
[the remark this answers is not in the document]
cosmoJFH: The method was removed.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+
+        try:
+            self.__euclidcutout.logout(verbose=verbose)
+            log.info("Euclid cutout server logout OK")
+        except HTTPError as err:
+            log.error('Error logging out cutout server: %s' % (str(err)))
+
+    def __check_quantity_input(self, value, msg):
+        if not (isinstance(value, str) or isinstance(value, units.Quantity)):
+            raise ValueError(f"{msg} must be either a string or astropy.coordinates")
```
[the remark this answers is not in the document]
cosmoJFH: The method __check_quantity_input was removed, since it was not used.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+            log.error('Error logging out data server: %s' % (str(err)))
+            log.error("Error logging out data server")
+
+        try:
+            self.__euclidcutout.logout(verbose=verbose)
+            log.info("Euclid cutout server logout OK")
+        except HTTPError as err:
+            log.error('Error logging out cutout server: %s' % (str(err)))
+
+    def __check_quantity_input(self, value, msg):
```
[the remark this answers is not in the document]
cosmoJFH: The method __check_quantity_input was removed, since it was not used.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+        if value is None:
+            raise ValueError(f"Missing required argument: '{msg}'")
+        if not (isinstance(value, str) or isinstance(value, units.Quantity)):
+            raise ValueError(f"{msg} must be either a string or astropy.coordinates")
+        if isinstance(value, str):
+            return Quantity(value)
+        else:
+            return value
+
+    def __check_coord_input(self, value, msg):
```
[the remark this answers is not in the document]
cosmoJFH: The method __check_coord_input was removed.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+        if not (isinstance(value, str) or isinstance(value, commons.CoordClasses)):
+            raise ValueError(
+                str(msg) + " must be either a string or astropy.coordinates")
+
+    def __get_coord_input(self, value, msg):
```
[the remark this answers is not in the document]
cosmoJFH: The method __get_coord_input was removed.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+                    table_name=None,
+                    ra_column_name=None,
+                    dec_column_name=None,
+                    async_job=False,
+                    background=False,
+                    dump_to_file=False,
+                    output_file=None,
+                    output_format="csv",
+                    verbose=False,
+                    columns=()):
```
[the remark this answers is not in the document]
cosmoJFH: Done.

astroquery/esa/euclid/tests/test_sastap.py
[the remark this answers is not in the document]
cosmoJFH: The file has been renamed to test_euclidtap.py.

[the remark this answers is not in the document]
cosmoJFH: Regarding to the tests that would make use of the actual server, nowadays it can only be accessed by authenticated users. I mean we would need to make use of a particular user/passw to run the tests. 

Could we wait till the service is open to the public in order to design and execute a new set of remote tests?

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+            if empty, all columns will be selected
+
+        Returns
+        -------
+        The job results (astropy.table).
+        """
+        return self.__query_object(coordinate, radius=radius, width=width, height=height, async_job=True,
+                                   verbose=verbose, columns=columns)
+
+    def cone_search(self, coordinate, radius, *,
```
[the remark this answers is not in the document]
cosmoJFH: Pending to confirm this change with the rest of the team.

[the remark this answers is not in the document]
keflavich: Following the conversation here: https://github.com/astropy/astroquery/pull/3216#discussion_r1963682998, I think we are prepared to merge this with the current function names to make sure you guys can meet your target deadlines, but we'll want to refactor later to add the standard user-facing methods.

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1448 @@
...
+from astropy.units import Quantity
+from requests.exceptions import HTTPError
+
+from astroquery import log
+from astroquery.utils import commons
+from astroquery.utils.tap import TapPlus
+from . import conf
+
+
+class EuclidClass(TapPlus):
```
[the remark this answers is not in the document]
cosmoJFH: We followed the same approach as it was done in the gaia module, so we could make use of all the features of the TAP service.

docs/esa/euclid/euclid.rst
cosmoJFH: The method was removed.

docs/esa/euclid/euclid.rst
cosmoJFH: You can exceed  the row limit if the query that is executed under the hood returns more than 2000 rows. This is an example for this scenario.

docs/esa/euclid/euclid.rst
cosmoJFH: Done.

docs/esa/euclid/euclid.rst
cosmoJFH: Done

docs/esa/euclid/euclid.rst
cosmoJFH: Fixed.

docs/esa/euclid/euclid.rst
cosmoJFH: We would like to keep it. Similar information is included in other modules, for example in jwst.

docs/esa/euclid/euclid.rst
cosmoJFH: We changed the text to 

```
The results of the execution of the queries can be stored locally in memory (by default) or in a file. The following 
command stores the result in memory:
```

docs/esa/euclid/euclid.rst
cosmoJFH: We would like to keep it. Similar information is included in other modules, for example in jwst.

docs/esa/euclid/euclid.rst
cosmoJFH: From my point of view, 'on-the-fly' means that it is not necessary to upload a table to the archive in advance to be used by the sql command that is launched by the command.

But taking into account that I am not a native  English speaker I am open to any suggestion. Note that the same term is used in the jwst and gaia modules.

astroquery/esa/euclid/tests/test_sastap.py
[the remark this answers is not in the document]
bsipocz: > Could we wait till the service is open to the public in order to design and execute a new set of remote tests?

Yes, that sounds more than reasonable. Here is a follow-up issue for it: https://github.com/astropy/astroquery/issues/3221

docs/esa/euclid/euclid.rst
bsipocz: I still see some failures, some of this may be due to I'm not logged in:

```
==================================================================== FAILURES =====================================================================
______________________________________________________________ [doctest] euclid.rst _______________________________________________________________
151 To load only table names metadata (TAP+ capability):
152 
153 .. doctest-remote-data::
154 
155   >>> from astroquery.esa.euclid import Euclid
156   >>> tables = Euclid.load_tables(only_names=True, include_shared_tables=True)
157   INFO: Retrieving tables... [astroquery.utils.tap.core]
158   INFO: Parsing tables... [astroquery.utils.tap.core]
159   INFO: Done. [astroquery.utils.tap.core]
160   >>> print("Found", len(tables), "tables")
Expected:
    Found 45 tables
Got:
    Found 34 tables

/Users/bsipocz/munka/devel/astroquery/docs/esa/euclid/euclid.rst:160: DocTestFailure
153 .. doctest-remote-data::
154 
155   >>> from astroquery.esa.euclid import Euclid
156   >>> tables = Euclid.load_tables(only_names=True, include_shared_tables=True)
157   INFO: Retrieving tables... [astroquery.utils.tap.core]
158   INFO: Parsing tables... [astroquery.utils.tap.core]
159   INFO: Done. [astroquery.utils.tap.core]
160   >>> print("Found", len(tables), "tables")
161   Found 45 tables
162   >>> for table in tables:
UNEXPECTED EXCEPTION: IndentationError("expected an indented block after 'for' statement on line 1", ('<doctest euclid.rst[8]>', 1, 21, 'for table in tables:\n', 1, -1))
  File "/Users/bsipocz/.pyenv/versions/3.12.1/lib/python3.12/doctest.py", line 1359, in __run
    exec(compile(example.source, filename, "single",
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<doctest euclid.rst[8]>", line 1
    for table in tables:
                        ^
IndentationError: expected an indented block after 'for' statement on line 1
/Users/bsipocz/munka/devel/astroquery/docs/esa/euclid/euclid.rst:162: UnexpectedException
200   Description: None
201   Size (bytes): 0
202   Num. columns: 12
203 
204 
205 Once a table is loaded, its columns can be inspected:
206 
207 .. doctest-remote-data::
208 
209   >>> for column in raw_detector_table.columns:
Differences (unified diff with -expected +actual):
    @@ -11,3 +11,2 @@
     q4_oid
     to_be_published
    -...

/Users/bsipocz/munka/devel/astroquery/docs/esa/euclid/euclid.rst:209: DocTestFailure
============================================================= short test summary info =============================================================
FAILED docs/esa/euclid/euclid.rst::euclid.rst
========================================================== 1 failed, 31 passed in 20.15s ==========================================================

```

astroquery/esa/euclid/core.py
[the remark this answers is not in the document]
bsipocz: Nice. Indeed I didn't check gaia but the other modules under the `esa` directory.

docs/esa/euclid/euclid.rst
cosmoJFH: We think that now it is ready:

```
$ python3 -m pytest docs/esa/euclid/euclid.rst -R 
================================================================================================================================================================== test session starts ===================================================================================================================================================================
platform linux -- Python 3.11.5, pytest-7.4.0, pluggy-1.4.0

Running tests with astroquery version 0.4.10.dev9886_testrun.

Running tests with astropy_helpers version 4.0.1.
Running tests in docs/esa/euclid/euclid.rst.

Date: 2025-02-22T08:45:16

Platform: Linux-6.8.0-52-generic-x86_64-with-glibc2.35

Executable: /home/anaconda3/envs/test_astroquery/bin/python3

Full Python Version: 
3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0]

byteorder: little

Package versions: 
Numpy: 1.26.3
Scipy: not available
Matplotlib: 3.8.2
h5py: not available
Pandas: not available
Astropy: 6.0.0
regions: 0.8
pyVO: 1.6.1
mocpy: 0.13.1
astropy-healpix: 1.0.2
vamdclib: not available

Using Astropy options: remote_data: any.

rootdir: /home//astroquery_EUCLIDPCR-1863
configfile: setup.cfg
plugins: mock-3.10.0, arraydiff-0.6.1, cov-4.1.0, filter-subpackage-0.1.2, hypothesis-6.82.0, anyio-3.5.0, astropy-header-0.2.2, astropy-0.11.0, pep8-1.0.6, openfiles-0.5.0, remotedata-0.4.1, doctestplus-1.1.0
collected 1 item                                                                                                                                                                                                                                                                                                                                         

docs/esa/euclid/euclid.rst .                                                                                                                                                                                                                                                                                                                       [100%]

=================================================================================================================================================================== 1 passed in 6.05s ====================================================================================================================================================================
```

astroquery/esa/euclid/tests/test_sastap.py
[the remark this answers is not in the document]
cosmoJFH: Thanks. We would include that task in our next sprint.

docs/esa/euclid/euclid.rst
keflavich: Do you mean you removed the box search functionality?  I think that's very useful and would recommend keeping it!  We can figure out API regularization later.

docs/esa/euclid/euclid.rst
keflavich: OK, thanks, I wouldn't have interpreted it that way, but I understand the meaning now.  

I can't think of a better term, but I would recommend including an explanation immediately below the title, e.g.,:
```rst
"'On-the-fly' queries allow you to submit a table and perform a query using that table all in one command; unlike tables uploaded with `upload_table` (see :ref:`Uploading table from URL`), these tables will be deleted after the query is complete."
```

cosmoJFH: Thanks for your feedback. 

We included the explanation you wrote.

docs/esa/euclid/euclid.rst
cosmoJFH: Sorry, my previous answer was wrong. That answer corresponds to other question. 

We have contacted the Euclid team regarding #https://github.com/astropy/astroquery/issues/3219, since they should have to update some documents and notebooks.

keflavich: OK, then I'll mark this conversation resolved & we can continue API discussion in #3219

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1205 @@
...
+                                      upload_resource=upload_resource,
+                                      upload_table_name=upload_table_name,
+                                      format_with_results_compressed=('votable_gzip',))
+
+        except HTTPError as err:
+            log.error(f'Query failed: {query}: HTTP error: {err}')
+        except Exception as exx:
+            log.error(f'Query failed: {query}, {str(exx)}')
+
+    def launch_job_async(self, query, *, name=None, dump_to_file=False, output_file=None, output_format="csv",
```
bsipocz, line 145: Not necessarily as part of this PR, but I wonder if we could/should merge the async and sync methods into one and have an `async_job` kwarg instead that switched between the two?

astroquery/esa/euclid/core.py
```diff
@@ -0,0 +1,1205 @@
...
+                                            upload_table_name=upload_table_name,
+                                            autorun=autorun,
+                                            format_with_results_compressed=('votable_gzip',))
+
+        except HTTPError as err:
+            log.error(f'Query failed: {query}: HTTP error: {err}')
+        except Exception as exx:
+            log.error(f'Query failed: {query}, {str(exx)}')
+
+    def query_object(self, coordinate, radius=None, width=None, height=None,
```
bsipocz, line 200: make the parameters mandatory keyword params especially given their usage of either radius or width/height needed to be provided

```suggestion
    def query_object(self, coordinate, *, radius=None, width=None, height=None,
```

bsipocz (approved):
OK, so one minor comment about making some keywords mandatory, and some thinking about the merging of launch_job and launch_job_async, but this latter is a bigger picture question and is more like part of https://github.com/astropy/astroquery/issues/3075

astroquery/esa/euclid/core.py
bsipocz: I went ahead with this one, and then with the merging of the PR, so it's easier to shake out any remaining issues in potential follow-ups.

bsipocz:
One minor overall comment: there are a lot of commits with repeated commit messages. It would be nice to squash these down, or add consecutive ones with ``--amend``, so the commits stay as logical chunks rather than this incremental, especially when it's addressing docs/styling/etc trivialities. I'm going ahead with the merge now though, as is, but please keep the commit history a little bit more concise.

bsipocz:
Quick release containing this PR is on pypi: `https://pypi.org/project/astroquery/0.4.10.dev9924/`

Use `pip install --pre -U astroquery` to install it.

PR #3230, opened 2025-02-27
Add batching to MastMissions.get_product_list

snbianco:
Add batching to ``MastMissions.get_product_list`` to avoid server errors and allow for a larger number of input datasets.

Also added some tests to check for this case, although they don't return any results to avoid performance-intensive queries.


bsipocz (approved):
Thanks!

PR #3238, opened 2025-03-03
GAIA:  Revert change in the name of the datalink RVS type, in the PR #3207

cosmoJFH:
In  PR #3207 we changed the name of the datalink retrieve type RVS by RVS_MEAN_SPECTRUM.  We would like to revert this change. The rests of the changes done in the PR #3207 were right.

cc @esdc-esac-esa-int 

jira:GAIAMNGT-1828)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,7 +34,7 @@ Service fixes and enhancements
 gaia
 ^^^^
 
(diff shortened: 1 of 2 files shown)


bsipocz (approved):
Thanks!

PR #3255, opened 2025-03-15
EUCLID: Override TAP method get_datalinks

cosmoJFH:
Hi Astroquery team,

we need to overload the method TAP get_datalinks, so that the correct TapPlus class associated to the sas-data service is used.

cc @esdc-esac-esa-int 

jira: EUCLIDPCR-1914


keflavich (approved):

bsipocz (approved):
thanks!

PR #3258, opened 2025-03-17
[SIMBAD] Fix upstream changes

ManonMarchand:
SIMBAD upgraded VOLLT today. This includes the fix from https://github.com/gmantele/vollt/commit/407753f875c899840d53f5e0a8acf659efc56dc6 that was expected to break some of the examples (see the comments that are now removed)

The new version also makes SIMBAD's columns names respect the ADQL reserved keywords, this appears in the table `biblio` where the column `year` is now protected by quotation marks.

PS: I'm not sure about weather this needs a changelog entry or not? The changes are not from astroquery and only appear in the docs


keflavich (approved):

bsipocz (approved):
Thanks! No need for changelog for anything test/docs related or things that are only internal refactoring with no user-facing consequence.

PR #3260, opened 2025-03-17
Make criteria in MastMissions queries case-sensitive

snbianco:
When I added parameter checking to `MastMissions`, I incorrectly assumed that criteria were case-insensitive in the underlying API. This PR performs a case-sensitive comparison to check that parameters match available criteria exactly.


bsipocz (approved):
Thank you, this all looks good!

PR #3290, opened 2025-04-14
Log a warning if a cached file's length cannot be verified

snbianco:
See this [discussion](https://github.com/astropy/astroquery/pull/3232#discussion_r2031963533).

Rather than raise an error when a cached file's length cannot be verified, warn the user and re-download the file. This is needed for certain files in the MAST Archive that are generated dynamically and don't include a `Content-Length` header in their response at this time.

My understanding was that this could not be tested in `test_query.py`, but I did add a case in the MAST module tests.


keflavich (approved):
Looks good, thanks for implementing this (and the test!)

bsipocz (approved):
Thanks!

PR #3303, opened 2025-04-28
Use https URL for SVO

lpsinger:


lpsinger:
Note that the pipeline failures are fixed by #3304.

bsipocz (approved):
Thanks!

PR #3304, opened 2025-04-28
MAINT: Replace deprecated isiterable with np.iterable

lpsinger:
`astropy.utils.isiterable` was deprecated in astropy 7.1. The suggested replacement is `np.iterable`.


bsipocz (approved):
Thanks!

PR #3314, opened 2025-05-07
Enhancements for mast.Observations.get_cloud_uris()

snbianco:
- Fix bug in `utils.remove_duplicate_products` that does not retain the order of the products in an input table.
- Added `return_uri_map` parameter to `Observations.get_cloud_uris` to return a mapping of the input data product URIs to the returned cloud URIs.
- Added `verbose` parameter to `Observations.get_cloud_uris` to control whether warnings are logged when a product cannot be found in the cloud.
- Also fixed an issue with some unrelated tests/documentation. The results of the original query changed so that subsequent commands were no longer working as expected (TESS FFI observations were being returned, and we no longer allow users to get product lists for these observations). The updated queries filter out any results from TESS.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -72,6 +72,9 @@ mast
 - Added ``return_uri_map`` parameter to ``Observations.get_cloud_uris`` to return a mapping of the input data product URIs 
   to the returned cloud URIs. [#3314]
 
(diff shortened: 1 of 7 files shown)


bsipocz (approved):
All looks good, thanks @snbianco!

PR #3319, opened 2025-05-15
Accept lists for MastMissions query criteria

snbianco:
Previously, in order to filter by multiple values for a criterion in `MastMissions`, users would have to pass in a comma-delimited string. This PR adds support for a list of values. It also allows users to pass in a single string for the `sort_by` keyword and a single boolean for the `sort_desc` keyword rather than a list.

In addition, this PR does some additional error handling to be more helpful to users when they make invalid queries. If the request comes back with a HTTP 400 or 500 error (both indicative of bad input), an `InvalidQueryError` is raised that advises the user to check their query.


bsipocz (approved):
All looks good. Thanks!

PR #3339, opened 2025-06-05
Addresses Issue #3138 - ESO - Enable Table Access Protocol (TAP) queries

### LINKED ISSUE #3138: ESO - Enable Table Access Protocol (TAP) queries

juanmcloaiza:
Closes #3138

### Changes:

Switch querying interface from WDB to TAP in querying functions:
- [x] query_main
- [x] query_instruments
- [x] query_surveys
- [x] query_apex_quicklooks
- [x] list_instruments
- [x] list_surveys

Deprecate arguments:
- [x] `open_form`
- [x] `cache`

New functionality:
- [x] Allow plain ADQL queries via `query_tap_service` (with authentication as well) 
- [x] Cone search using `cone_ra`, `cone_dec`, `cone_radius` arguments
- [x] Retrieve record count before querying the archive, via `count_only` argument
- [x] Ask query functions to print the underlying ADQL queries without issuing them

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -80,7 +80,7 @@ eso
 ^^^
 
 - Switch querying interface from WDB to TAP in querying functions. [#3339]
-- Allow plain ADQL queries via ``query_tap_service`` (with authentication as well). [#3339]
+- Allow plain ADQL queries via ``query_tap`` (with authentication as well). [#3339]
 - Cone search using ``cone_ra`, ``cone_dec`, ``cone_radius`` arguments. [#3339]
 - Retrieve record count before querying the archive, via ``count_only`` argument. [#3339]
 - Ask query functions to print the underlying ADQL queries without issuing them. [#3339]
--- a/astroquery/eso/core.py
+++ b/astroquery/eso/core.py
@@ -41,6 +41,9 @@ from .utils import _UserParams, raise_if_coords_not_valid, _reorder_columns, \
     DEFAULT_LEAD_COLS_PHASE3, DEFAULT_LEAD_COLS_RAW
 
 
+__all__ = ['Eso', 'EsoClass']
+
+
 class CalSelectorError(Exception):
     """
     Raised on failure to parse CalSelector's response.
@@ -319,7 +322,7 @@ class EsoClass(QueryWithLogin):
         return table_to_return
 
     @unlimited_maxrec
-    @deprecated_renamed_argument('cache', None, since='0.4.11')
+    @deprecated_renamed_argument('cache', None, since='0.4.12')
     def list_instruments(self, cache=True) -> List[str]:
         """
         List all the available instrument-specific queries offered by the ESO archive.
@@ -346,7 +349,7 @@ class EsoClass(QueryWithLogin):
         return l_res
 
     @unlimited_maxrec
-    @deprecated_renamed_argument('cache', None, since='0.4.11')
+    @deprecated_renamed_argument('cache', None, since='0.4.12')
     def list_surveys(self, *, cache=True) -> List[str]:
         """
         List all the available surveys (phase 3) in the ESO archive.
@@ -408,7 +411,7 @@ class EsoClass(QueryWithLogin):
         return list(ret_table[0].values())[0] if user_params.count_only else ret_table
 
     @deprecated_renamed_argument(('open_form', 'cache'), (None, None),
-                                 since=['0.4.11', '0.4.11'])
+                                 since=['0.4.12', '0.4.12'])
     def query_surveys(
             self,
             surveys: Union[List[str], str] = None, *,
@@ -498,7 +501,7 @@ class EsoClass(QueryWithLogin):
         return t
 
     @deprecated_renamed_argument(('open_form', 'cache'), (None, None),
-                                 since=['0.4.11', '0.4.11'])
+                                 since=['0.4.12', '0.4.12'])
     def query_main(
             self,
             instruments: Union[List[str], str] = None, *,
@@ -588,7 +591,7 @@ class EsoClass(QueryWithLogin):
         return t
 
     @deprecated_renamed_argument(('open_form', 'cache'), (None, None),
-                                 since=['0.4.11', '0.4.11'])
+                                 since=['0.4.12', '0.4.12'])
     def query_instrument(
             self,
             instrument: str, *,
@@ -975,7 +978,7 @@ class EsoClass(QueryWithLogin):
         return files[0] if files and len(files) == 1 and return_string else files
 
     @deprecated_renamed_argument(('open_form', 'cache'), (None, None),
-                                 since=['0.4.11', '0.4.11'])
+                                 since=['0.4.12', '0.4.12'])
     def query_apex_quicklooks(self,
                               project_id: Union[List[str], str] = None, *,
                               columns: Union[List, str] = None,


juanmcloaiza:
Hi @bsipocz , could you please give us a time estimate of the review process? No pressure at all though, we only need to plan our activities internally. Thanks a lot in advance.

bsipocz:
@juanmcloaiza - sorry for the delay. I'll try to get to the review this week.

astroquery/eso/core.py
```diff
@@ -1,59 +1,59 @@
...
 import subprocess
 import time
 import warnings
-import webbrowser
 import xml.etree.ElementTree as ET
-from io import BytesIO
-from typing import List, Optional, Tuple, Dict, Set
+from typing import List, Optional, Tuple, Dict, Set, Union
 
+import astropy.table
```
bsipocz: L28 explicitly imports from here, remove this one add to that list as necessary

astroquery/eso/core.py
```diff
@@ -1,59 +1,59 @@
...
 import astropy.utils.data
 import keyring
 import requests.exceptions
 from astropy.table import Table, Column
 from astropy.utils.decorators import deprecated_renamed_argument
 from bs4 import BeautifulSoup
+from pyvo.dal import TAPService
+import pyvo.dal.exceptions as pde
```
bsipocz: can you directly import the classes here? It's quite obvious reading the code that those are coming from pyvo (and this alias is not helping to make it more obvious anyway)

astroquery/eso/core.py
```diff
@@ -1,59 +1,59 @@
...
-    Check the response from an ESO service query for various types of error
 
-    If all is OK, return True
-    """
-    if b"NETWORKPROBLEM" in content:
-        raise RemoteServiceError("The query resulted in a network "
-                                 "problem; the service may be offline.")
-    elif b"# No data returned !" not in content:
-        return True
+__doctest_skip__ = ['EsoClass.*']
```
bsipocz: why do we skip doctesting?

astroquery/eso/core.py
```diff
@@ -1,59 +1,59 @@
...
```
bsipocz, line 46: have an explicit `__all__` with listing the public API

astroquery/eso/core.py
```diff
@@ -62,168 +62,90 @@ def _get_exp_time_from_token(self) -> int:
...
-        self._auth_info: Optional[AuthInfo] = None
+        self._auth_info: Optional[_AuthInfo] = None
+        self._hash = None
+        self._maxrec = None
+        self.maxrec = conf.row_limit
 
-    def _activate_form(self, response, *, form_index=0, form_id=None, inputs={},
-                       cache=True, method=None):
+    @property
+    def maxrec(self):
```
bsipocz: Why does maxrec need to be a class properly rather than a parameter for the query functions? After all it should be specific to individual queries rather than being set on the instance. (See the other modules)

astroquery/eso/core.py
```diff
@@ -62,168 +62,90 @@ def _get_exp_time_from_token(self) -> int:
...
-            response = self._request("POST", url, files=payload, cache=cache)
-        elif fmt == 'application/x-www-form-urlencoded':
-            response = self._request("POST", url, data=payload, cache=cache)
-
-        return response
+            mr = value
+
+        self._maxrec = mr
+
+    def _tap_url(self) -> str:
```
bsipocz, line 145: now, this can be a class property, though currently we have a mixture if it's an attribute inheriting its value from the config or a property (I'm inclined to go with what we have in the IRSA module, but the property approach of alma is OK, too. But preferably don't have a third approach

astroquery/eso/core.py
```diff
@@ -271,8 +193,8 @@ def _get_auth_info(self, username: str, *, store_password: bool = False,
 
         return username, password
 
-    def _login(self, *, username: str = None, store_password: bool = False,
-               reenter_password: bool = False) -> bool:
+    def _login(self, *args, username: str = None, store_password: bool = False,
+               reenter_password: bool = False, **kwargs) -> bool:
```
bsipocz: what args and kwargs are expected here? It would be preferable to spell them out as I assume there won't be subclassing that allows for multiple authentication

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        Issues a warning when a table is empty or when the
+        results are truncated
+        """
+        if len(table) < 1:
+            warnings.warn("Query returned no results", NoResultsWarning)
+
+        if len(table) == self.maxrec:
+            warnings.warn(f"Results truncated to {self.maxrec}. "
+                          "To retrieve all the records set to None the maxrec attribute",
+                          MaxResultsWarning)
```
bsipocz: warnings should come from pyvo for unexpectedly hitting maxrec values, maybe just let those through?

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+
+    def _try_download_pyvo_table(self,
```
bsipocz, line 242: Why is this a function? This is called only once, please do the TAP call there, directly.

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        try:
+            table_to_return = tap.search(query_str, maxrec=self.maxrec).to_table()
+            self._maybe_warn_about_table_length(table_to_return)
+        except pde.DALQueryError:
+            log.error(message(query_str))
+        except pde.DALFormatError as e:
+            raise pde.DALFormatError(message(query_str) + f"cause: {e.cause}") from e
+        except Exception as e:
+            raise RuntimeError(
+                f"Unhandled exception {type(e)}\n" + message(query_str)) from e
```
bsipocz: we usually have the `get_query_payload` argument to return the query string, it maybe useful to add that and rely less on customising the exceptions and swallowing the traceback.

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+
+        return table_to_return
+
+    def _tap_service(self, authenticated: bool = False) -> TAPService:
```
bsipocz: we tend to just have a `tap` property, please see alma/irsa/heasarc for examples.

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        if authenticated:
+            h = self._get_auth_header()
+            self._session.headers = {**self._session.headers, **h}
+            tap_service = TAPService(self._tap_url(), session=self._session)
+        else:
+            tap_service = TAPService(self._tap_url())
+
+        return tap_service
+
+    def query_tap_service(self,
```
bsipocz: use the method name `query_tap` to be consistent with the rest of the API. You may also consider adding maxrec, async_job, getquery_payload as optional arguments.

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+                          query_str: str,
```
bsipocz: To be consistent with the other modules
```suggestion
                          query: str, *
```

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+            The ADQL query string to be executed.
+        authenticated : bool, optional
+            If ``True``, the query is run as an authenticated user.
+            Authentication must be performed beforehand via
+            :meth:`astroquery.eso.EsoClass.login`. Authenticated queries
+            may be slower. Default is ``False``.
+
+        Returns
+        -------
+        astropy.table.Table or None
```
bsipocz: Now, this is nice to return Tables, we are not yet doing that with all the other modules. But consider returning an empty Table instead of `None` for no data response.

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        Examples
+        --------
+        >>> from astroquery.eso import Eso
+        >>> eso_instance = Eso()
+        >>> eso_instance.query_tap_service("SELECT * FROM ivoa.ObsCore")
+        """
+        table_to_return = None
+        tap_service = self._tap_service(authenticated)
+        table_to_return = self._try_download_pyvo_table(query_str, tap_service)
+        return table_to_return
```
bsipocz: please look at the other modules (alma/irsa/heasarc) for examples how how to do this more simply

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
-            Defaults to True. If set overrides global caching behavior.
-            See :ref:`caching documentation <astroquery_cache>`.
+            Deprecated - unused.
+        """
+        _ = cache  # We're aware about disregarding the argument
+        query_str = ("select table_name from TAP_SCHEMA.tables "
+                     "where schema_name='ist' order by table_name")
+        res = self.query_tap_service(query_str)["table_name"].data
+
+        l_res = list(res)
```
bsipocz, line 341: I think it's OK to break the API now that this is practically a new module, and return the Table here. You may also consider returning the full table similarly of what we do for `Irsa.list_catalogs(full=True)`

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        if _EsoNames.apex_quicklooks_table in l_res:
+            l_res.remove(_EsoNames.apex_quicklooks_table)
+
+        l_res = list(map(lambda x: x.split(".")[1], l_res))
 
+        return l_res
+
+    @unlimited_max_rec
+    @deprecated_renamed_argument('cache', None, since='0.4.11')
+    def list_surveys(self, *, cache=True) -> List[str]:
```
bsipocz, line 353: Same here, consider breaking API and return a Table if it makes sense

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
-                      help=False, open_form=False, **kwargs):
+        _ = cache  # We're aware about disregarding the argument
+        t = _EsoNames.phase3_table
+        c = _EsoNames.phase3_surveys_column
+        query_str = f"select distinct {c} from {t}"
+        res = list(self.query_tap_service(query_str)[c].data)
+        return res
+
+    @unlimited_max_rec
+    def _print_table_help(self, table_name: str) -> None:
```
bsipocz: consider having this as a public method as `list_column` (see similar method in irsa)

astroquery/eso/core.py
```diff
@@ -62,168 +62,90 @@ def _get_exp_time_from_token(self) -> int:
...
+        """
+        Returns the name of the instrument specific table (IST)
+        """
+        return f"ist.{instrument_name}"
+
+    apex_quicklooks_table = ist_table.__func__("apex_quicklooks")
+    apex_quicklooks_pid_column = "project_id"
+
+
+def unlimited_max_rec(func):
```
bsipocz: To be consistent with how we spell when passing it upstream
```suggestion
def unlimited_maxrec(func):
```

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        log.info(f"\nColumns present in the table {table_name}:\n{available_cols}\n"
+                 f"\nNumber of records present in the table {table_name}:\n{num_records}\n")
+        astropy.conf.max_lines = n_
+        astropy.conf.max_width = m_
+
+    def _query_on_allowed_values(
+        self,
+        user_params: _UserParams
+    ) -> Union[astropy.table.Table, int, str]:
+        up = user_params  # shorthand
```
bsipocz: please avoid using a shorthand, it makes the code less readable, and we don't need to save on a few characters

astroquery/eso/utils.py
```diff
@@ -0,0 +1,197 @@
...
+    is_a_valid_combination = are_all_none or are_all_float
+    if not is_a_valid_combination:
+        raise ValueError(
+            "Either all three (cone_ra, cone_dec, cone_radius) must be present or none.\n"
+            "Values provided:\n"
+            f"\tcone_ra = {cone_ra}, cone_dec = {cone_dec}, cone_radius = {cone_radius}"
+        )
+
+
+def _py2adql(user_params: _UserParams) -> str:
```
bsipocz: can we call this something more descriptive? `parameters_to_adql` or something similar?

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+            self._print_table_help(up.table_name)
+            return
+
+        _raise_if_has_deprecated_keys(up.column_filters)
+
+        raise_if_coords_not_valid(up.cone_ra, up.cone_dec, up.cone_radius)
+
+        query = _py2adql(up)
+
+        if up.query_str_only:
```
bsipocz: have a `get_query_payload` argument instead of this to be consistent with the rest of the API

astroquery/eso/utils.py
```diff
@@ -0,0 +1,197 @@
...
+
+def _split_str_as_list_of_str(column_str: str):
+    if column_str == '':
+        column_list = []
+    else:
+        column_list = list(map(lambda x: x.strip(), column_str.split(',')))
+    return column_list
+
+
+def _raise_if_has_deprecated_keys(filters: Optional[Dict[str, str]]) -> bool:
```
bsipocz, line 42: No need for this, the deprecation warnings should be raised the standard way

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        n_ = astropy.conf.max_lines
+        m_ = astropy.conf.max_width
+        astropy.conf.max_lines = nlines
+        astropy.conf.max_width = sys.maxsize
+        log.info(f"\nColumns present in the table {table_name}:\n{available_cols}\n"
+                 f"\nNumber of records present in the table {table_name}:\n{num_records}\n")
+        astropy.conf.max_lines = n_
+        astropy.conf.max_width = m_
+
+    def _query_on_allowed_values(
```
bsipocz, line 393: I'm not convinced that this method is necessary, `list_columns` should deal with the help, the deprecations are handled by the deprecation decorator, the coordinates can be ducktyped (or run through a SkyCoord construction) and then the query can be called in the method that calls this one.

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+                         print_help=help,
+                         authenticated=authenticated,
+                         )
+        t = self._query_on_allowed_values(user_params=up)
+        t = reorder_columns(t, DEFAULT_LEAD_COLS_PHASE3)
+        return t
+
+    @deprecated_renamed_argument(('open_form', 'cache'), (None, None),
+                                 since=['0.4.11', '0.4.11'])
+    def query_main(
```
bsipocz, line 505: maybe this is a good time to rename this method to be more descriptive? Maybe `query_instruments`?

Then we can fully deprecate the single instrument query from below, as that would just be a special case for running this query

astroquery/eso/core.py
```diff
@@ -296,248 +218,472 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+                         print_help=help,
+                         authenticated=authenticated,
+                         )
+        t = self._query_on_allowed_values(up)
+        t = reorder_columns(t, DEFAULT_LEAD_COLS_RAW)
+        return t
+
+    @deprecated_renamed_argument(('open_form', 'cache'), (None, None),
+                                 since=['0.4.11', '0.4.11'])
+    def query_instrument(
```
bsipocz, line 595: is there a need for this now as it feels to be just a special case for `query_main` above

astroquery/eso/core.py
```diff
@@ -606,10 +752,10 @@ def get_headers(self, product_ids, *, cache=True):
                     columns.append(key)
                     column_types.append(type(header[key]))
         # Add all missing elements
-        for i in range(len(result)):
+        for r in result:
```
bsipocz: nitpicking, but please try to avoid single letter variables

astroquery/eso/core.py
```diff
@@ -667,14 +813,13 @@ def _download_eso_files(self, file_ids: List[str], destination: Optional[str],
...
-                    log.info(f"Successfully downloaded dataset"
-                             f" {file_id} to {filename}")
+                    log.info(f"Successfully downloaded dataset {file_id} to {filename}")
             except requests.HTTPError as http_error:
                 if http_error.response.status_code == 401:
                     log.error(f"Access denied to {file_link}")
                 else:
                     log.error(f"Failed to download {file_link}. {http_error}")
-            except Exception as ex:
+            except RuntimeError as ex:
```
bsipocz, line 814: Thanks for fixing this! We should indeed avoid catching generic `Exceptions`

astroquery/eso/core.py
```diff
@@ -769,9 +915,11 @@ def get_associated_files(self, datasets: List[str], *, mode: str = "raw",
         # remove input datasets from calselector results
         return list(associated_files.difference(set(datasets)))
 
-    @deprecated_renamed_argument(('request_all_objects', 'request_id'), (None, None), since=['0.4.7', '0.4.7'])
-    def retrieve_data(self, datasets, *, continuation=False, destination=None, with_calib=None,
-                      request_all_objects=False, unzip=True, request_id=None):
+    @deprecated_renamed_argument(('request_all_objects', 'request_id'), (None, None),
+                                 since=['0.4.7', '0.4.7'])
```
bsipocz: This is a bit unrelated so no need to do it in this PR, but 0.4.7 was a while ago, so these deprecations can be now removed.

astroquery/eso/utils.py
```diff
@@ -0,0 +1,197 @@
...
+    if allowed_values:
+        where_constraints.append(_format_helper(allowed_values))
+
+    where_constraints += [
+        f"{k} {_adql_sanitize_op_val(v)}" for k, v in column_filters.items()
+    ]
+    return where_constraints
+
+
+def reorder_columns(table: Table,
```
bsipocz: I have multiple questions here:

- is this really necessary? Why the default column order is not suffucient?
- but if this is necessary, then please just go by suggested way for table column reordering from the astropy docs, "Select or reorder columns" on this page: https://docs.astropy.org/en/stable/table/modify_table.html

Also, minor nitpicking, but all the other functions here are private, why is this one not?

astroquery/eso/utils.py
```diff
@@ -0,0 +1,197 @@
...
+                    leading_cols: Optional[List[str]] = None):
+    """
+    Reorders the columns of the pased table so that the
+    colums given by the list leading_cols are first.
+    If no leading cols are passed, it defaults to
+    ['object', 'ra', 'dec', 'dp_id', 'date_obs']
+    Returns a table with the columns reordered.
+    """
+    if not isinstance(table, Table):
+        return table
```
bsipocz, line 93: Is this an expected case?

astroquery/eso/tests/test_eso_remote.py
bsipocz, line 1: I haven't reviewed the tests, but running them returned 77 failures, so please have a look.

`pytest -P eso -R` should work if you have all the test dependencies installed.

docs/eso/eso.rst
```diff
@@ -4,6 +4,33 @@
 ESO Queries (`astroquery.eso`)
 ******************************
 
+.. warning::
```
bsipocz, line 7: This is very good! Thanks for adding.

bsipocz (commented):
I'm sorry to have this many comments and thank you for your patience while waiting for this review.

I acknowledge that this seems to a lot, but most of them are minor and/or helping API consistency with the other modules and/or making the code cleaner and easier to read and follow; which will be helpful for the future maintenance.

astroquery/eso/core.py
juanmcloaiza: I know this is not a valid answer, but "because it was so in the main branch, L30", so I left it like that. I'll address this comment later, when I have a valid answer.

astroquery/eso/core.py
juanmcloaiza: `__all__` is present in `eso/__init__.py` (L28):

```
__all__ = ['Eso', 'EsoClass',
           'Conf', 'conf',
           ]
```

Is this what you're referring to? Perhaps I'm missing the point.

astroquery/eso/core.py
juanmcloaiza: We also discussed this point internally. While both `maxrec` and `top` limit the number of results, their intent is different. `top` is part of the ADQL query, used limit the result size from a logical/scientific standpoint. `maxrec`, on the other hand, is a TAP server parameter used as a safety cap regardless of the query’s intent. For example, I'd use `maxrec` to avoid system overload when I'm on my laptop, and for quickly testing the notebook’s logic and plots before the "real-thing-run" that could last some hours.

Since `maxrec` reflects user's environment rather than query logic, it makes more sense to have it as a configurable property, rather than a per-query argument (as `top`). Regarding other modules: if they don’t yet distinguish between `maxrec` and `top`, I’d warmly suggest they do, since it’s an important conceptual separation.

Please let me know your thoughts on this, before I proceed to make any changes. I'd very much appreciate if we can leave `maxrec` as an instance property, though :)

astroquery/eso/utils.py
juanmcloaiza: yes, `table` may be a string or an int, depending on the user parameters `query_str_only` only and `count_only`.

I admit this is not the best software design at this point, but a complete refactoring would have taken more time from both developers and reviewers. We will come with better encapsulation, separation of concerns,  etc in the next pull request. Our aim for now was to switch to TAP with as little impact as possible to the API and to the overall code.

astroquery/eso/tests/test_eso_remote.py
juanmcloaiza: All tests were passing originally, but now I updated the fork from upstream and indeed, they are failing. I'll have a look.

astroquery/eso/core.py
juanmcloaiza: I see... this is getting the `tap_url` from the config *UNLESS* the env var `ESO_TAP_URL` is set. This is mostly for debugging purposes: to use a test server rather than the one in the configuration. This is transparent for the user, but if you have a strong position against it, I can just remove it:

Change

```
    def _tap_url(self) -> str:
        url = os.environ.get('ESO_TAP_URL', conf.tap_url)
        return url
```

for

```
_tap_url = conf.tap_url
```

And propagate that change throughout the module.... or use conf.tap_url directly all-around.

juanmcloaiza: (just adding that allowing to use the ESO_TAP_URL env var, would make my life easier as a developer.)

astroquery/eso/core.py
juanmcloaiza: It improves readability and separates the concerns of exception handling:

```
    def query_tap_service(...):
        table_to_return = None
        
        # handles authentication:
        tap_service = self._tap_service(authenticated)
        
         # handles the querying and its exceptions:
        table_to_return = self._try_download_pyvo_table(query_str, tap_service)
        
        return table_to_return
```

Putting all the code in a single function would be messy and the logical parts would be difficult to distinguish.

astroquery/eso/core.py
juanmcloaiza: This is relevant for some of your previous comments as well. I'm only marking it for the moment to take a closer look to all related comments together.

astroquery/eso/core.py
juanmcloaiza: Well, I can assure you it made the code more writeable 😄  ... but you're right -- shorthand removed.

astroquery/eso/utils.py
juanmcloaiza: Not really... this is part of a larger problem that makes me uneasy: **The underlying querying language (wdb, tap, or whatever) is creeping up to the python API**... I'd like to address this problem in an upcoming pull request. I think that the API should be strictly python, no semi-strings related to the underlying querying tool... i.e., either users go all-in writing an ADQL query via `query_tap(...)` or they use the python API with the available 100% python arguments... 

Now, focusing on the purpose of this function in the current API: the user can pass a very valid dictionary to the query functions, for example `{... , 'box' = 'whatever', ...}`. However, the keyword `box` is a WDB keyword not available in TAP. As far as the **python** API is concerned, there's no deprecation since the dictionary is well formed, but TAP will surely complain. This special function is a courtesy to our users, to tell them what went wrong and how to fix. If you insist, we can remove the function... it is not strictly necessary and discussing it further doesn't deserve our time.

astroquery/eso/core.py
juanmcloaiza: You're right, and indeed, an upcoming pull request is planed to greatly improve the way the code is written. This PR focuses on switching the backend from WDB to TAP trying to impact as little as possible the API and the code structure (and it still grew larger than expected). Please be patient with these kind of apparently un-needed methods for the moment. They will disappear very soon.

astroquery/eso/core.py
juanmcloaiza: `query_instruments` and `query_main` query from two different data sources, so they are both needed. Some renaming is indeed needed but, after some internal discussion, it was decided that the names will be kept for now and changed in the next PR.

astroquery/eso/core.py
juanmcloaiza: query_instruments and query_main query from two different data sources, so they are both needed.

astroquery/eso/core.py
juanmcloaiza: renamed to `item`

astroquery/eso/core.py
juanmcloaiza: For the moment, I am renaming the function `_tap_service` to `tap`. Making it a property would have repercussions in other parts of the code which at this point I'd avoid touching. I would leave these larger refactoring changes to a next Pull Request, where we plan a larger refactoring, if you agree.

astroquery/eso/core.py
juanmcloaiza: Here I am after taking a closer look...

ALMA:
```
    @property
    def tap(self):
        if not self._tap:
            self._tap = pyvo.dal.tap.TAPService(..., session=self._session)
        return self._tap

    def query_tap(self, query, *, maxrec=None, uploads=None):
        return self.tap.search(query, language='ADQL', maxrec=maxrec, uploads=uploads)
```

Alma does indeed looks simpler. The reason for this seems to be that the authentication is being handled differently by the two modules, which simplifies the `query_tap` and `tap` methods. 

We didn't want to touch authentication (among other aspects) to keep the PR focused on moving from WDB to TAP but, as I mention in several of my comments, the ESO module requires a large refactoring and we can address all of these issues there.

We could make the module similar to ALMA, or any blueprint module you suggest, since it makes a lot of sense to have things homogeneized.

astroquery/eso/core.py
juanmcloaiza: There are several parts we can improve further, but this PR is already larger than expected (and took from our side more time than expected) for "only" changing the query backend. We would prefer to keep the updates bound to the bare essentials of enabling TAP. The rest will come later. Unless you consider it is a real blocker, we would like to restrict the changes to the bare minimum.

Thanks for the advice though. As I mention elsewhere, I think it is a very good practice to keep the modules homogenized, so we are open to your directions. 👍

astroquery/eso/core.py
juanmcloaiza: https://github.com/astropy/astroquery/pull/3339#discussion_r2253944672 <-- also applies here

astroquery/eso/utils.py
juanmcloaiza: Renamed to `_build_adql_string`

astroquery/eso/utils.py
juanmcloaiza: Hmmm... but the function is already doing exactly what the docs suggest (isn't it?)...

From the docs:
```
new_order = ['a', 'c', 'b']  # List or tuple
t_acb = t[new_order]
```

From the ESO module:
```
[...]
last_cols = first_cols + last_cols  # this is a list of column names
table = table[last_cols]
```

The code above those two lines is just computing the list of column names.

Other than that,
 - the function is now private, good point.
 - yes, the reordering is necessary, because the default is alphabetical order and the first fields are long urls and other info that is not legible and take much space. The reorder was suggested by several beta testers, which were expecting the columns in the order that this function generates.

juanmcloaiza:
@bsipocz, thanks a lot for your comments, I have addressed most of them (the easy ones); for the rest, I'd still need some feedback from you. If you can give me feedback today, I can work immediately on it and push and update by tomorrow. Afterwards, I won't be able to work on Astroquery for the next couple of weeks. Just so you know, in case you can squeeze this review in your day 🤓 Thanks again! 👍

astroquery/eso/tests/test_eso_remote.py
juanmcloaiza: tests are passing again ✅

astroquery/eso/core.py
juanmcloaiza: We do have the `get_query_payload` argument (previously called `query_string_only`), however, showing the query string by default on these exceptions is really helpful for debugging. I would like to keep these exception-handling as it is, but await for your final verdict.

astroquery/eso/core.py
juanmcloaiza: OK, I removed the flag `__doctest_skip__`.

bsipocz:
@juanmcloaiza - I'm sorry, it's very unlikely that I can get back to this today, but I'll try to do it soon so you can pick it up again in a few weeks time.

juanmcloaiza:
Hi @bsipocz , just a ping about the second review round. Have you had some time for it? We're looking forward to hearing back from you.

astroquery/eso/core.py
keflavich: I'm onboard with keeping these as separate, and the use case makes sense - could you add some documentation about it?

I would recommend, however, that you also make it a per-instance configurable item, such that you can do:
```python
eso_max50 = EsoClass(maxrec=50)
```
or 
```python
eso_max50 = EsoClass()
eso_max50.maxrec = 50
```

I personally find the former approach more useful, since it can potentially serve both use cases.

Either way, I'd recommend documenting it like this, but @bsipocz please chime in:
 * Describe it specifically for the ESO module, but
 * Plan for us to eventually move that description to a more generic section of the docs and just link from ESO if/when other modules adopt the same approach

astroquery/eso/core.py
keflavich: I see the use case here, but I also see good reason not to have so many ways to declare the same thing.

Throughout astroquery, we don't really support using shell-level environmental variables; our general approach has been to keep things within python.

So instead of
```
ESO_TAP_URL = 'http://test'
python -c "from astroquery.eso import Eso; Eso.query_foo()"
```
we'd prefer
```
python -c "from astroquery.eso import Eso; Eso(tap_url='http://test').query_foo()"
```

@bsipocz this is a somewhat broader discussion, though.  I really like the approach I've described here, as opposed to
```python
from astroquery.eso import Eso
Eso.tap_url = 'http://test'
Eso.query_foo()
```
because it's much harder to force the latter approach into a one-liner.

@juanmcloaiza what is the pattern you're following that uses the environmental variable?  Is it like the first example I showed above?  Would the alternative approach I suggested work as well, or is there reason the env var is easier?

astroquery/eso/core.py
keflavich: I see the use case here; I think it's nice to see the query code when it fails.  

However, is there a reason to change the exception type to `RuntimeError`?  Why not keep the exception type the same as was originally raised (but append the query code to the message)?

astroquery/eso/core.py
keflavich: Fine to leave as-is based on your logic.  Let's make a new issue with suggested breaking API changes, then?

@bsipocz 's point is that now is a "safe" time, from the users' perspective, to make API-breaking changes, since there is little reason to expect the API to stay consistent when completely overhauling the backend.  To make this change in the future, we will need to go through a deprecation cycle, which will take additional time.

astroquery/eso/core.py
```diff
@@ -296,248 +214,468 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
-            for element in (collection_elts + other_elts):
-                if 'value' in element.attrs:
-                    survey = element.attrs['value']
-                    if survey and survey not in self._survey_list and 'Any' not in survey:
-                        self._survey_list.append(survey)
-        return self._survey_list
-
-    def query_surveys(self, *, surveys='', cache=True,
-                      help=False, open_form=False, **kwargs):
+        _ = cache  # We're aware about disregarding the argument
```
keflavich, line 363: Can we instead
```suggestion
        if cache:
            raise NotImplementedError("caching is not enabled for this query type")
```
and default to cache=False?

As a user, I really do not want to run a query with cache=True a large number of times and find out only by digging into the source code that I was performing a lot of queries that I thought were just local cache reads.

astroquery/eso/core.py
```diff
@@ -296,248 +214,468 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        # shown tidier
+        nlines = len(available_cols) + 2
+        n_ = astropy.conf.max_lines
+        m_ = astropy.conf.max_width
+        astropy.conf.max_lines = nlines
+        astropy.conf.max_width = sys.maxsize
+        log.info(f"\nColumns present in the table {table_name}:\n{available_cols}\n"
+                 f"\nNumber of records present in the table {table_name}:\n{num_records}\n")
+        astropy.conf.max_lines = n_
+        astropy.conf.max_width = m_
```
keflavich: It feels like there should be a [context manager approach](https://docs.astropy.org/en/stable/api/astropy.config.ConfigNamespace.html#astropy.config.ConfigNamespace.set_temp) to this.  I haven't tested this code, but a chatbot claims it works...
```suggestion
        with astropy.conf.set_temp("max_lines", nlines, "max_width", sys.maxsize):
            log.info(
                f"\nColumns present in the table {table_name}:\n{available_cols}\n"
                f"\nNumber of records present in the table {table_name}:\n{num_records}\n"
            )
```

astroquery/eso/core.py
```diff
@@ -296,248 +214,468 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+            - When ``get_query_payload`` is ``True``, returns the query string that
+              would be issued to the TAP service given the specified arguments.
         """
-        url = self.QUERY_INSTRUMENT_URL + "/eso_archive_main/form"
-        return self._query(url, column_filters=column_filters, columns=columns,
-                           open_form=open_form, help=help, cache=cache, **kwargs)
-
-    def query_instrument(self, instrument, *, column_filters={}, columns=[],
-                         open_form=False, help=False, cache=True, **kwargs):
+        _ = open_form, cache  # make explicit that we are aware these arguments are unused
```
keflavich, line 573: ditto [above](https://github.com/astropy/astroquery/pull/3339/files#r2370839294): tell the _users_ that these are being ignored

astroquery/eso/core.py
```diff
@@ -296,248 +214,468 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
-            # First line is always garbage
-            content = content.split(b'\n', 1)[1]
-            log.debug("Response content:\n{0}".format(content))
-            if _check_response(content):
-                table = Table.read(BytesIO(content), format="ascii.csv",
-                                   comment='^#')
-                return table
-            else:
-                warnings.warn("Query returned no results", NoResultsWarning)
+        _ = open_form, cache  # make explicit that we are aware these arguments are unused
```
keflavich, line 661: ditto [above](https://github.com/astropy/astroquery/pull/3339/files#r2370839294)

astroquery/eso/core.py
```diff
@@ -801,15 +939,18 @@ def retrieve_data(self, datasets, *, continuation=False, destination=None, with_
...
         --------
-        >>> dptbl = Eso.query_instrument('apex', pi_coi='ginsburg')
-        >>> dpids = [row['DP.ID'] for row in dptbl if 'Map' in row['Object']]
-        >>> files = Eso.retrieve_data(dpids)
+        dptbl = Eso.query_instrument('apex', pi_coi='ginsburg')
+        dpids = [row['DP.ID'] for row in dptbl if 'Map' in row['Object']]
+        files = Eso.retrieve_data(dpids)
 
         """
+        _ = request_all_objects, request_id
```
keflavich: ditto [above](https://github.com/astropy/astroquery/pull/3339/files#r2370839294)

astroquery/eso/core.py
```diff
@@ -837,191 +979,80 @@ def retrieve_data(self, datasets, *, continuation=False, destination=None, with_
...
-                        checkbox = "   "
-                    if name != u"":
-                        result_string.append("{0} {1}: {2}"
-                                             .format(checkbox, name, value))
-                        if name.strip() in helptext_dict:
-                            result_string.append(helptext_dict[name.strip()])
-
-        log.info("\n".join(result_string))
-        return result_string
+        _ = open_form, cache  # make explicit that we are aware these arguments are unused
```
keflavich, line 1040: Here I think there is a good user-facing message that these arguments are already deprecated (because of the deprecation decorator), so this text is fine.

astroquery/eso/utils.py
keflavich: I think I see the use case here, but I am ignorant of wdb/tab.  If this is really a separate messaging about creating bad TAP queries, I think that's valuable to include?  But I would recommend documenting this and showing an example of how a malformed query might come about (presumably only if a user writes their own custom ADQL?) and the message that will be shown.

But I might be talking nonsense as I'm not sure I appreciate the context yet.  Also it's possible you've documented it and I haven't gotten that far in the review yet...

astroquery/eso/core.py
keflavich: Coming back to this at the end of the review: Please change `maxrec` to `ROW_LIMIT`.  This is part of the astroquery user interface and should not be changed. 

https://github.com/search?q=repo%3Aastropy%2Fastroquery%20ROW_LIMIT&type=code

I'll raise another issue about this, though, as it appears we have mixed case...

astroquery/eso/core.py
juanmcloaiza: Sure, but isn't it the canonical way to treat these situations the `deprecated_renamed_argument` decorator from `astropy`?

```
def deprecated_renamed_argument(
    old_name,
    new_name,
    since,
...
):
    """Deprecate a _renamed_ or _removed_ function argument.

    The decorator assumes that the argument with the ``old_name`` was removed
    from the function signature and the ``new_name`` replaced it at the
    **same position** in the signature.  If the ``old_name`` argument is
    given when calling the decorated function the decorator will catch it and
    issue a deprecation warning and pass it on as ``new_name`` argument.

    Parameters
    ----------
    old_name : str or sequence of str
        The old name of the argument.

    new_name : str or sequence of str or None
        The new name of the argument. Set this to `None` to remove the
        argument ``old_name`` instead of renaming it.

    since : str or number or sequence of str or number
        The release at which the old argument became deprecated.
...
```

So just to be sure... could you please confirm that the `deprecated_renamed _argument` decorator is not the way to go / not enough? - for example, like [here](https://github.com/astropy/astroquery/pull/3339#discussion_r2370846629)

astroquery/eso/core.py
juanmcloaiza: ditto [above](https://github.com/astropy/astroquery/pull/3339/files#r2370839294): aren't the users already being told via de astropy deprecation warning?

astroquery/eso/core.py
juanmcloaiza: Nice suggestion! I'll try it 👍

astroquery/eso/core.py
juanmcloaiza: This function is different than the rest... 

On the `main` branch, the signature looks like this:

```
    @deprecated_renamed_argument(('request_all_objects', 'request_id'), (None, None), since=['0.4.7', '0.4.7'])
    def retrieve_data(self, datasets, *, continuation=False, destination=None, with_calib=None,
                      request_all_objects=False, unzip=True, request_id=None):
        """
        Retrieve a list of datasets form the ESO archive.
...
```

Then the `request_all_objects` and `request_id` are ignored and not used anymore throughout the function. There's not even a hint to the developers that those arguments are being ignored (Like the underscores I use).

I think that, in this case, since the arguments were deprecated already in version 0.4.7, we can remove the deprecation decorator AND also remove the arguments from the signature altogether. (In fact, the deprecation decorator was already removed, [see here](https://github.com/astropy/astroquery/pull/3339#discussion_r2244201212)).

Unless you object, I'd remove the arguments `request_all_objects`, `request_id` altogether.

astroquery/eso/core.py
keflavich: No, the deprecation warning is adequate.  I misread the diff because the deprecation is so far away from this change.  Sorry! Ignore these comments if deprecation is already handled

juanmcloaiza:
There's no field for me to reply to your las two comments (strange...), but here are my replies:

----

[About the maxrec](https://github.com/astropy/astroquery/pull/3339#discussion_r2370853722) - Sure thing, ~I'll make the change~ I've replaced `maxrec` by `ROW_LIMIT`.

----

[About the deprecated keys message](https://github.com/astropy/astroquery/pull/3339#discussion_r2370848771) - The problem is that the astroquery ESO module (like many others) is currently using the keywords of the underlying querying engine directly in its API (like `box`, `coord1`, `coord2` of WDB). These are not python arguments, but keys of a very normal dictionary that is passed to the python API and transformed internally to a WDB or TAP query. Since these are not python arguments, we need ad-hoc cecking, and we do it only for the most common WDB keywords.

The messages users get are, for example, `"box, coord1 and coord2 are deprecated; use cone_ra, cone_dec and cone_radius instead."`

Without  `_raise_if_has_deprecated_keys`, the code would explode later on with a message about a bad TAP query, printing the query itself, but without saying explicitly that the problem is in the keywords.
 
@ashleythomasbarnes has already an extensive documentation with examples ready to be submitted after this PR. For the moment, the changes to the documentation in this PR, include a warning and a how-to regarding the incompatibility between WDB and TAP:
 - https://github.com/juanmcloaiza/astroquery/blob/332b36e19f0cb34ea5436ef986b1b6cad6d1753f/docs/eso/eso.rst?plain=1#L9
  - https://github.com/juanmcloaiza/astroquery/blob/332b36e19f0cb34ea5436ef986b1b6cad6d1753f/docs/eso/eso.rst?plain=1#L446

----

Thanks for the feedback, @keflavich, I'll get everything done this week. Please just let me know about my question regarding the deprecation of arguments; for the rest everything is clear 👍

astroquery/eso/core.py
juanmcloaiza: arguments `request_all_objects` and `request_id` removed

astroquery/eso/core.py
juanmcloaiza: So, the chatbot approach didn't work exactly, but needed a small tweak:

```python
        with (astropy.conf.set_temp(
                "max_lines", len(available_cols) + 2),
                astropy.conf.set_temp(
                "max_width", sys.maxsize)):
            log.info(f"\nColumns present in the table {table_name}:\n{available_cols}\n"
                     f"\nNumber of records present in the table {table_name}:\n{num_records}\n")
```    

-- `astropy.conf.set_temp` needs to be called for each parameter.
Thanks for the suggestion! 👍

juanmcloaiza:
@keflavich, your comments have been taken into account, thanks for having taken the time. [cc @bsipocz, @ashleythomasbarnes, @szampier].

astroquery/eso/core.py
juanmcloaiza: No worries here, I removed the env var altogether.

juanmcloaiza:
@keflavich, your comments have been taken into account, thanks for having taken the time. [cc @bsipocz, @ashleythomasbarnes, @szampier]. -- EDIT: Apologies, I went again one by one through all the conversations and I noticed three comments that had escaped my sight. I have taken them into account now. Please double check and let me know if anything is still missing.

astroquery/eso/core.py
juanmcloaiza: I'm noticing this just now. The reply is here: https://github.com/astropy/astroquery/pull/3339#issuecomment-3327400641, `maxrec` was substituted by `ROW_LIMIT`, and yes,

```
eso_max50 = EsoClass()
eso_max50.maxrec = 50
```

This works 👍

juanmcloaiza:
@keflavich, @bsipocz, just FYI, on making the requested changes I introduced a bug that is now fixed in commit [aef5c0f](https://github.com/astropy/astroquery/pull/3339/commits/aef5c0f2352adf9bd10aafa737872042a702b8d9) [@ashleythomasbarnes, @szampier].

astroquery/eso/core.py
```diff
@@ -62,168 +58,90 @@ def _get_exp_time_from_token(self) -> int:
...
+        """
+        Returns the name of the instrument specific table (IST)
+        """
+        return f"ist.{instrument_name}"
+
+    apex_quicklooks_table = ist_table.__func__("apex_quicklooks")
+    apex_quicklooks_pid_column = "project_id"
+
+
+def unlimited_maxrec(func):
```
bsipocz, line 86: we may not want to expose this to the users, have an `__all__` to work it around

astroquery/eso/core.py
```diff
@@ -1,59 +1,55 @@
...
-    If all is OK, return True
-    """
-    if b"NETWORKPROBLEM" in content:
-        raise RemoteServiceError("The query resulted in a network "
-                                 "problem; the service may be offline.")
-    elif b"# No data returned !" not in content:
-        return True
+from .utils import _UserParams, raise_if_coords_not_valid, _reorder_columns, \
+    _raise_if_has_deprecated_keys, _build_adql_string, \
+    DEFAULT_LEAD_COLS_PHASE3, DEFAULT_LEAD_COLS_RAW
```
bsipocz, line 41: have an all to keep the tab complete clean and to signal the public API.
```suggestion
    DEFAULT_LEAD_COLS_PHASE3, DEFAULT_LEAD_COLS_RAW
    
    
__all__ = ['Eso', 'EsoClass']
```

astroquery/eso/core.py
bsipocz: I agree with the env variable (though, I think we need to keep the configs). However, I prefer using the config system rather than changing the URLs on the instances like in both examples above.

Anyway, that is a bigger picture item that we need to sort out independently of this PR/module.

astroquery/eso/core.py
bsipocz: but what is a `pyvo_table`, and is this really a download or instead of running a query? 
(we usually use "download" to refer to downloading files from an archive as opposed to run a table query.

Having a parser method is quite usual in the other modules, especially if it can be reused for multiple methods; but here I don't see we gain that much readibility with the added complexity.

astroquery/eso/core.py
```diff
@@ -296,248 +214,465 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+
+        try:
+            row_limit_plus_one = self.ROW_LIMIT
+            if self.ROW_LIMIT < sys.maxsize:
+                row_limit_plus_one = self.ROW_LIMIT + 1
+
+            table_with_an_extra_row = tap.search(query=query_str, maxrec=row_limit_plus_one).to_table()
+            self._maybe_warn_about_table_length(table_with_an_extra_row, row_limit_plus_one)
+        except DALQueryError:
+            log.error(message(query_str))
```
bsipocz, line 262: Why don't we raise this error, and why swallow the original exception type from the log?

astroquery/eso/core.py
```diff
@@ -296,248 +214,465 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        except DALFormatError as e:
+            raise DALFormatError(message(query_str) + f"cause: {e.cause}") from e
+        except Exception as e:
+            raise type(e)(f"{e}\n" + message(query_str)) from e
```
bsipocz, line 266: practically these two are doing the same, you raise the original exception along with the new exception and a new explanation; merge them together.

astroquery/eso/core.py
```diff
@@ -296,248 +214,465 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+
+        return table_with_an_extra_row[:self.ROW_LIMIT]
+
+    def tap(self, authenticated: bool = False) -> TAPService:
```
bsipocz, line 270: Make this a property as we have it in the other modules

astroquery/eso/core.py
bsipocz: yeap, exactly the reasoning that now is a good time to break the API. 

But also, "now" is not necessarily right now in this PR, but "now" as before the next release.

astroquery/eso/core.py
```diff
@@ -296,248 +214,465 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        # Remove ist.apex_quicklooks, which is not actually a raw instrument
+        if _EsoNames.apex_quicklooks_table in l_res:
+            l_res.remove(_EsoNames.apex_quicklooks_table)
+
+        l_res = list(map(lambda x: x.split(".")[1], l_res))
 
+        return l_res
+
+    @unlimited_maxrec
+    @deprecated_renamed_argument('cache', None, since='0.4.11')
```
bsipocz:
```suggestion
    @deprecated_renamed_argument('cache', None, since='0.4.12')
```

astroquery/eso/core.py
```diff
@@ -296,248 +214,465 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        eso_instance = Eso()
+        eso_instance.query_tap("SELECT * FROM ivoa.ObsCore")
+        """
+        table_to_return = Table()
+        tap_service = self.tap(authenticated)
+        table_to_return = self._try_download_pyvo_table(query, tap_service)
+        return table_to_return
+
+    @unlimited_maxrec
+    @deprecated_renamed_argument('cache', None, since='0.4.11')
```
bsipocz:
```suggestion
    @deprecated_renamed_argument('cache', None, since='0.4.12')
```

astroquery/eso/core.py
```diff
@@ -296,248 +214,465 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+        query = _build_adql_string(user_params)
+
+        if user_params.get_query_payload:
+            return query
+
+        ret_table = self.query_tap(query=query, authenticated=user_params.authenticated)
+        return list(ret_table[0].values())[0] if user_params.count_only else ret_table
+
+    @deprecated_renamed_argument(('open_form', 'cache'), (None, None),
+                                 since=['0.4.11', '0.4.11'])
```
bsipocz:
```suggestion
                                 since=['0.4.12, '0.4.12'])
```

astroquery/eso/core.py
```diff
@@ -296,248 +214,465 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+    def query_surveys(
+            self,
+            surveys: Union[List[str], str] = None, *,
```
bsipocz, line 417: While doing typing is your choice of an archive who contribute and maintain their module, IMO it doesn't improve anything for astroquery, we don't enforce it, don't test it, it just make it less readable.

astroquery/eso/core.py
```diff
@@ -296,248 +214,465 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+                                  get_query_payload=get_query_payload,
+                                  print_help=help,
+                                  authenticated=authenticated,
+                                  )
+        t = self._query_on_allowed_values(user_params=user_params)
+        t = _reorder_columns(t, DEFAULT_LEAD_COLS_PHASE3)
+        return t
+
+    @deprecated_renamed_argument(('open_form', 'cache'), (None, None),
+                                 since=['0.4.11', '0.4.11'])
```
bsipocz:
```suggestion
                                 since=['0.4.12', '0.4.12'])
```

astroquery/eso/core.py
bsipocz: Keeping this unresolved so it's easier to pick up the comment for a follow-up PR

astroquery/eso/core.py
bsipocz: Unresolve this to make it easier to find for a follow-up

astroquery/eso/core.py
```diff
@@ -801,15 +935,17 @@ def retrieve_data(self, datasets, *, continuation=False, destination=None, with_
 
         Examples
         --------
-        >>> dptbl = Eso.query_instrument('apex', pi_coi='ginsburg')
-        >>> dpids = [row['DP.ID'] for row in dptbl if 'Map' in row['Object']]
-        >>> files = Eso.retrieve_data(dpids)
+        dptbl = Eso.query_instrument('apex', pi_coi='ginsburg')
+        dpids = [row['DP.ID'] for row in dptbl if 'Map' in row['Object']]
+        files = Eso.retrieve_data(dpids)
```
bsipocz, line 943: Why did you remove the doctesting here? IMO if there are examples included they should be runable (meaning that they will also likely need to be fixed up with an import line)

astroquery/eso/core.py
```diff
@@ -837,191 +974,80 @@ def retrieve_data(self, datasets, *, continuation=False, destination=None, with_
         log.info("Done!")
         return files[0] if files and len(files) == 1 and return_string else files
 
-    def query_apex_quicklooks(self, *, project_id=None, help=False,
-                              open_form=False, cache=True, **kwargs):
+    @deprecated_renamed_argument(('open_form', 'cache'), (None, None),
+                                 since=['0.4.11', '0.4.11'])
```
bsipocz:
```suggestion
                                 since=['0.4.12, '0.4.12'])
```

astroquery/eso/core.py
```diff
@@ -271,7 +189,7 @@ def _get_auth_info(self, username: str, *, store_password: bool = False,
 
         return username, password
 
-    def _login(self, *, username: str = None, store_password: bool = False,
+    def _login(self, username: str = None, store_password: bool = False,
```
bsipocz, line 195: Why did you remove the mandatoriness of kwargs here? general rule of thumb to not allow function calls with long positional signatures.

astroquery/eso/core.py
```diff
@@ -296,248 +214,465 @@ def _login(self, *, username: str = None, store_password: bool = False,
...
+
+        ret_table = self.query_tap(query=query, authenticated=user_params.authenticated)
+        return list(ret_table[0].values())[0] if user_params.count_only else ret_table
+
+    @deprecated_renamed_argument(('open_form', 'cache'), (None, None),
+                                 since=['0.4.11', '0.4.11'])
+    def query_surveys(
+            self,
+            surveys: Union[List[str], str] = None, *,
+            cone_ra: float = None, cone_dec: float = None, cone_radius: float = None,
```
bsipocz, line 418: oh, I've just noticed these to be named weirdly. Please in a follow-up make it change to the usual API to rename `cone_ra/dec` to one argument that is called `coordinates` and which takes a SkyCoord object (or a string that can be turned to a SkyCoord/etc, this is very similar to all the other modules. 

And also rename `cone_radius` to just `radius`.

CHANGES.rst
```diff
@@ -71,6 +76,15 @@ esa.hubble
 
 - Internal refactor of the module to use to PyVO. [#3367]
 
+eso
+^^^
+
+- Switch querying interface from WDB to TAP in querying functions. [#3339]
+- Allow plain ADQL queries via ``query_tap_service`` (with authentication as well). [#3339]
```
```suggestion
- Allow plain ADQL queries via ``query_tap`` (with authentication as well). [#3339]
```

CHANGES.rst
```diff
@@ -71,6 +76,15 @@ esa.hubble
...
+- Cone search using ``cone_ra`, ``cone_dec`, ``cone_radius`` arguments. [#3339]
```
bsipocz, line 84: These will need to be renamed/refactored to match other modules.

bsipocz:
RTD failure is unrelated and is due to a recent plugin release which I'll resolved in a separate PR.

bsipocz:
Thanks @juanmcloaiza for the patience here. I'm looking fwd for the follow-up PRs.

juanmcloaiza:
@bsipocz , thanks a lot for your time today and the careful review. Looking forward for a continuous and fruitful collaboration! 🚀

PR #3342, opened 2025-06-10
GAIA: new DR4 datalink retrieve type MEAN_SPECTRUM_RVS

cosmoJFH:
Dear astroquery team,

we would like to rename the DR4 dataLink  "RVS"  retrieval_type to "MEAN_SPECTRUM_RVS" (compared to the Gaia DR3 value "RVS").  

Please keep in mind that this does not break the code because the Gaia DR4 is not yet available to the public.

cc @esdc-esac-esa-int 
jira:GAIASWRQ-45


cosmoJFH:
In https://github.com/astropy/astroquery/pull/3238 we committed the changes by mistake. Now we have also updated the gaia archive to handle this new retrieve type.

PR #3364, opened 2025-06-30
Clarifying documentation for `MastMissions` radius value

snbianco:
Quick PR to clarify the units on the default radius for positional searches in the `MastMissions` class. Some other minor docstring changes to make things more consistent.


bsipocz (approved):
Thanks!

PR #3368, opened 2025-07-10
Splatalogue: fix docstring to match valid intensity type

keflavich:
Minor change.  There may be others that need updating; give me a day to check


keflavich:
I didn't find anything in my 1-day deadline, so merged.

PR #3370, opened 2025-07-11
FIX: incorrect parameter for JPL Horizons, changed 'BODY EQUATOR' -> 'BODY'

TheWand3rer:
When querying the results of a JPL Horizons query via `elements(refplane='body')` the API returns this error:
```
ValueError: Query failed without known error message; received the following response:
API VERSION: 1.2
API SOURCE: NASA/JPL Horizons API


 INPUT ERROR in VLADD  in following line:
 REF_PLANE=BODY EQUATOR
                ^  Too many constants
BATVAR: problem loading execution-control setting:
LINE=REF_PLANE=BODY EQUATOR
WLDINI: error loading execution-control file.
```

By changing the line in `core.py` from `BODY EQUATOR` to `BODY`, that API call works again.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,11 @@ gaia
 - New method cross_match_basic that simplifies the positional x-match method [#3320]
 - new DR4 datalink retrieve type MEAN_SPECTRUM_RVS [#3342]
 
+jplhorizons
+^^^^^^^^^^^
+
+- Fixed incorrect ``refplane`` parameter value: the API did not accept ``BODY EQUATOR`` but required ``BODY``. [#3370]
+
 linelists.cdms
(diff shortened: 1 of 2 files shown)


mkelley:
Thanks, this fix looks good to me.  As a side note, I thought that quoting "BODY EQUATOR" would fix the problem too, but it resulted in the same error, so this could be a Horizons bug or problem with their documentation.  Regardless, the documentation does say that abbreviations are allowed, so I think this fix is correct.

@TheWand3rer Please add a change log entry.

sarahayu:
I contacted the folks behind the Horizons API and this has now been fixed on their end. 'BODY EQUATOR' and 'BODY' both work as REF_PLANE values, so no rollbacks are required, but thought this was a good FYI.

PR #3396, opened 2025-08-26
JPLSpec: Download everything as a workaround for nonfunctional queroes

keflavich:
They're still trying to fix the query tool:
https://github.com/astropy/astroquery/issues/3363

But in the meantime, the data are all there, and they only amount to half a gigabyte total, so it's reasonable for a user to just download the whole thing locally.  I may add some tools to interact with the locally-cloned database too.

Tests are needed.

@bsipocz I'm interested to hear your take on this sort of thing.  There are a couple other databases that I'd like to add, like the NASA-hosted OCDB ice database and the Leiden LIDA, that have similar issues - i.e., it's a lot easier to download the whole thing rather than interact with their UI.  I still think astroquery's the right place to put useful hacks like this, but let's talk about it.

### DIFF
--- a/astroquery/jplspec/core.py
+++ b/astroquery/jplspec/core.py
@@ -1,6 +1,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import os
+import re
+from urllib.parse import urljoin, urlparse
 import warnings
+from bs4 import BeautifulSoup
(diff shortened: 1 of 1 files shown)


keflavich:
superseded by #3302, #3455, and #3456

PR #3441, opened 2025-10-15
DOC: Remove defaults channel from RTD

pllim:
See:

* https://github.com/astropy/astropy/pull/18726


bsipocz (approved):
fair enough, though I wonder why we even bother with anything conda.

pllim:
Actually you might not even need it. I think you can grab graphviz from PyPI nowadays.

PR #3467, opened 2025-11-26
tesscut moving target cutouts make individual req per sector

falkben:


astroquery/mast/cutouts.py
```diff
@@ -314,6 +317,41 @@ def download_cutouts(self, *, coordinates=None, size=5, sector=None, product='SP
...
+                sector_table = self.get_sectors(objectname=objectname, moving_target=True, mt_type=mt_type)
+
+            localpath_table = Table(names=["Local Path"], dtype=[str])
+
+            if len(sector_table) == 0:
+                warnings.warn("Coordinates are not in any TESS sector.", NoResultsWarning)
+                return localpath_table
+
+            # Get unique sectors
+            unique_sectors = sorted(set(sector_table["sector"]))
```
snbianco: Can we pull this section of code into a separate helper function that can also be used in `get_cutouts`?

snbianco:
Could you add another test case to cover the scenario where no TESS sectors are returned?

snbianco:
Not sure what's up with that test failure, but it doesn't look like it's coming from MAST.

snbianco (approved):

PR #3475, opened 2025-12-04
add generator version of mast.Observations.download_products

eteq:
The main purpose of this PR is to add `download_products_iter`  method to `mast.Observations`. My motivation here is to allow something like the following:
```
for whatever in tqdm(Observations.download_products_iter(product_list)):
    print(... something of interest using whatever...)
```
which then allows downloading a large number of  files with both a progress bar and some custom progress reporting. Of course, ``verbose=True`` does something akin to this, but I find sometimes for large downloads I want more control of the progress reporting.

This is a draft at this stage, with no tests or doc updates because I figured it was better to do that before doing all the rest of the work in case this has been considered and rejected in the past.  There might also be cleaner ways to split up the generator and the non-generator version, but this seemed the most low-impact way in case the private method is used in the wild elsewhere.

cc @snbianco I think is the right person?

### DIFF
--- a/astroquery/mast/observations.py
+++ b/astroquery/mast/observations.py
@@ -718,7 +718,43 @@ class ObservationsClass(MastQueryWithLogin):
         response : `~astropy.table.Table`
         """
 
-        manifest_array = []
+        generator = self._download_files_generator(products, base_dir, flat=flat, 
+                                                             cache=cache, cloud_only=cloud_only, 
+                                                             verbose=verbose)
+        manifest_list = []
+        for local_path, status, msg, url, data_product_row in generator:
+            manifest_list.append((local_path, status, msg, url))
+        return Table(rows=manifest_list, names=('Local Path', 'Status', 'Message', "URL"))
+
+    def _download_files_generator(self, products, base_dir, *, flat=False, cache=True, cloud_only=False, verbose=True):
+        """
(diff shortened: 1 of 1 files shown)


snbianco:
Hey Eric, thanks for this PR! I do think this is worth doing as it provides more powers to users without breaking the API. Users often want to download a ton of data from us, and this solves a real unmet need that we have.

Since so much of `download_products` is duplicated in `download_products_iter`, I wondered if it might be better to add a parameter to `download_products`, something like `return_iter`. However, after thinking on it some more, that would muddy the API a bit, and I think the separated functions is the way to go. That being said, the duplicated logic should probably be pulled out into a helper function(s).

keflavich:
Agreed @snbianco, I like the idea of factoring out common code (e.g., filtering the selections) but keeping the API @eteq proposes

bsipocz:
> Agreed @snbianco, I like the idea of factoring out common code (e.g., filtering the selections) but keeping the API @eteq proposes

I'm not sure about the proposed API though, MAST is already such an odd ball out and this just makes it even more complicated. 

We still working on removing function duplicates that are doing almost the same, and thus I would be careful recreating the same problem down the line for the download functions.
So, ultimately having a kwarg that makes download responses generators is something we could/would suggest to other modules and I would highly prefer that.

Issue #3473, opened 2025-12-02
ESO: `authenticated` not enabled by default after login for queries


vandalt:
Hi! I have been testing the updated ESO implementation based on TAP query now that #3339 has been merged.

Everything is working great so far, but I had one hiccup with authentication: after logging with `Eso.login()`, TAP queries are not authenticated by default since the `authenticated` keyword argument to `query_instrument()` and `query_main()` is False by default. This is not a bug, but I found it somewhat unexpected. I was wondering if a note about this could be added in [the documentation](https://astroquery.readthedocs.io/en/latest/eso/eso.html#authentication-with-eso-user-portal), and/or if the default behaviour could be changed to `authenticated=True` when the user is logged in.

Happy to submit a docs and/or code PR depending on the preferred fix.

Thank you!

Here is a snippet to reproduce:

```python
eso = Eso()
eso.ROW_LIMIT = None
eso.login(username="APERO")
# Some files are missing without `authenticated=True`
files_tbl = eso.query_instrument("nirps", authenticated=True)
```

bsipocz:
cc @juanmcloaiza

juanmcloaiza:
Hi @vandalt, thanks for your feedback. It is indeed not clear that one needs the `authenticated=True` argument to download data as authenticated user. We have discussed internally and the way to proceed, at least for the moment, is to keep the default `authenticated=False`, since running authenticated queries by default makes the process much slower every time, even when you only need publicly available data.

I am trying to update the documentation with your suggestions in PR https://github.com/astropy/astroquery/pull/3477, but I don't manage to build the docs neither locally nor in the astropy CI-CD. The process hangs forever. @bsipocz , is this a common/known problem?

bsipocz:
> I am trying to update the documentation with your suggestions in PR https://github.com/astropy/astroquery/pull/3477, but I don't manage to build the docs neither locally nor in the astropy CI-CD. The process hangs forever. @bsipocz , is this a common/known problem?

Yes, there was a known issue with sphinx9 compatibility, I have merged a PR earlier today which fixed it.

PR #3477, opened 2025-12-09
Add docs note about authenticated queries

juanmcloaiza:
Addresses https://github.com/astropy/astroquery/issues/3473

### DIFF
--- a/docs/eso/eso.rst
+++ b/docs/eso/eso.rst
@@ -103,26 +103,32 @@ interaction with the ESO archive.
     INFO: Downloading file 1/1 https://dataportal.eso.org/dataPortal/file/ADP.2023-03-02T01:01:24.355
     ERROR: Access denied to https://dataportal.eso.org/dataPortal/file/ADP.2023-03-02T01:01:24.355
 
+
+.. The doctest-skip directive is not working for pytest within the note directive, thus
+   we leave the snippet below as a code-block
+
 .. note::
+
     Even after logging with `~astroquery.eso.EsoClass.login()`, TAP queries are not authenticated by default. To issue
-    authenticated queries, the keyword argument ``authenticated`` must be set to `True` explicitly in
-    the query functions. For example:
+    authenticated queries, the keyword argument ``authenticated`` must be set to `True` explicitly in the query functions.
+    For example:
 
-    .. doctest-skip::
+    .. code-block:: python
 
-        >>> eso = Eso()
-        >>> eso.ROW_LIMIT = None
-        >>> eso.login(username="ICONDOR")
-        >>>
-        >>> # Some files are missing without `authenticated=True`
-        >>> files_tbl = eso.query_instrument("nirps")
-        >>>
-        >>> # Get all files to which the authenticated user has access
-        >>> files_tbl = eso.query_instrument("nirps", authenticated=True)
+       eso = Eso()
+       eso.ROW_LIMIT = None
+       eso.login(username="ICONDOR")
+
+       # Some files are missing without `authenticated=True`
+       files_tbl = eso.query_instrument("nirps")
+
+       # Get all files to which the authenticated user has access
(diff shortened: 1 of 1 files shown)


keflavich (approved):
Good clear addition to the instructions.

bsipocz:
I see some docs build errors, I'll address them in review, rebase and then merge this.

Thank you @juanmcloaiza!

```
/Users/bsipocz/munka/devel/astroquery/docs/eso/eso.rst:107: WARNING: py:obj reference target not found: Eso.login() [ref.obj]
/Users/bsipocz/munka/devel/astroquery/docs/eso/eso.rst:107: WARNING: py:obj reference target not found: authenticated [ref.obj]
/Users/bsipocz/munka/devel/astroquery/docs/eso/eso.rst:123: WARNING: py:obj reference target not found: authenticated=False [ref.obj]
```

docs/eso/eso.rst
```diff
@@ -103,6 +103,27 @@ interaction with the ESO archive.
...
+
+        eso = Eso()
+        eso.ROW_LIMIT = None
+        eso.login(username="ICONDOR")
+
+        # Some files are missing without `authenticated=True`
+        files_tbl = eso.query_instrument("nirps")
+
+        # Get all files to which the authenticated user has access
+        files_tbl = eso.query_instrument("nirps", authenticated=True)
```
bsipocz: Even when we skip them, please make code snippets valid for doctest to pick up.

docs/eso/eso.rst
```diff
@@ -103,6 +103,27 @@ interaction with the ESO archive.
     INFO: Downloading file 1/1 https://dataportal.eso.org/dataPortal/file/ADP.2023-03-02T01:01:24.355
     ERROR: Access denied to https://dataportal.eso.org/dataPortal/file/ADP.2023-03-02T01:01:24.355
 
+.. note::
+    Even after logging with `Eso.login()`, TAP queries are not authenticated by default. To issue
```
bsipocz: you need to use full path otherwise sphinx is not picking up the API reference
```suggestion
    Even after logging with `~astroquery.eso.EsoClass.login()`, TAP queries are not authenticated by default. To issue
```

docs/eso/eso.rst
```diff
@@ -103,6 +103,27 @@ interaction with the ESO archive.
...
+    authenticated queries, the keyword argument `authenticated` must be set to `True` explicitly in
```
bsipocz: double backtick hightlights like this otherwise sphinx thinks it needs to find it as an API reference

```suggestion
    authenticated queries, the keyword argument ``authenticated`` must be set to `True` explicitly in
```

docs/eso/eso.rst
```diff
@@ -103,6 +103,27 @@ interaction with the ESO archive.
...
+
+        # Some files are missing without `authenticated=True`
+        files_tbl = eso.query_instrument("nirps")
+
+        # Get all files to which the authenticated user has access
+        files_tbl = eso.query_instrument("nirps", authenticated=True)
+
+    Review your queries carefully and update them accordingly to ensure you are retrieving
+    all your expected results. If you don't stricly need to run an authenticated query, use 
+    the default `authenticated=False`, since it will retrieve your results much faster.
```
bsipocz: same, double backtick
```suggestion
    the default ``authenticated=False``, since it will retrieve your results much faster.
```

bsipocz (approved):
I'm adding one more commit and rebase this, otherwise the note box look great. Thanks!

docs/eso/eso.rst
bsipocz: apparently pytest doesn't pick up doctest-skip as expected from inside a note directive, so I switch this to be a code-block, along with a comment about it.

docs/eso/eso.rst
juanmcloaiza: I see, thanks for the clarification 👍

PR #3488, opened 2026-01-05
Get cloud missions dynamically and fix cloud download workflow

snbianco:
This PR improves support for downloading MAST data products from public cloud storage (AWS S3) and clarifies cloud-related behavior for the `Observations` class. The changes introduce a more robust cloud download workflow with clearer fallback logic, improved error handling, and expanded test coverage for cloud and mixed cloud/on-prem download scenarios.

**Key Changes**

- Added discovery of cloud-supported datasets
  - Introduces a new internal method to query the public MAST S3 bucket and determine which datasets are available in the cloud.
  - Exposed via `Observations.list_cloud_datsets()` to allow users to inspect cloud availability.
- Refactored cloud download logic
  - Renamed and clarified cloud-specific download functionality (`download_file_from_cloud`) to separate cloud and on-prem paths.
  - Improved handling of `cloud_only`, fallback behavior, and warnings when products are missing from the cloud.
  - Ensures cloud URIs are resolved in batches rather than one-by-one.
- Improved cloud URI resolution
  - Updated path lookup logic to explicitly request cloud paths from MAST.
  - Unified cloud path handling across `get_cloud_uri`, `get_cloud_uris`, and download code paths.
- Expanded test coverage

**Backwards Compatibility**

- Cloud access remains opt-in via `enable_cloud_dataset()`.
- On-prem downloads remain the default fallback unless `cloud_only=True` is specified.
- This PR does change the download workflow in that if the cloud dataset is enabled, Astroquery will attempt to locate every requested product on the cloud. Previously, Astroquery would only attempt this if the product's mission was in a hard-coded list in `cloud.py`. This will prevent us from having to update Astroquery every time new data is made available on the cloud.
  - As a result of this change, when `cloud_only=True`, products that are not found on the cloud will not be downloaded. Previously, products of a mission not in the hard-coded list would go directly to on-prem and still be downloaded. In this case, I think that not downloading actually makes more sense in the context of the parameter.
- Deprecates `mast_relative_path` in favor of `get_cloud_paths`, which is more accurate and descriptive. This is a utility function, so I don't think many people will be using it anyway.

### DIFF
--- a/astroquery/mast/cloud.py
+++ b/astroquery/mast/cloud.py
@@ -148,7 +148,7 @@ class CloudAccess:  # pragma:no-cover
 
         # Making sure we got at least 1 URI from the query above.
         if not uri_list or uri_list[0] is None:
-            warnings.warn('Unable to locate file {}.'.format(data_product[0]), NoResultsWarning)
+            return
         else:
             # Output from ``get_cloud_uri_list`` is always a list even when it's only 1 URI
             return uri_list[0]
@@ -200,7 +200,7 @@ class CloudAccess:  # pragma:no-cover
                     if e.response['Error']['Code'] != "404":
                         raise
                     if verbose:
-                        warnings.warn("Unable to locate file {}.".format(path), NoResultsWarning)
+                        warnings.warn(f"Failed to retrieve cloud path for {path}", NoResultsWarning)
                     uri_list.append(None)
 
         return uri_list
@@ -222,9 +222,6 @@ class CloudAccess:  # pragma:no-cover
             Default is True. Whether to show download progress in the console.
         """
         # TODO: Function that checks if a particular product, by dataURI, can be found in the cloud
-        if not data_product or not isinstance(data_product, str):
-            raise ValueError("A valid data product URI must be provided.")
-
         # Normalize to an S3 key (no bucket)
         if data_product.strip().startswith("s3://"):
             s3_key = data_product.replace(f's3://{self.pubdata_bucket}/', '', 1)
--- a/astroquery/mast/cloud.py
+++ b/astroquery/mast/cloud.py
@@ -14,7 +14,7 @@ from astroquery import log
 from astropy.utils.console import ProgressBarOrSpinner
 from astropy.utils.exceptions import AstropyDeprecationWarning
 
-from ..exceptions import NoResultsWarning
+from ..exceptions import RemoteServiceError, NoResultsWarning
 
 from . import utils
 
@@ -223,14 +223,14 @@ class CloudAccess:  # pragma:no-cover
         """
         # TODO: Function that checks if a particular product, by dataURI, can be found in the cloud
         # Normalize to an S3 key (no bucket)
-        if data_product.strip().startswith("s3://"):
+        if data_product.strip().startswith('s3://'):
             s3_key = data_product.replace(f's3://{self.pubdata_bucket}/', '', 1)
         else:
             s3_key = self.get_cloud_uri_list([data_product], include_bucket=False, verbose=False)[0]
 
         # If s3_key is None, the product was not found in the cloud
         if s3_key is None:
-            raise ValueError(f"Data product {data_product} not found in S3 cloud storage.")
+            raise RemoteServiceError(f'The product {data_product} was not found in the cloud.')
 
         # Query S3 for expected file size
         head = self.s3_client.head_object(Bucket=self.pubdata_bucket, Key=s3_key)
--- a/astroquery/mast/cloud.py
+++ b/astroquery/mast/cloud.py
@@ -13,6 +13,7 @@ import threading
 from astroquery import log
 from astropy.utils.console import ProgressBarOrSpinner
 from astropy.utils.exceptions import AstropyDeprecationWarning
+from botocore.exceptions import ClientError, BotoCoreError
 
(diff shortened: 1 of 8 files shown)


bsipocz:
Could you please rebase?

astroquery/mast/cloud.py
```diff
@@ -52,39 +52,75 @@ def __init__(self, provider="AWS", profile=None, verbose=False):
...
+                    datasets.extend(
+                        cp['Prefix'].rstrip('/')
+                        for cp in mast_response.get('CommonPrefixes', [])
+                    )
+                else:
+                    datasets.append(prefix)
+
+            return datasets
+
+        except Exception as e:
```
bsipocz: Could you please be specific with the exceptions, do we really need to have an extremely blanket try/except here?

astroquery/mast/cloud.py
```diff
@@ -112,7 +148,7 @@ def get_cloud_uri(self, data_product, include_bucket=True, full_url=False):
 
         # Making sure we got at least 1 URI from the query above.
         if not uri_list or uri_list[0] is None:
-            warnings.warn("Unable to locate file {}.".format(data_product), NoResultsWarning)
```
bsipocz, line 115: Why not do the warning any more, is line https://github.com/astropy/astroquery/pull/3488/changes#diff-d3969a803b08a81a8cbe4b0146b60638e4aa2b2a6a2763c71b7e1aae08971245R203 covering all the cases?

astroquery/mast/observations.py
```diff
@@ -652,112 +655,165 @@ def download_file(self, uri, *, local_path=None, base_url=None, cache=True, clou
...
+        data_url = f"{base_url}?uri={uri}"
+        escaped_url = f"{base_url}?uri={quote(uri, safe=':/')}"
 
-        # create the full data URL
-        base_url = base_url if base_url else self._portal_api_connection.MAST_DOWNLOAD_URL
-        data_url = base_url + "?uri=" + uri
-        escaped_url = base_url + "?uri=" + quote(uri, safe=":/")
+        if cloud_only and force_on_prem:
+            warnings.warn("Both `cloud_only` and `force_on_prem` are set to True. "
+                          "`force_on_prem` will be ignored and cloud download will be attempted first.", InputWarning)
```
bsipocz: Given both of these are user parameters, I think this can be an `InputError` knowing that users tend to do a blanket warning ignore in their scripts/notebooks.

astroquery/mast/observations.py
```diff
@@ -652,112 +655,165 @@ def download_file(self, uri, *, local_path=None, base_url=None, cache=True, clou
...
+                    # Product not found in cloud
+                    if cloud_only:
+                        warnings.warn(f'The product {uri} was not found in the cloud. Skipping download.',
+                                      NoResultsWarning)
+                        return 'SKIPPED', None, None
+
+                    warnings.warn(f'The product {uri} was not found in the cloud. '
+                                  'Falling back to MAST download.', InputWarning)
+                    self._download_file(escaped_url, local_path, cache=cache, head_safe=True, verbose=verbose)
                 except Exception as ex:
```
bsipocz: Again, can we not do blanket try/excepts?

astroquery/mast/observations.py
```diff
@@ -652,112 +655,165 @@ def download_file(self, uri, *, local_path=None, base_url=None, cache=True, clou
...
+            status, msg, url = 'ERROR', None, None
+
+            try:
+                cloud_uri = cloud_uri_map.get(mast_uri) if cloud_uri_map else None
+
+                if cloud_uri:
+                    try:
+                        self._cloud_connection.download_file_from_cloud(cloud_uri, local_path, cache, verbose)
+                        status = 'COMPLETE'
+                    except Exception as ex:
```
bsipocz: be specific

astroquery/mast/observations.py
```diff
@@ -652,112 +655,165 @@ def download_file(self, uri, *, local_path=None, base_url=None, cache=True, clou
...
+                    else:
+                        # Cloud is not enabled
+                        if cloud_only:
+                            warnings.warn("`cloud_only` is True but cloud data access is not enabled. "
+                                          "Falling back to MAST download.", InputWarning)
+                        status, msg, url = self.download_file(mast_uri, local_path=local_path, cache=cache,
+                                                              cloud_only=False, force_on_prem=True, verbose=verbose)
 
-            manifest_array.append([local_path, status, msg, url])
+            except Exception as ex:
```
bsipocz: This is a rather long blanket try/except; can we refactor it and make it more specific?

bsipocz (commented):
Overall looks good, but I would prefer to have much fewer blanket try/excepts and wonder if we can be more specific about them.

Also, one minor suggestion to raise an error instead of a warning for contradicting input parameters.

astroquery/mast/cloud.py
snbianco: Yes, if a product doesn't have a cloud path, it should warn in `utils.get_cloud_paths`. If a cloud path is found but can't be accessed, a warning or error will be issued in `get_cloud_uri_list`.

bsipocz (approved):
Thank you for the review follow-up, it all looks good now.

bsipocz:
Ahh, RTD failure is related, I'll push a commit for it.

PR #3489, opened 2026-01-07
GAIA: new datalink retrieval type RESIDUAL_IMAGE

cosmoJFH:
Dear Astroquery team, 

happy new year! We want to star this year with a simple PR for Gaia. We would like to define a new datalink retrieval type, RESIDUAL_IMAGE, that will be available for the DR4.

CC @esdc-esac-esa-int 
JIRA GAIAPCR-1368


bsipocz (approved):
Thank you!

PR #3492, opened 2026-01-13
More powerful column selection in `MastMissions`

snbianco:
Make the `select_cols` parameter in `MastMissions` queries more powerful:

- Specify a comma-separated string of columns in addition to a list of columns.
- Specify all available columns with `*` or `all`
- Now validates columns and warns (with suggestions) if a passed column does not exist.

I have also made the `_json_to_table` function more robust so that columns that have values of an unexpected type don't raise an error. Instead, a warning is logged and those particular values are masked.

### DIFF
--- a/astroquery/mast/missions.py
+++ b/astroquery/mast/missions.py
@@ -8,6 +8,7 @@ This module contains methods for searching MAST missions.
 
 import difflib
 import warnings
+from collections.abc import Iterable
 from json import JSONDecodeError
 from pathlib import Path
 from urllib.parse import quote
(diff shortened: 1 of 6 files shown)


astroquery/mast/missions.py
```diff
@@ -197,6 +197,50 @@ def _build_params_from_criteria(self, params, **criteria):
...
+                select_cols = self._default_ullyses_cols
+            return select_cols
+
+        all_columns = self.get_column_list()['name'].value.tolist()
+        if isinstance(select_cols, str) and (select_cols.lower() == 'all' or select_cols == '*'):
+            return all_columns
+
+        if isinstance(select_cols, str):
+            select_cols = select_cols.split(',')
+        elif not isinstance(select_cols, list):
```
bsipocz: nitpicking, but could it be other reasonable iterable, too, e.g. why don't we accept tuples or even a generator?

astroquery/mast/services.py
```diff
@@ -97,13 +97,43 @@ def _json_to_table(json_obj, data_key='data'):
...
         # add the column if it does not exist already
         if col_name not in data_table.colnames:
-            data_table.add_column(MaskedColumn(col_data.astype(col_type), name=col_name, mask=col_mask))
+            try:
+                # Try to coerce entire column at once
+                coerced = col_data.astype(col_type)
+                data_table.add_column(
+                    MaskedColumn(coerced, name=col_name, mask=ignore_mask)
+                )
+            except Exception:
```
bsipocz: should this be a blanket all Exceptions or could we narrow it a bit?

astroquery/mast/services.py
```diff
@@ -97,13 +97,43 @@ def _json_to_table(json_obj, data_key='data'):
...
+                out = np.empty(len(col_data), dtype=col_type)
+                fail_mask = np.zeros(len(col_data), dtype=bool)
+                for i, val in enumerate(col_data):
+                    if val == ignore_value:
+                        # Ignored values are already masked by ignore_mask
+                        continue
+
+                    try:
+                        out[i] = col_type(val)
+                    except Exception:
```
bsipocz: again, this should be done with a very specific exception only

bsipocz (commented):
Overall this looks good, but I wonder if the try/except exceptions can be better specified rather than blanket everything.

snbianco:
Made those changes!

bsipocz (approved):
Thank you!

PR #3493, opened 2026-01-14
Adding to Heasarc model a query_all method

trjaffe:
This is a draft because I don't know how to fix the problems below.  The file changes shown here are incorrect.  

This PR adds a method to Heasarc.  

This method queries all HEASARC tables for a given source, but only in a limited way which allows it to be fast.  It tells you how many matches are found in each catalog.  You then choose a catalog name and query it as usual to fetch the data from within that catalog.  

This PR includes additions to test_heasarc.py and test_heasarc_remote.py.  

Questions for maintainers:  
- Bizarrely, my branch shows my recent commits but this PR does not.  It is stuck on 4 months ago.  What's wrong?  
- in the docs, I see `.. doctest-remote-data::` in places.  Where and why are they needed?  I put none in the new text. 
-  Also, @zoghbi-a mentioned I can use new code to remove duplicates;  where?  

@zoghbi-a @bsipocz

### DIFF
--- a/astroquery/heasarc/core.py
+++ b/astroquery/heasarc/core.py
@@ -816,10 +816,6 @@ class HeasarcClass(BaseVOQuery, BaseQuery):
         >>> result = Heasarc.query_all(position)
         >>> print(result)
 
-        Returns
-        -------
-        table : A `~astropy.table.Table` object.
-
         """
         if position is not None:
             coords_icrs = parse_coordinates(position).icrs


keflavich:
` .. doctest-remote-data::` directives are needed to tell the test runner not to run the tests when an internet connection is unavailable, so these are needed for any tests that are meant for users to run in 'real world' scenarios (i.e., as they'd actually use astroquery).  We have to do this so the tests can pass when servers are down or tests are run without an internet connection.

abdus.txt
bsipocz, line 1: squash out this file from the git history.

astroquery/heasarc/core.py
```diff
@@ -612,6 +613,240 @@ def query_object(self, object_name, mission, *,
...
+            return (
+                np.cos(d) * np.sin(r),
+                np.cos(d) * np.cos(r),
+                np.sin(d)
+            )
+        except ValueError:
+            prefix = ra.split('.')[0]  # e.g., 'a' from 'a.ra'
+            return (f"{prefix}.__x_ra_dec", f"{prefix}.__y_ra_dec", f"{prefix}.__z_ra_dec")
+        except Exception as e:
+            raise e
```
bsipocz, line 641: This is unnecessary
```suggestion
```

astroquery/heasarc/core.py
```diff
@@ -612,6 +613,240 @@ def query_object(self, object_name, mission, *,
         return self.query_region(pos, catalog=mission, spatial='cone',
                                  get_query_payload=get_query_payload)
 
+    def _get_vec(ra=None, dec=None):
```
bsipocz: please use full words for naming methods, variables, etc.

astroquery/heasarc/core.py
```diff
@@ -612,6 +613,240 @@ def query_object(self, object_name, mission, *,
...
+        the table being queried. The latter is used to compute the input position unit
+        vector only once and put the numeric value in the query constraint.
+        """
+        try:
+            r, d = np.radians([float(ra), float(dec)])
+            return (
+                np.cos(d) * np.sin(r),
+                np.cos(d) * np.cos(r),
+                np.sin(d)
+            )
```
bsipocz, line 636: I wonder if we could just rely on some of the astropy.coordinates machinery instead of doing it all by hand?

astroquery/heasarc/core.py
```diff
@@ -612,6 +613,240 @@ def query_object(self, object_name, mission, *,
...
+            where  (  (  a.table_name  =  b.name  )  ) and
+            {constraint_time}
+            group by  b.name , b.description , b.regime , b.mission , b.type
+            order by count desc
+            """
+        # remove all extraneous white space and line breaks
+        return re.sub(r'\s+', ' ', full_query.replace('\n', '')).strip()
+        return full_query
+
+    def query_all(self, position=None, get_query_payload=False, start_time=None,
```
bsipocz, line 755: I think this should be within `query_region` and with a special catalog name e.g. `'all'`? or a new kwarg.

cc @keflavich

bsipocz, line 755: Nitpick: use mandatory keyword arguments for optional parameters
```suggestion
    def query_all(self, *, position=None, get_query_payload=False, start_time=None,
```

astroquery/heasarc/core.py
```diff
@@ -612,6 +613,240 @@ def query_object(self, object_name, mission, *,
...
+                  end_time=None, verbose=False, maxrec=None, radius=None, **kwargs):
```
bsipocz: what is `**kwargs` expected to be? we strongly prefer not using outside of convenience methods that are mapping multiple underlying methods.

also, nitpick as it doesn't matter once we have mandatory kwargs, but the API would be more consistent with a bit of reordering of the parameters (e.g. radius usually goes after position, etc.)

```suggestion
                  end_time=None, verbose=False, maxrec=None, radius=None):
```

tj-test.py
bsipocz, line 1: remove this file from history

astroquery/heasarc/core.py
keflavich: Yes, that's right @bsipocz : `query` or `query_region` with `catalog='all'` or `catalog='**'` or similar would be the equivalent to other modules.

trjaffe:
> I did a quick review here, there are some files added by accident that needs removing and the API could be made a bit more consistent with the rest of the package. Otherwise I would think the PR should pass CI once the remote-data issues are patched over.

Those files were removed ages ago.  The reason you see them is a mystery to me.  But as I said in the initial text, there's something wrong in the PR itself.  If you go look at my repo, it shows a lot of recent edits as I finished this up.  But the PR is showing you a version from 4 months ago, and I have no idea why.  It is a revision that was no where near ready.  So how do I get you/github to see the actual latest revsion?  [Edit:  well, those zombies came back again and I just killed them again.  Sigh.  I blame Visual Studio Code.]

trjaffe:
> ` .. doctest-remote-data::` directives are needed to tell the test runner not to run the tests when an internet connection is unavailable, so these are needed for any tests that are meant for users to run in 'real world' scenarios (i.e., as they'd actually use astroquery). We have to do this so the tests can pass when servers are down or tests are run without an internet connection.

Sorry for the newbie question, but the tests are defined in the astroquery/heasarc/tests/test_heasarc*.py files.  What is connecting them to these docs?   Edit:  oh, does this mean it also grabs code blocks from the docs to test?  I didn't know that.  So I just did a new commit to add the tag above the code block the way that Abdu did it.

bsipocz:
>  Edit: oh, does this mean it also grabs code blocks from the docs to test?

Yes, we do test every code example from the docs, too, after all that documentation is user facing so we need to make sure they keep working as expected.

bsipocz:
> Those files were removed ages ago. The reason you see them is a mystery to me. But as I said in the initial text, there's something wrong in the PR itself. 

The base of your branch seems to be off. I would recommend a rebase on top of the latest `main` of the main repo rather than you fork. I expect that would fix most of these issues (right now there are a lot of merge commits that are bringing in branches at different stages.
I would also recommend a little bit of commit squashes and sanitizing commit messages. This is not something we enforce, but it's usually nice to have logical chunks in the separate commits and cleaning up iterations and frustrations with CI, etc.

astroquery/heasarc/core.py
trjaffe: What you get back is a fundamentally different thing, though.  I would argue it should be its own function.  What you get back is not a catalog.

keflavich: Ah, I see - that was not obvious.  It is specified in the docstring, but not clearly - I'll comment below

astroquery/heasarc/core.py
```diff
@@ -612,6 +613,240 @@ def query_object(self, object_name, mission, *,
...
+        maxrec : int, optional
+            The maximum number of records to return. If None, all matching records are returned up to the server limit.
+        **kwargs : dict, optional
+            Additional keyword arguments:
+
+        Returns
+        -------
+        result : `~astropy.table.Table`
+            A table containing the results of the query. If no results are found,
+            an empty table is returned and a warning is issued.
```
keflavich: This string should include the statement that the resulting table is _not_ a catalog but a list of catalogs.

astroquery/heasarc/core.py
```diff
@@ -612,6 +613,240 @@ def query_object(self, object_name, mission, *,
...
+        --------
+        >>> from astropy.coordinates import SkyCoord
+        >>> from astropy import units as u
+        >>> position = SkyCoord(ra=10.68458, dec=41.26917, unit=(u.degree, u.degree), frame='icrs')
+        >>> result = Heasarc.query_all(position)
+        >>> print(result)
+
+        Returns
+        -------
+        table : A `~astropy.table.Table` object.
```
keflavich: this is a repeat of the above; it's redundant

astroquery/heasarc/core.py
keflavich: "Notes" specifies this, but if I'm looking for what a function returns, I'm generally not looking at Notes, I'm looking at Returns.

astroquery/heasarc/core.py
trjaffe: I was just copying query_region.  At some point I might want to add a way to input specifications that get passed to list_catalogs() and then e.g. query all 300 Chandra catalogs.  But as this isn't currently in use, I'll remove it.

astroquery/heasarc/core.py
bsipocz: then have `list_catalogs` or `query_catalogs` as the method name. `query_all` is just not descriptive enough; so I suppose we need to brainstorm/iterate a few more rounds about it.

keflavich: yes, you're right @bsipocz , this should be renamed `find_catalogs`: Vizier uses that approach.

Vizier also has the more complex "if you're searching for a list of tables, you get a TableList back instead of a table".  I think it would be a good idea to adopt that here; I've found that interface intuitive and we've heard few complaints about it (any? tell me if I'm just forgetting that people hate this approach).

astroquery/heasarc/core.py
trjaffe: The  backend code uses unit vectors where x and y are swapped compared to the ICRS norm used by astropy.coordinates.  I have no idea why except I speculate it has to do with making it easier for PG Sphere operations.  Just a guess.  I've added a note why it doesn't use astropy.

astroquery/heasarc/core.py
trjaffe: Fixed.

trjaffe:
I think we should close and discard this.  Agreed?  I have another clean branch.

PR #3499, opened 2026-01-15
Adding query_all() to heasarc.

trjaffe:
Rebased and squashed after the #3493  fiasco, which we can discard.

This includes comments made there, but the changes were confused so it needs review again.

### DIFF
--- a/docs/mast/mast_obsquery.rst
+++ b/docs/mast/mast_obsquery.rst
@@ -566,6 +566,7 @@ Once the URIs are obtained, they can be used directly in cloud-based workflows o
 use the `~astropy.io.fits.open` function with the S3 URI and appropriate ``fsspec`` keyword arguments.
 
 .. doctest-remote-data::
+.. doctest-requires:: fsspec,s3fs
 
    >>> from astropy.io import fits
    ...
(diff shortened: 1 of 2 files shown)


trjaffe:
@zoghbi-a  I'm adding the extra angular offset column you requested.  Each row in the result is a count of matching rows for that table, and I have added a distance computation into the query.  But that's the distance for every matching row, and so we need to choose an aggregate in the SQL that makes the most sense here, like MAX() or AVG().  MAX makes sense when you do not specify a radius, because that means it's using a different radius for each catalog and you otherwise don't see any indication of that in the results.  If you do specify a radius, then I'm not sure there's anything useful to aggregate about the angular offset.  Did you have a use case in mind that helps?

trjaffe:
@bsipocz  Would you please take another look at this sometime?  I have fixed a bunch of things and learned to love git pre-commit hooks.  There are still some test failures but not in the parts of the code I have changed AFAIK.

bsipocz:
@trjaffe - do you consider this ready for review or still as a draft?

trjaffe:
> @trjaffe - do you consider this ready for review or still as a draft?

I think it's pretty close to done, so no longer a draft, yes, ready for review.

PR #3501, opened 2026-01-16
ESA HST & ISLA: abstract EsaTap using PyVO, HST and ISLA refactored

jespinosaar:
Dear Astroquery team,

As part of the migration to PyVO, we have generated a more general class called EsaTap, extending PyVO capabilities with custom code (authentication, parameters in the request, methods to query tables with specific filters...). We will be using this class to extend ESA modules, so the common code will remain under EsaTap and the mission-specific methods will be defined in their specific modules. As a reference, we have already applied this approach to eHST and ISLA. Please let us know if this is ok with you.

Happy to receive your feedback and implement any change you require!

Kind regards,
@jespinosaar 

cc @esdc-esac-esa-int

### DIFF
--- a/docs/mast/mast_obsquery.rst
+++ b/docs/mast/mast_obsquery.rst
@@ -566,6 +566,7 @@ Once the URIs are obtained, they can be used directly in cloud-based workflows o
 use the `~astropy.io.fits.open` function with the S3 URI and appropriate ``fsspec`` keyword arguments.
 
 .. doctest-remote-data::
+.. doctest-requires:: fsspec,s3fs
 
    >>> from astropy.io import fits
    ...
--- a/setup.cfg
+++ b/setup.cfg
@@ -162,6 +162,7 @@ test=
    pytest-rerunfailures
    fsspec[http]
    moto[s3]
+   s3fs
 docs=
    sphinx
    # https://github.com/astropy/astroquery/issues/3102


jespinosaar:
Ok, after some iterations I managed to have this fixed and compliant with all checks, except for the docs one. But it seems that this is due to an unrelated error. @bsipocz , can you please have a look at it and let me know if further updates are needed? Many thanks!

bsipocz:
Yes, RTD has a known failure, please ignore it for now while I'm fixing it separately.

Otherwise, wow, this PR looks great. I'll come back with a review once I clear some review backlog for the other PRs. I will also try to see how much of the customisation could actually be upstreamed into pyvo itself.

jespinosaar:
> Yes, RTD has a known failure, please ignore it for now while I'm fixing it separately.
> 
> Otherwise, wow, this PR looks great. I'll come back with a review once I clear some review backlog for the other PRs. I will also try to see how much of the customisation could actually be upstreamed into pyvo itself.

Glad to hear that! many thanks and happy weekend!

jespinosaar:
Hi @bsipocz , we have been preparing other missions making use of this new approach and they are almost ready. Is it ok with you if we add them to this PR? or maybe should we wait until this one is merged and then create a new one? They are basic, just extending the EsaTap class and adding two specific methods each, this is why I am asking. Thanks in advance!

jespinosaar:
Wrongly closed, but I will create another PR with more information.

PR #3503, opened 2026-01-16
MAINT: refactor test example to workaround float comparison in string

bsipocz:
We regularly see one of the simbad tests to fail in CI. It all boils down to including float numbers in a string and then doing strict string comparisons on the result. I couldn't really find an elegant way to do this, so double up on the test function and chopped out the floats from the comparison. I would think this is robust enough for testing and practicality beats purity.

### DIFF
--- a/astroquery/simbad/tests/test_utils.py
+++ b/astroquery/simbad/tests/test_utils.py
@@ -1,4 +1,3 @@
-
 import pytest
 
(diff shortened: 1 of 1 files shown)


keflavich (approved):
lgtm.  weird bug, but I see what you did - it's clear and :+1: for pragmatism

bsipocz:
yes, I tried to wrap it in a doctest (as we do have float comparison there) but it still didn't work well, so just fall back to this.

PR #3504, opened 2026-01-16
MAINT: dropping python 3.9 support

bsipocz:
It had EOL already, so time to drop it. I have also updated some other minimal dependencies to drop versions that didn't have python 3.10 support yet.

I was also double checking the diff with pyupgrade but it was full of cruft mostly around f-strings, so I decided not to add that much noise in here.


bsipocz:
I suppose the drops could be a bit more agressive, especially on the astropy/numpy side, but also there isn't much version workarounds baked into the code.

keflavich (approved):

PR #3505, opened 2026-01-21
TEST: Check build with theme updates

Cadair:


bsipocz:
@Cadair - if you want to push for a theme change here, it would be more than welcome!

PR #3514, opened 2026-01-30
EUCLID: new parameters in  `get_observation_products`, `get_product_list`, `get_product `and `get_scientific_product_list ` to retrieve the products by the dataset release

cosmoJFH:
Dear astroquery team,

we would like to upgrade the signature of the methods `get_observation_products`, `get_product_list`, `get_product `and `get_scientific_product_list ` so that they now include the new parameters dsr_part1, dsr_part2, dsr_part3 that can be used to retrieve the products by dataset release (that is made of of 3 parts).

cc @esdc-esac-esa-int 
jira:EUCLIDMNGT-1430

### DIFF
--- a/astroquery/esa/euclid/core.py
+++ b/astroquery/esa/euclid/core.py
@@ -71,6 +71,9 @@ class EuclidClass(TapPlus):
         self.main_table = conf.ENVIRONMENTS[self.environment]['main_table']
         self.main_table_ra = conf.ENVIRONMENTS[self.environment]['main_table_ra_column']
         self.main_table_dec = conf.ENVIRONMENTS[self.environment]['main_table_dec_column']
+        self.dsr_1 = conf.ENVIRONMENTS[self.environment]['data_set_release_part1']
+        self.dsr_2 = conf.ENVIRONMENTS[self.environment]['data_set_release_part2']
+        self.dsr_3 = conf.ENVIRONMENTS[self.environment]['data_set_release_part3']
 
         url_server = conf.ENVIRONMENTS[environment]['url_server']
 
@@ -827,8 +830,9 @@ class EuclidClass(TapPlus):
             files.append(output_file_full_path)
             return files
 
-    def get_observation_products(self, *, id=None, schema="sedm", product_type=None, product_subtype="STK",
-                                 filter="VIS", output_file=None, verbose=False):
+    def get_observation_products(self, *, id=None, release="sedm", product_type=None, product_subtype="STK",
+                                 filter="VIS", dsr_part1=None, dsr_part2=None, dsr_part3=None, output_file=None,
+                                 verbose=False):
         """
         Downloads the products for a given EUCLID observation_id (observations) or tile_index (mosaics)
         For big files the download may require a long time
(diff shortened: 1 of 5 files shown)


astroquery/esa/euclid/core.py
```diff
@@ -837,8 +841,8 @@ def get_observation_products(self, *, id=None, schema="sedm", product_type=None,
         ----------
         id : str, mandatory
             observation identifier (observation id for observations, mosaic id for mosaics)
-        schema : str, optional
-            schema name. Default value is 'sedm'.
+        release : str, optional
+            release name. Default value is 'sedm'.
```
bsipocz, line 845: This will need to be done by an argument rename-deprecation

astroquery/esa/euclid/core.py
```diff
@@ -1214,11 +1320,20 @@ def get_product(self, *, file_name=None, product_id=None, schema='sedm', output_
             is mandatory
         product_id : str, optional, default None
             product id. More than one can be specified between comma. Either file_name or product_id is mandatory
-        schema : str, optional, default 'sedm'
-            the data release name (schema) in which the product should be searched
+        release : str, optional, default 'sedm'
+            the data release name in which the product should be searched
```
bsipocz, line 1324: again, parameter renames should be done with a deprecation

bsipocz (changes requested):
Thank you for the pull request. Most of it looks good, but there are two topics that needs to be addressed before merging  the PR:


- The argument renames should be done with a deprecation grace using the `deprecated_renamed_argument` decorator.

- And running the test suite including the remote tests ( `tox -e test-online -- -P esa.euclid`) I see various failures coming from the documentation, so I expect a couple of the examples in there should be updated, too.
Please fix those, too.

astroquery/esa/euclid/core.py
cosmoJFH: We have reverted the changes and the parameter has the original name `schema`.

astroquery/esa/euclid/core.py
cosmoJFH: We have reverted the changes and the parameter has the original name `schema`.

bsipocz: Just to be clear, you can totally rename this argument, we just need to do it with a deprecation period when we accept both versions.

cosmoJFH: Hi @bsipocz, we made the change by mistake, so, it is not necessary to change the name of this parameter. But it is good to know how to rename the parameter in the future.

bsipocz (approved):
The overall diff looks good, but there are a couple of duplicated commits as well as multiple merge commits, so I go ahead and cleanup those a little bit (and also rebase it on the top of the current `main` as I have some difficulty running the tests locally).

I would expect there is nothing else left from your side, so leave the approval and I go ahead with the merge once the above actions are done with a green CI.

bsipocz:
There are a couple of doctest failures in `docs/esa/euclid/euclid.rst`, but those are unrelated to this PR, so I go ahead with the merge now.

Thank you for the PR!

PR #3517, opened 2026-02-02
Download products from event subscription JSON files

snbianco:
Main Objectives:
- Enhance `MastMissions.download_products` to accept JSON data from the upcoming MAST subscription service (either as a local JSON file or an in-memory object).
- Add full `MastMissions` support for IUE to streamline workflow.

Other Notes:
- More robust error handling when a user tries to input a filename to `MastMissions.download_file` when the mission requires a URI. Unfortunately, I don't know of an easy way to tell which missions require this from the API, so the list is hard-coded (for now).
- `MastMissions.download_products` looks for `dataset` or `fileset` keyword when creating the download path for files. If neither are found, instructs user to use `flat=True`.
- Finally got around to making some changes to `mast_test.py` that simplify mocking and making calls to functions. It looks like a lot of changes, but really, I'm just removing the prepended `mast.` from everywhere it was used.
- We recently had a change to one of our services that was evoking a warning in tests that query the PS1 catalog in `test_mast_remote.py`. I avoid these warnings by selecting certain columns in the tests.

### DIFF
--- a/astroquery/mast/missions.py
+++ b/astroquery/mast/missions.py
@@ -720,10 +720,15 @@ class MastMissionsClass(MastQueryWithLogin):
         base_dir = Path(base_dir)
 
         for data_product in products:
+            col_names = data_product.colnames
             # Determine local path for each file
             filename = data_product['filename']
-            uri = data_product.get('uri', filename)
-            dataset = data_product.get('dataset') or data_product.get('fileset')
+            uri = data_product['uri'] if 'uri' in col_names else filename
+            dataset = None
+            if 'dataset' in col_names:
+                dataset = data_product['dataset']
+            elif 'fileset' in col_names:
+                dataset = data_product['fileset']
             if not dataset and not flat:
                 raise InvalidQueryError('Data product is missing "dataset" or "fileset" field required for '
                                         'constructing local download path. Specify `flat=True` to avoid this '


bsipocz:
devdeps failure is unrelated and has been dealt with separately.

astroquery/mast/missions.py
```diff
@@ -764,11 +783,30 @@ def download_products(self, products, *, download_dir=None, flat=False,
...
+            # Products coming from local JSON filepath from subscription service
+            try:
+                with open(products, 'r') as f:
+                    json_data = json.load(f)
+            except JSONDecodeError as ex:
+                raise InvalidQueryError(f'Failed to decode JSON file at {products}: {ex}')
+
+            if not isinstance(json_data, (list, tuple)):
+                raise InvalidQueryError(f'Expected a list of product rows in JSON file at {products}.')
+            products = Table(rows=json_data)
```
falkben, line 800: we need to deduplicate the json_data since the events coming from the subscriptions database may have many of the same files.

snbianco: That's done later on line 816

falkben: i could be misreading the code, i'm not familiar with this codebase, but does `remove_duplicate_products` only work on uri's? whereas this json data is filesets and datasets?

snbianco: Good question and totally fair to ask! `remove_duplicate_products` actually works on an Astropy `Table` of products and deduplicates based on the key it’s given (`filename`). The JSON is converted into a `Table` earlier in the code, so as long as the JSON format is what we expect, it should work as intended.

PR #3526, opened 2026-02-11
feat: add extra_user_agents to BaseVOQuery

### LINKED ISSUE #3507: Extra user-agents in BaseVOQuery
Hi astroquery maintainers!

We'd need to allow some users (ex: a whole software that depends on astroquery) to add their own user-agent when they use the SIMBAD module. I think this implies a modification to the `BaseVOQuery` class. Would you be ok with such a change?

ManonMarchand:
This PR fixes #3507 

Might not be the best way to do it, I was a bit unsure.

There will be a following PR documenting its use in the `SimbadClass` but I thought that it would be best to separate these changes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -119,6 +119,9 @@ Infrastructure, Utility and Other Changes and Additions
 
 - Versions of numpy <1.22 are no longer supported. [#3504]
 
+- ``BaseVOQuery`` now accepts a ``extra_user_agents`` parameter to allow the addition
+  of user agents on top of astroquery's ones [#3526]
+
 
 utils.tap
 ^^^^^^^^^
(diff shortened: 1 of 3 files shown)


bsipocz (approved):
This looks all good, thanks!

(I'm not sure if you've seen, but Markus also changed the user agent in pyvo; though of course that is not yet usable for astroquery purposes (will be released in pyvo 1.9 and it will take some time for that version to be the minimum requirement)

ManonMarchand:
Thanks for the merge!

I've seen Markus's changes and his suggestions for a standard user agent that'd tell VO services that the current user is doing actual science. We all have huge load issues these days, this might help? :crossed_fingers:

PR #3528, opened 2026-02-11
ENH: SVO FPS - add method to easily retrieve filter metadata

cgobat:
This PR adds a new method: `SvoFpsClass.get_filter_params()`. It follows the same call signature as the existing `SvoFpsClass.get_transmission_data()`, and returns a `dict` containing the VOTable PARAMs returned by the FPS. This includes things like the filter zero point, min/max/effective wavelength, description, etc.

For example, one often needs the effective wavelength of a filter. Without this method, the practical options were either to retrieve the full transmission profile and do the numerical integration yourself, or manually go to the FPS website and find it on the webpage, then hardcode it. With this new method, it is as simple as `lambda_eff = SvoFps.get_filter_params(my_filter_id)["WavelengthEff"]`.

### DIFF
--- a/astroquery/svo_fps/core.py
+++ b/astroquery/svo_fps/core.py
@@ -108,6 +108,41 @@ class SvoFpsClass(BaseQuery):
                 "succeed. Try increasing the timeout limit if a large range is needed."
             )
 
+    def get_filter_params(self, filter_id, cache=True, timeout=None):
+        f"""Get metadata/parameters for the requested Filter ID from SVO
+
+        Parameters
+        ----------
+        filter_id : str
+            Filter ID in the format SVO specifies it: 'facilty/instrument.filter'.
+            This is returned by `get_filter_list` and `get_filter_index` as the
+            ``filterID`` column.
+        cache : bool
+            Defaults to True. If set overrides global caching behavior.
+            See :ref:`caching documentation <astroquery_cache>`.
+        timeout : int
+            Timeout in seconds. If not specified, defaults to {conf.timeout}.
+
+        Returns
+        -------
+        params : dict
+            Dictionary of VOTable PARAM names and values.
+        """
+        query = {'ID': filter_id, 'VERB': 0}
+        response = self._request("GET", self.SVO_MAIN_URL, params=query,
+                                 timeout=timeout or self.TIMEOUT,
+                                 cache=cache)
+        response.raise_for_status()
+        response_content = io.BytesIO(response.content)
+        votable = parse_single_table(response_content)
+        params = {}
+        for param in votable.iter_fields_and_params():
+            if param.unit is not None:
+                params[param.name] = param.value*param.unit
+            else:
+                params[param.name] = param.value
+        return params
+
     def get_transmission_data(self, filter_id, **kwargs):
         """Get transmission data for the requested Filter ID from SVO
 
--- a/astroquery/svo_fps/core.py
+++ b/astroquery/svo_fps/core.py
@@ -108,8 +108,8 @@ class SvoFpsClass(BaseQuery):
                 "succeed. Try increasing the timeout limit if a large range is needed."
             )
 
-    def get_filter_params(self, filter_id, cache=True, timeout=None):
-        f"""Get metadata/parameters for the requested Filter ID from SVO
+    def get_filter_metadata(self, filter_id, *, cache=True, timeout=None):
+        """Get metadata/parameters for the requested Filter ID from SVO
 
         Parameters
         ----------
@@ -121,7 +121,7 @@ class SvoFpsClass(BaseQuery):
             Defaults to True. If set overrides global caching behavior.
             See :ref:`caching documentation <astroquery_cache>`.
         timeout : int
-            Timeout in seconds. If not specified, defaults to {conf.timeout}.
+            Timeout in seconds. If not specified, defaults to ``conf.timeout``.
 
         Returns
         -------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,7 +37,7 @@ Service fixes and enhancements
 svo_fps
 ^^^^^^^
 
-- Add ``get_filter_params`` to allow retrieval of filter metadata. [#3528]
+- Add ``get_filter_metadata`` to allow retrieval of filter metadata. [#3528]
 
 heasarc
 ^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,11 @@ vizier
 Service fixes and enhancements
 ------------------------------
 
+svo_fps
+^^^^^^^
+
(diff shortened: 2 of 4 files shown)


astroquery/svo_fps/core.py
```diff
@@ -108,6 +108,41 @@ def get_filter_index(self, wavelength_eff_min, wavelength_eff_max, **kwargs):
                 "succeed. Try increasing the timeout limit if a large range is needed."
             )
 
+    def get_filter_params(self, filter_id, cache=True, timeout=None):
```
bsipocz:
```suggestion
    def get_filter_params(self, filter_id, *, cache=True, timeout=None):
```

astroquery/svo_fps/core.py
```diff
@@ -108,6 +108,41 @@ def get_filter_index(self, wavelength_eff_min, wavelength_eff_max, **kwargs):
...
+        f"""Get metadata/parameters for the requested Filter ID from SVO
```
bsipocz:
```suggestion
        """Get metadata/parameters for the requested Filter ID from SVO
```

astroquery/svo_fps/core.py
bsipocz: I would favour method names that consists of full words, so please call it `get_filter_metadata` or `get_filter_parameters` instead.

bsipocz (commented):
Thank you. Overall it looks good, but there are a few minor changes that would make it better.

Also, would you mind adding an example for the narrative docs?

astroquery/svo_fps/core.py
cgobat: Got it, makes sense. Funnily enough I actually originally had it as `get_filter_metadata`. but changed it to params because of the VOTable attribute it uses being called `PARAM`s. I'll switch it back to metadata.

astroquery/svo_fps/core.py
cgobat: I had this as an f-string to adaptively refer to the default value `{conf.timeout}`. Should I hardcode that value (60 seconds) in the docstring instead? Or just refer to it literally as `conf.timeout`?

keflavich:
Could you clarify the aim of this?  The suggested motivation is that "Without this method, the practical options were either to retrieve the full transmission profile and do the numerical integration yourself, or manually go to the FPS website and find it on the webpage, then hardcode it. ", but as far as I can tell, this is part of the default returns, e.g.:

```python
from astroquery.svo_fps import SvoFps
filterlist = SvoFps.get_filter_list('JWST')
filterlist[ 'filterID',  'WavelengthEff' ]
```
gives a list like:
```
       filterID        WavelengthEff
                          Angstrom
--------------------- ---------------
    JWST/NIRCam.F070W 6988.4272768359
JWST/NIRCam2025.F070W 6996.0621457337
                  ...             ...
     JWST/MIRI.F2100W 205601.06046254
     JWST/MIRI.F2300C  224493.7622614
     JWST/MIRI.F2550W 251515.98944653
Length = 88 rows
```

keflavich:
The full set of metadata is accessible:
```python
>>> SvoFps.get_filter_list('JWST').colnames
['FilterProfileService',
 'filterID',
 'WavelengthUnit',
 'WavelengthUCD',
 'PhotSystem',
 'DetectorType',
 'Band',
 'Instrument',
 'Facility',
 'ProfileReference',
 'CalibrationReference',
 'Description',
 'Comments',
 'WavelengthRef',
 'WavelengthMean',
 'WavelengthEff',
 'WavelengthMin',
 'WavelengthMax',
 'WidthEff',
 'WavelengthCen',
 'WavelengthPivot',
 'WavelengthPeak',
 'WavelengthPhot',
 'FWHM',
 'Fsun',
 'PhotCalID',
 'MagSys',
 'ZeroPoint',
 'ZeroPointUnit',
 'Mag0',
 'ZeroPointType',
 'AsinhSoft',
 'TrasmissionCurve']
```

then you can select by matching on filterID.  Since the tables of instruments are pretty small, this isn't much of an overhead

("Trasmission Curve" instead of "Transmission Curve" is apparently an upstream issue?)

astroquery/svo_fps/core.py
bsipocz: One feature of astroquery is to remove the VO jargon and replace it with a bit more user friendly description: we look at our API with the eye of astronomer who doesn't know (and I firmly believe shouldn't know) VO internals :)

cgobat:
The biggest problem with that proposal is that a number of filters at SVO's FPS don't have an associated "facility", so the `get_filter_list` method doesn't work at all for them.

Even if that weren't the case, I would still argue that this is a useful addition&mdash;it is a very (maybe even more?) common use case to not actually care about the exact structure of a filter's response curve, and really just need to know some defining characteristics like the FWHM, effective width, pivot wavelength, effective wavelength, etc. Given that, there should be a straightforward, non-hacky way to retrieve that kind of information directly, without having to pull down other filters or data you don't need.

astroquery/svo_fps/core.py
cgobat: Changed to refer to it literally as ` ``conf.timeout`` ` for now

cgobat:
Nice catch about "TrasmissionCurve" though&mdash;I had never noticed that. :)

keflavich:
> The biggest problem with that proposal is that a number of filters at SVO's FPS don't have an associated "facility", so the get_filter_list method doesn't work at all for them.

Is this really true?  The website is listed facility-first; I didn't think there were any filters without an associated facility, even if it's just a placeholder.  Do you have examples?

> Even if that weren't the case, I would still argue that this is a useful addition—it is a very (maybe even more?) common use case to not actually care about the exact structure of a filter's response curve, and really just need to know some defining characteristics like the FWHM, effective width, pivot wavelength, effective wavelength, etc. Given that, there should be a straightforward, non-hacky way to retrieve that kind of information directly, without having to pull down other filters or data you don't need.

I agree, this is a more convenient way to get access to those metadata, which I frequently use.  I don't think the current approach is hacky, but it is a tiny bit more clunky, so I'm onboard with adding this.

cgobat:
Yeah, maybe hacky wasn't quite the right word there. Thanks for your approval.

Here are a couple examples of facility-less filters:
- [Generic/Bessell.V](https://svo2.cab.inta-csic.es/theory/fps/index.php?id=Generic/Bessell.V) (along with most of the "Generic" filters)
- [Cameras/SQM.B](https://svo2.cab.inta-csic.es/theory/fps/index.php?id=Cameras/SQM.B)
- [Misc/UCAC.R](https://svo2.cab.inta-csic.es/theory/fps/index.php?id=Misc/UCAC.R)

bsipocz (approved):
This all looks good to me now; thank you!

bsipocz:
I interpret Adam's comment above as a thumbs up, so let's merge this now.

PR #3531, opened 2026-02-13
Add ESO catalogue functions

ashleythomasbarnes:
We've added functions in the same structure similar to the others to query the ESO catalogues: 

- list_catalogues -> lists all available catalogues 
- query_catalogue -> query results for a specific catalogue with constraints 

Docs are coming in another PR. 

-- Thanks! 
@juanmcloaiza @szampier @almicol


CHANGES.rst
```diff
@@ -13,6 +13,11 @@ noirlab
 API changes
 -----------
 
+eso
+^^^
+
+- Add functionality to list and query ESO catalogues. [#8]
+
```
szampier: I seem to remember that this file is normally edited by the astroquery maintainers

szampier:
General comment: it’s a bit difficult to focus on the new feature since the PR also contains many formatting changes. Splitting these into a separate commit or PR would make the review easier.

CHANGES.rst
bsipocz: Nope, this is correct, preferably the changelog entry is added in the PR by the PR author.

astroquery/eso/tests/setup_package.py
bsipocz, line 1: This file is not needed any more, please remove.

astroquery/eso/__init__.py
```diff
@@ -9,16 +9,21 @@ class Conf(_config.ConfigNamespace):
...
+        'Maximum number of rows allowed by the TAP service.')
     ROW_LIMIT = _config.ConfigItem(
         1000,
-        'Maximum number of rows returned (set to -1 for unlimited).')
+        'Maximum number of rows returned (set to -1 for maximum allowed via TAP service).')
     username = _config.ConfigItem(
         "",
         'Optional default username for ESO archive.')
-    tap_url = _config.ConfigItem(
+    tap_obs_url = _config.ConfigItem(
```
bsipocz, line 21: this is technically an API change; but for now it is acceptable as we haven't yet released the ESO refactor work.

astroquery/eso/core.py
```diff
@@ -32,16 +33,26 @@
...
-    DEFAULT_LEAD_COLS_PHASE3, DEFAULT_LEAD_COLS_RAW
+from .utils import (
+    _UserParams,
+    raise_if_coords_not_valid,
+    _reorder_columns,
+    _raise_if_has_deprecated_keys,
+    _build_adql_string,
+    DEFAULT_LEAD_COLS_PHASE3,
+    DEFAULT_LEAD_COLS_RAW,
+)
```
bsipocz: I really wish these type of style refactorings would not be mixed into feature branches.

Also, we do allow longer lines in astroquery, up to 120 characters, so please don't apply these black type spreads over many lines.

bsipocz (commented):
I've started reviewing this, but the code style changes are a bit overwhelming, so I stopped when I got to the actual changes in core.py. 

Could you please remove any of the style changes (and anyway, please don't force short line length on code) and force push back the actual code changes?

ashleythomasbarnes:
Indeed this should have been a draft PR like we agreed (sorry @szampier)... How should we proceed - shall I revert to draft and make changes, or okay to keep as-in?

ashleythomasbarnes:
Okay, I cleaned up my mess - it should be clearer now. Let me know how you want to proceed...

astroquery/eso/core.py
ashleythomasbarnes: Done.

CHANGES.rst
ashleythomasbarnes: Let me know if you need more here...

ashleythomasbarnes:
Just to let you know @bsipocz  - I spoke with @juanmcloaiza and @szampier this morning, and we're happy for you to continue the review. I will be away from 1st March until ~ 7th April, so it would be great if we could get this wrapped up by then, otherwise @juanmcloaiza has offered to take over any needed changes whilst I'm gone. Thanks!

ashleythomasbarnes:
Sorry @bsipocz, one final commit before I leave tomorrow. This was a very minor one such that @almicol and co can move forward with TAP 1.1 for tap_cap -- still okay now to review when you can. @juanmcloaiza will take over if there is anything urgent while I'm gone.

bsipocz (commented):
Unfortunately I see a lot of test failures in this PR for the remote tests. Could you please have a look and fix them before we do more reviews?

The easiest is to test locally with ` tox -e test-online -- -P eso`

astroquery/eso/core.py
```diff
@@ -279,32 +320,34 @@ def tap(self, authenticated: bool = False) -> TAPService:
...
         ----------
-        query_str : str
+        query : str
             The ADQL query string to be executed.
         authenticated : bool, optional
             If ``True``, the query is run as an authenticated user.
             Authentication must be performed beforehand via
             :meth:`astroquery.eso.EsoClass.login`. Authenticated queries
             may be slower. Default is ``False``.
+        which_tap : {"tap_obs", "tap_cat"}, optional
```
bsipocz: I think this could be named better, more with a descriptor than a question?
 E.g `tap_endpoint`

astroquery/eso/core.py
```diff
@@ -317,7 +360,7 @@ def query_tap(self,
         eso_instance.query_tap("SELECT * FROM ivoa.ObsCore")
         """
         table_to_return = Table()
-        tap_service = self.tap(authenticated)
+        tap_service = self.tap(authenticated, which_tap=which_tap)
         table_to_return = self._try_download_pyvo_table(query, tap_service)
```
bsipocz: I haven't noticed in the previous review, but I think this method should not be called download as we don't really download a file containing the table but keeping everything in memory. (everywhere else `download` kind of indicates some file download functionality.

(doesn't have to be part of this PR,  but would be nice to change it before a release)

astroquery/eso/core.py
```diff
@@ -390,24 +439,66 @@ def list_column(self, table_name: str) -> None:
             log.info(f"\nColumns present in the table {table_name}:\n{available_cols}\n"
                      f"\nNumber of records present in the table {table_name}:\n{num_records}\n")
 
+    @unlimited_maxrec
+    def list_catalogues(self, all_versions: bool = False, cache: bool = True) -> List[str]:
```
bsipocz: make optional params kwarg only
```suggestion
    def list_catalogues(self, *, all_versions: bool = False, cache: bool = True) -> List[str]:
```

bsipocz: and I hate to say this as I love the 'u's and proper spellings, but it's better for the user experience to have a more uniform API. So, please change the method name to `list_catalogs`. 
British spelling can stay in the narrative docs and comments.

astroquery/eso/core.py
```diff
@@ -1052,5 +1143,107 @@ def query_apex_quicklooks(self,
                                   authenticated=authenticated)
         return self._query_on_allowed_values(user_params)
 
+    def query_catalogue(self,
```
bsipocz: same here, please change this to `query_catalog`

bsipocz (commented):
There are some naming and API issues that would be nice to address, otherwise the code looks good.

I also still see 3 test failures due to maxrec warnings, but they are also present on the main branch, so this PR doesn't have to fix them.

bsipocz:
@ashleythomasbarnes @szampier - I'm planning to do an astroquery release around the end of the month, but by mid-May the latest.

Do you think we can wrap this and the documentation PR up by then? I would really love to release the refactored module when most of its follow-up has been also merged.

ashleythomasbarnes:
Yep, I've just returned to the office today but put this high on my priority list for the coming week or so...

ashleythomasbarnes:
Just to keep you updated @bsipocz - We've implemented your suggested changes, and are working on some final checks internally. Once done, I'll push the commit(s) with everything --  this should be done next week.

astroquery/eso/core.py
ashleythomasbarnes: Done in all cases.

astroquery/eso/core.py
ashleythomasbarnes: updated to `_try_retrieve_pyvo_table`

astroquery/eso/core.py
ashleythomasbarnes: Updated to `tap_endpoint` throughout

ashleythomasbarnes:
Now ready for review @bsipocz  - thanks!

ashleythomasbarnes:
Hey @bsipocz  - just wondering if you have an estimate on when this would be reviewed? I will be travelling from 8th May, so would be good to get this wrapped up by then, and also I still need to finalise the docs from https://github.com/astropy/astroquery/pull/3542.

astroquery/eso/core.py
```diff
@@ -267,7 +300,7 @@ def message(query_str):
 
         return table_with_an_extra_row[:self.ROW_LIMIT]
 
-    def tap(self, authenticated: bool = False) -> TAPService:
+    def tap(self, authenticated: bool = False, *, tap_endpoint: str = "tap_obs") -> TAPService:
```
keflavich, line 307: @bsipocz other modules specify the tap endpoint less explicitly and directly, e.g., via `login_url` here: https://github.com/astropy/astroquery/blob/6fa144d510d322ce2746727ee864685989d0744d/astroquery/esa/utils/utils.py#L54

...and maybe are configurable via conf?

I like this approach though.

keflavich (approved):
Looks good to me.  As far as I can tell, the API looks consistent with perhaps one minor inconsistency in TAP specification.

astroquery/eso/core.py
bsipocz: ohh, that docs you link to is not even correct, but that's a separate question.

And in the other modules the self.tap is a properly,  I need to think about it how applicable it's here -- also I'll need to check for examples where there are multiple TAP endpoints based on the data products.

bsipocz (commented):
I see several related test failures, those will need to be fixed.
In the meantime I'll think a bit more about the API regarding the multiple options for the endpoint, what would make sense that also works for other modules.

Also, the doctest are failing, but I suppose we should not care about it in this PR and have https://github.com/astropy/astroquery/pull/3542 to fix it once this is merged.

keflavich:
There's a big error that I want to understand better: running `pytest astroquery/eso/tests/test_eso.py` runs error-free but `pytest astroquery/eso/tests/test_eso.py --remote-data` raises an error: this is very bad, it means `test_eso` is not reliably running non-remote tests.

astroquery/eso/core.py
```diff
@@ -390,24 +438,64 @@ def list_column(self, table_name: str) -> None:
...
+        List available catalogue tables offered by the ESO archive.
+
+        Parameters
+        ----------
+        all_versions : bool, optional
+            If True, list all versions of each catalogue table. If False (default),
+            return only the latest version of each catalogue (as determined by the
+            TAP_CAT metadata query).
+        cache : bool, optional
+            Deprecated and unused. Retained for backwards compatibility.
```
bsipocz: Hmm, I don't exactly get this. This method was not present in previous versions, so no need to introduce a deprecated and unused functionality, there is nothing to patch up for backwards compatibility.

(and if I missed something, then please use @deprecated_renamed_argument so we can appropriate warnings and easier future cleanup)

bsipocz (approved):
Thanks for all the patience. I have one super minor comment left, and that if you could rebase rather than add merge commits in the history.

astroquery/eso/core.py
ashleythomasbarnes: removed cache and open_form, which were copied over from the old queries and indeed not needed --> https://github.com/astropy/astroquery/pull/3531/commits/de02e4d44af216c493470d9df0d75719c7d9a0ed

bsipocz:
Thank you @ashleythomasbarnes and again, thanks for all the patience!

ashleythomasbarnes:
Thank you @bsipocz - and thanks for the patience on your side, this was a bit of learning process for me!

PR #3532, opened 2026-02-14
MAINT: installing all dependencies for docs build

bsipocz:
This has been done for RTD, but not for the tox jobs, so it bubbled up for the cron jobs


bsipocz:
Merging it as none of the PR CI jobs are affected.

PR #3533, opened 2026-02-16
Added `get_query_payload kwarg` to Ogle

nkphysics:
Hey yall!

I added in the `get_query_payload` kwarg into the Ogle module per #2040.
Its a relatively small addition.
For the tests my thought process was to cover all the existing test cases payloads. 
I wasn't sure about the nested lists because of the potential deprecation, but I figured there's no harm in including tests for it since its working right now.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -125,8 +125,7 @@ ogle
 ^^^^
 
 - Added ``get_query_payload`` kwarg to aid in debugging. [#3533]
-- Deprecated support for non-coordinate use in queries. [#3533]
-- Refactor: Moved the entirety of ``OgleClass._args_to_payload`` into ``OgleClass.query_region``. [#3533]
+- Removed support for deprecated non-coordinate use in queries. [#3533]
 
 xmatch
 ^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -121,6 +121,13 @@ linelists
 - General tools for both CDMS/JPL moved to linelists.core [#3456]
 - Added jplspec, moved from its previous location (astroquery.jplspec to astroquery.linelists.jplspec) [#3455]
 
+ogle
+^^^^
+
+- Added ``get_query_payload`` kwarg to aid in debugging. [#3533]
+- Deprecated support for non-coordinate use in queries. [#3533]
+- Refactor: Moved the entirety of ``OgleClass._args_to_payload`` into ``OgleClass.query_region``. [#3533]
+
 xmatch
 ^^^^^^
 
--- a/astroquery/ogle/core.py
+++ b/astroquery/ogle/core.py
@@ -125,7 +125,7 @@ class OgleClass(BaseQuery):
         return files
 
     @prepend_docstr_nosections(_args_to_payload.__doc__)
-    def query_region_async(self, *args, **kwargs):
+    def query_region_async(self, *args, get_query_payload=False, **kwargs):
         """
         Returns
         -------
@@ -133,6 +133,8 @@ class OgleClass(BaseQuery):
             The HTTP response returned from the service.
         """
         files = self._args_to_payload(*args, **kwargs)
+        if get_query_payload:
+            return files
         # Make request
         params = {'dnfile': 'submit'}
         response = self._request("POST", url=self.DATA_URL, data=params,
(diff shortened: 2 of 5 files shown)


keflavich (approved):
Looks good to me.

@nkphysics did anything specific prompt this?  i.e., was there a particular bug you were hunting down that led you to this fix?  I don't mind general PRs to improve the consistency of astroquery modules, but I suspect there was more purpose behind this one.

nkphysics:
Oddly enough there wasn't any particular bug that prompted this. 
I got a few deprecation warnings (this seems to be common but I only recently started dipping into ogle data) when using it which prompted me to start poking around and looking at the source code after `get_query_payload=True` didn't work.

I remembered that last year I came across #2040 when something happened with skyview. I don't quite remember fully what happened with that, but in #3318 I ended up adding `get_query_payload` to skyview, and just figured since I came across this again I'd do the same for ogle.

I went ahead an did a quick audit of which modules have/don't have `get_query_payload` kwargs and dropped the list in #2040.

Since I came across wanting/needing `get_query_payload` twice now, I figure that when I find some spare time I can dedicate to this, I may end up trying to add this into other modules where possible.
No promises, but seems worthwhile.

astroquery/ogle/tests/test_ogle.py
```diff
@@ -61,3 +61,53 @@ def test_ogle_list_values(patch_post):
...
+
+
+def test_ogle_nested_list_payload(patch_post):
+    """
+    Test the payload of multiple pointings using a nested-list of decimal
+    degree Galactic coordinates
+    """
+    co_list = [[0, 0, 0], [3, 3, 3]]
+    expected_payload = '# RD NG GOOD\n0 3\n0 3\n0 3'
+    with pytest.warns(AstropyDeprecationWarning):
```
bsipocz: I appreciate covering this with tests, but overall I think we should instead remove this mode of passing non coordinates altogether; as we added that deprecation warning 6 years ago :)

astroquery/ogle/core.py
```diff
@@ -125,14 +125,16 @@ def _args_to_payload(self, *, coord=None, algorithm='NG', quality='GOOD',
         return files
 
     @prepend_docstr_nosections(_args_to_payload.__doc__)
-    def query_region_async(self, *args, **kwargs):
+    def query_region_async(self, *args, get_query_payload=False, **kwargs):
```
bsipocz: Since this method practically pulls in everything from `_args_to_payload`, I would say the new parameter needs to be added there; along with listing it in the docstring.

Alternatively, if you are up for a bit more cleanup, that method can totally be moved into this one. It is only called once as I rather dislike the obstructing nature of having `*args` and `**kwargs` in the signature along with some docstring hackery.

bsipocz (changes requested):
Thanks!

Overall looks good, but it would be better if the new kwarg was placed into the method where all the other kwargs are handled, so we can have it listed in the docstring.

CHANGES.rst
```diff
@@ -121,6 +121,13 @@ linelists
 - General tools for both CDMS/JPL moved to linelists.core [#3456]
 - Added jplspec, moved from its previous location (astroquery.jplspec to astroquery.linelists.jplspec) [#3455]
 
+ogle
+^^^^
+
+- Added ``get_query_payload`` kwarg to aid in debugging. [#3533]
+- Deprecated support for non-coordinate use in queries. [#3533]
+- Refactor: Moved the entirety of ``OgleClass._args_to_payload`` into ``OgleClass.query_region``. [#3533]
```
bsipocz: There was already a deprecation, this now removed that altogether.

And there is no need to mention internal refactors in the changelog, and `_args_to_payload` was an internal private method.

```suggestion
- Removed support for deprecated non-coordinate use in queries. [#3533]
```

bsipocz (approved):
Thank you! I have one small suggestion and I see one remote test failure which I'll just fix and push in a commit, too so we don't hold up merging.

nkphysics:
Sorry I completely overlooked the docs!

Thanks a bunch @bsipocz!

bsipocz:
No worries. I need to get those remote tests run for PRs, right now it's a bit of a manual process.

PR #3534, opened 2026-02-16
MAST: Enable cloud dataset by default

snbianco:
## Current Behavior
- Access to cloud dataset is not enabled by default.
- Users have to explicitly run the `enable_cloud_dataset`  function before they use any cloud-related functions.

## Proposed Behavior
- The cloud dataset is enabled by default, so users do not have to explicitly call `enable_cloud_dataset`.
- To disable the cloud dataset, users can alter a config value or call the `disable_cloud_dataset` function.

## Advantages
- Faster downloads (in some cases)
- Removes a hidden step that users may not even know is available to them.
- More in line with the direction that MAST is heading re. Roman and other large datasets
- More convenient for users in cloud platforms
- Fewer download failures
- Reduce operational load on MAST servers

## What Changes for Users?
- No longer have to call `enable_cloud_dataset` explicitly
- Cloud downloads will be preferred automatically.
- Files in the cloud will be pulled from there instead of MAST servers.
- This is just default behavior and can be changed.
- Cloud dataset will NOT be enabled if the user does not have the prerequisite packages (`boto3`, `botocore`)
- Users can change whether the cloud dataset is automatically enabled with a configuration variable 

## Other Things to Note
- Cloud access is instantiated lazily only when a relevant method is called (`download_file`, `download_products`, `get_cloud_uri(s)`)
- Fixes a bug introduced in #3488 where `botocore`, an optional package, is imported without a guard. Now, both `boto3` and `botocore` are imported with try blocks and `ImportError` is properly handled.
- `Observations` internally keeps track of whether the cloud dataset is explicitly enabled or disabled. If not explicitly enabled by the user (they call `enable_cloud_dataset`), warning messages are not logged when a product cannot be found in the cloud and the download falls back to on-prem. This is to avoid clogging up the console and introducing warning messages where there were none before.

Close #3546

### DIFF
--- a/astroquery/mast/observations.py
+++ b/astroquery/mast/observations.py
@@ -16,7 +16,6 @@ import numpy as np
 import astropy.units as u
 import astropy.coordinates as coord
 from requests import HTTPError
-from botocore.exceptions import ClientError, BotoCoreError
 from astropy.table import Table, Row, vstack
 from astropy.utils.decorators import deprecated_renamed_argument
 
@@ -34,9 +33,15 @@ from .core import MastQueryWithLogin
 try:
     from botocore.exceptions import ClientError, BotoCoreError
 except ImportError:
-    ClientError = BotoCoreError = ()
+    pass
 
 __all__ = ['Observations', 'ObservationsClass', 'MastClass', 'Mast']
+
+CLOUD_DISABLED_MESSAGE = (
+    'Cloud data access is not enabled. You may be missing prerequisite packages, or cloud access may not be '
+    'enabled by default in module configuration. To enable, try calling the '
+    '`~astroquery.mast.ObservationsClass.enable_cloud_dataset` method.'
+)
 
 
 @async_to_sync
@@ -62,15 +67,19 @@ class ObservationsClass(MastQueryWithLogin):
         """Ensure cloud access is initialized if appropriate."""
         # User explicitly disabled
         if self._cloud_enabled_explicitly is False:
-            return
+            return False
 
         # Already initialized
         if self._cloud_connection is not None:
-            return
+            return True
 
         # Default behavior is to enable cloud access if the config option is set, so we check that here
         if self._cloud_enabled_explicitly is None and conf.enable_cloud_dataset:
             self.enable_cloud_dataset(_internal=True)
+
+        # Return False if cloud access failed to initialize
+        if self._cloud_connection is not None:
+            return True
 
     def _parse_result(self, responses, *, verbose=False):  # Used by the async_to_sync decorator functionality
         """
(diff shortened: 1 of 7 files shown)


astroquery/mast/tests/test_mast.py
```diff
@@ -14,16 +14,20 @@
...
 from astroquery.mast.cloud import CloudAccess
 from astroquery.utils.mocks import MockResponse
 from astroquery.exceptions import (BlankResponseWarning, InvalidQueryError, InputWarning, MaxResultsWarning,
-                                   NoResultsWarning, RemoteServiceError, ResolverError)
+                                   NoResultsWarning, RemoteServiceError, ResolverError, CloudAccessWarning)
+
+try:
+    from botocore.exceptions import ClientError
+except ImportError:
+    ClientError = BotoCoreError = ()
```
bsipocz: Is this version dependent? If yes, do you know when the changes were added?

bsipocz (commented):
Overall this looks good, but it still needs a rebase to resolve the conflicts.

Also, I have one minor comment, not a blocker but maybe we can cleanup a little bit based on the answer.

astroquery/mast/tests/test_mast.py
snbianco: I'm not sure what you mean. I do see that I left `BotoCoreError` in here by mistake, so I'll push a fix for that.

bsipocz: I mean why do we need this? We use the mock that already has the importskip, so I think the try/except should be enough for making the dependency optional (and maybe leave a comment that we do that try/except as this is optional dependency).

Also, in the code itself I think we should not override these errors but patch around the missing boto packages better; e.g. raise a warning sooner or maybe even an exception if the user wants to use cloud but doesn't have the optional dependencies installed.

snbianco: Ah, I see what you're saying. I went back into the code and refactored things a bit. The functions that require cloud access error out if the initialization fails and now give a more descriptive error message. Functions that don't require cloud access to work (downloads) will warn if cloud access can't be established, but fall back to an on-prem download by default.

astroquery/mast/observations.py
```diff
@@ -221,9 +229,16 @@ def enable_cloud_dataset(self, provider="AWS", profile=None, verbose=True, *, _i
             self._cloud_connection = CloudAccess(provider, profile, verbose)
             if not _internal:
                 self._cloud_enabled_explicitly = True
-        except ImportError as e:
-            # boto3 or botocore is not installed
+        except (Exception) as e:
```
bsipocz, line 232: I'm mildly worried about the non-specificity of this, but given the direct usage is covered well, we could have it in as is.

bsipocz (approved):
Thanks for all the iterations! Let's go ahead with the merge now.

PR #3538, opened 2026-02-24
Minor update to include "E" for float conversion in get_headers

ashleythomasbarnes:
Very minor update to allow conversion of string to floats which include "E" in get_headers() . 

That is to avoid error:

`invalid literal for int() with base 10: '5E-05'`


keflavich (approved):

bsipocz (approved):
Thanks!

PR #3539, opened 2026-02-24
Esa gaia issue1325 cone search and query object

Enriquerzrz:
Hi, this is my first contribution to this repository. My name is Enrique Ruiz, and I work at Starion for ESA (SCO-08), at ESAC, mainly involved in development and maintenance of the Survey Data Archives.

This PR updates the Gaia package to enforce .14f formatting for RA/DEC/radius in cone_search and query_object, ensuring consistent ADQL queries across Python versions.

CC: @esdc-esac-esa-int
Jira: GAIAPCR-1325

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -57,6 +57,7 @@ gaia
 
 - New datalink DR4 retrieval type RESIDUAL_IMAGE. [#3489]
 - The method ``load_data`` parses ecsv files [#3500].
+- Fixed decimal precision for query_object and cone_search to use 14 decimal places [#3539].
 
 esa.hubble
 ^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -58,6 +58,8 @@ gaia
 - New datalink DR4 retrieval type RESIDUAL_IMAGE. [#3489]
 - The method ``load_data`` parses ecsv files [#3500].
 - Fixed decimal precision for query_object and cone_search to use 14 decimal places [#3539].
+- Added ``get_query_payload`` kwarg to return the ADQL query string. [#3539]
+
 
 esa.hubble
 ^^^^^^^^^^
--- a/astroquery/gaia/core.py
+++ b/astroquery/gaia/core.py
@@ -518,7 +518,7 @@ class GaiaClass(TapPlus):
                       {row_limit}
                       DISTANCE(
                         POINT('ICRS', {ra_column}, {dec_column}),
-                        POINT('ICRS', {ra}, {dec})
+                        POINT('ICRS', {ra:.14f}, {dec:.14f})
                       ) as dist,
                       {columns}
                     FROM
@@ -528,10 +528,10 @@ class GaiaClass(TapPlus):
                         POINT('ICRS', {ra_column}, {dec_column}),
                         BOX(
                           'ICRS',
-                          {ra},
-                          {dec},
-                          {width},
-                          {height}
+                          {ra:.14f},
+                          {dec:.14f},
+                          {width:.14f},
+                          {height:.14f}
                         )
                       )
                     ORDER BY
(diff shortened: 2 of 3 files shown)


bsipocz:
Welcome Enrique and thank you for your PRs!

astroquery/gaia/tests/test_gaiatap.py
```diff
@@ -517,6 +518,117 @@ def test_query_object_async(column_attrs, mock_querier_async, kwargs):
         assert table[colname].attrs_equal(attrs)
 
 
+def test_query_object_precision_debug(monkeypatch, mock_querier):
+    captured_query = {}
+
+    # Intercept call to launch_job so we can capture the generated query
```
bsipocz: If we need to catch the payload/query for debug/testing/etc purposes I would strongly recommend to add a new kwarg, `get_query_payload`, added to the methods (some of the backgrounds are here: https://github.com/astropy/astroquery/issues/2040)

That would be way more preferable than these long and complicated testing scenarios.

astroquery/gaia/tests/test_gaiatap.py
```diff
@@ -517,6 +518,117 @@ def test_query_object_async(column_attrs, mock_querier_async, kwargs):
...
+    query = captured_query.get("query")
+    assert query is not None
+
+    # Regex to detect floats with exactly 14 decimal digits
+    float14 = r"[0-9]+\.[0-9]{14}"
+    matches = re.findall(float14, query)
+
+    # -----------------------------
+    # Debug: print out everything
+    # -----------------------------
```
bsipocz: I would really prefer to not have this kind of printout added to each test function. The preference for the tests would be to be short and down to the point, here I feel there are a lot more chances for bugs and mistakes in the complicated tests than in the code it suppose to test.

bsipocz (commented):
The changes to the code looks OK. However the tests are overcomplicated and I would instead recommend to add `get_query_payload` to the public API which would/should return the query string instead of sending it to the server. 

That would be then suitable for testing, debuging, and also would help the users.

astroquery/gaia/tests/test_gaiatap.py
Enriquerzrz: Thanks for the suggestion. I've added a get_query_payload keyword to both methods, following the same pattern as you told me. This makes the tests much simpler and avoids the need for complex or fake job setups, while keeping the default behavior unchanged.

astroquery/gaia/tests/test_gaiatap.py
Enriquerzrz: Along with simplifying the tests, I also removed the debug prints

jespinosaar:
Good job @Enriquerzrz ! All the checks are passing. Please have a look at the comments from @bsipocz and let us know if you have any doubt.

jespinosaar:
It seems you already did it 😆

CHANGES.rst
```diff
@@ -57,6 +57,7 @@ gaia
 
 - New datalink DR4 retrieval type RESIDUAL_IMAGE. [#3489]
 - The method ``load_data`` parses ecsv files [#3500].
+- Fixed decimal precision for query_object and cone_search to use 14 decimal places [#3539].
```
bsipocz, line 60:
```suggestion
- Fixed decimal precision for query_object and cone_search to use 14 decimal places [#3539].
- Added ``get_query_payload`` kwarg to return the ADQL query string. [#3539]

```

astroquery/gaia/core.py
```diff
@@ -722,6 +741,9 @@ def cone_search(self, coordinate, *, radius=None,
             if True, the results are saved in a file instead of using memory
         columns: list, optional, default ()
             if empty, all columns will be selected
+        get_query_payload : bool, optional
+            If True, return the SQL query string that would be sent to the server,
```
bsipocz:
```suggestion
            If True, return the ADQL query string that would be sent to the server,
```

astroquery/gaia/core.py
```diff
@@ -641,6 +653,9 @@ def __cone_search(self, coordinate, radius, *, table_name=None,
...
```
bsipocz:
```suggestion
            If True, return the ADQL query string that would be sent to the server,
```

astroquery/gaia/core.py
```diff
@@ -566,13 +574,16 @@ def query_object(self, coordinate, *, radius=None, width=None, height=None, verb
             flag to display information about the process
         columns: list, optional, default ()
             if empty, all columns will be selected
+        get_query_payload : bool, optional
+            If True, return the SQL query string that would be sent to the server,
```
bsipocz:
```suggestion
            If True, return the ADQL query string that would be sent to the server,
```

astroquery/gaia/core.py
```diff
@@ -491,6 +491,9 @@ def __query_object(self, coordinate, *, radius=None, width=None, height=None,
...
```
bsipocz:
```suggestion
            If True, return the ADQL query string that would be sent to the server,
```

bsipocz (approved):
This all looks good, thank you so much for the review follow-up.

 I have very minor comments that I'll just directly commit and then merge the PR.

PR #3540, opened 2026-02-26
Allow multiple search positions in MastMissions

snbianco:
`MastMissions` query functions now support single or multiple targets via `coordinates` and
 `object_names` (including combined use in `query_criteria`).

Notes:
- Multiple coordinates/object names can be provided as an iterable or a comma-separated string.
- The legacy `objectname` keyword is deprecated in favor of `object_names`.
- `query_criteria` no longer requires that a non-positional criterion be provided.
- Refactored so that `query_region` and `query_object` call `query_criteria`. Docs now emphasize that `query_region` and `query_object` are convenience wrappers.

### DIFF
--- a/astroquery/mast/missions.py
+++ b/astroquery/mast/missions.py
@@ -15,15 +15,16 @@ from pathlib import Path
 from urllib.parse import quote
 
 import astropy.units as u
-import astropy.coordinates as coord
+from astropy.coordinates import SkyCoord, BaseCoordinateFrame, Angle
 import numpy as np
 from astropy.table import Table, Row, Column, vstack
+from astropy.utils.decorators import deprecated_renamed_argument
 from requests import HTTPError, RequestException
 
 from astroquery import log
 from astroquery.utils import commons, async_to_sync
 from astroquery.utils.class_or_instance import class_or_instance
-from astroquery.exceptions import InputWarning, InvalidQueryError, MaxResultsWarning, NoResultsWarning
+from astroquery.exceptions import InputWarning, InvalidQueryError, MaxResultsWarning, NoResultsWarning, ResolverError
 
 from astroquery.mast import utils
 from astroquery.mast.core import MastQueryWithLogin
@@ -49,8 +50,11 @@ class MastMissionsClass(MastQueryWithLogin):
                              'spectral_type', 'bmv0_mag', 'u_mag', 'b_mag', 'v_mag', 'gaia_g_mean_mag', 'star_mass',
                              'instrument', 'grating', 'filter', 'observation_id']
 
-    # maximum supported query radius
+    # Maximum supported query radius
     _max_query_radius = 30 * u.arcmin
+
+    # Maximum number of input targets accepted in a single query
+    _max_input_targets = 100
 
     def __init__(self, *, mission='hst', mast_token=None):
         super().__init__(mast_token=mast_token)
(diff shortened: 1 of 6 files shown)


astroquery/mast/tests/test_mast.py
```diff
@@ -350,6 +343,55 @@ def test_missions_parse_select_cols(patch_post):
...
+    coord = SkyCoord(10.684, 41.269, unit='deg')
+    result = MastMissions._parse_multiple_targets(coordinates=coord)
+    assert result == ['10.684 41.269']
+
+    # Single coordinate string
+    result = MastMissions._parse_multiple_targets(coordinates="10.684 41.269")
+    assert result == ['10.684 41.269']
+
+    # List of coordinate objects
+    coords = [SkyCoord(10.684, 41.269, unit='deg'), SkyCoord(83.6331, 22.0145, unit='deg')]
```
bsipocz, line 357: I suppose this would also work with a vector SkyCoord as opposed to a list of singletons?

docs/mast/mast_missions.rst
```diff
@@ -63,116 +83,90 @@ Keyword arguments can also be used to refine results further. The following para
...
-    search_pos     sci_data_set_name   sci_targname         sci_start_time             sci_stop_time              ang_sep        sci_status
-   ------------------ ----------------- ---------------- -------------------------- -------------------------- -------------------- ----------
-   210.80227 54.34895         OBQU01050 NUCLEUS+HODGE602 2012-05-24T07:51:40.553000 2012-05-24T07:54:46.553000 0.017460048037303017     PUBLIC
-   210.80227 54.34895         OBQU010H0 NUCLEUS+HODGE602 2012-05-24T09:17:38.570000 2012-05-24T09:20:44.570000 0.017460048037303017     PUBLIC
-   210.80227 54.34895         OBQU01030 NUCLEUS+HODGE602 2012-05-24T07:43:20.553000 2012-05-24T07:46:26.553000 0.022143836477276503     PUBLIC
-   210.80227 54.34895         OBQU010F0 NUCLEUS+HODGE602 2012-05-24T09:09:18.570000 2012-05-24T09:12:24.570000 0.022143836477276503     PUBLIC
-   210.80227 54.34895         OBQU01070 NUCLEUS+HODGE602 2012-05-24T08:00:00.553000 2012-05-24T08:03:06.553000  0.04381046755938432     PUBLIC
+   >>> select_cols = ["sci_targname", "sci_pep_id", "sci_status"]
+   >>> results = missions.query_criteria(coordinates=[SkyCoord(245.89675, -26.52575, unit='deg'), "205.54842 28.37728"],
```
bsipocz, line 120: This example errors out when remote tests are run, could you please have a look at it?

astroquery/mast/missions.py
```diff
@@ -265,91 +269,88 @@ def _parse_select_cols(self, select_cols):
...
+            for name in object_name_items:
+                sc = resolved if isinstance(resolved, SkyCoord) else resolved.get(name)
+                if sc:
+                    targets.append(f"{sc.ra.deg} {sc.dec.deg}")
+
+        return targets
 
     @class_or_instance
-    def query_criteria_async(self, *, coordinates=None, objectname=None, radius=3*u.arcmin,
+    @deprecated_renamed_argument('objectname', 'object_names', since='0.4.11')
```
bsipocz, line 341: there are a couple more of these `objectname` kwargs, would you mind changing them to the snake case, too? (not in this PR, but sometime, for the sake of API consistency)

bsipocz (commented):
Some comments, the documentation example needs to be fixed otherwise it looks good.

Thanks!

astroquery/mast/tests/test_mast.py
snbianco: It does not! Good catch. I'll make that change.

astroquery/mast/missions.py
snbianco: Yes, this is in our plans! Catalogs will change with the new version, and I can make another PR for Observations and Cutouts.

bsipocz (approved):
Thank you, this all looks good now!

PR #3543, opened 2026-02-27
EUCLID: include the new parameter linking_parameter in the method get_spectrum

cosmoJFH:
Dear astroquery team,

we would like to update the method _get_spectrum_ so that the spectra can be searched by a _source_id_ (present implementation) or by a _sourcepatch_id_. To carry out this, we introduce the parameter linking_parameter that provides the meaning to the input id of the source.

Note that we have changed the input parameter source_id to id.

cc @esdc-esac-esa-int 
jira: EUCLIDSWR-226

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,8 @@ esa.euclid
 - Method ``get_product`` now supports the input file_name as a Python list (e.g. ["file1.fits", "file2.fits"]) while
   still accepting the original comma separated string format. [#3541]
 - update the output filename for downloads in the methods get_product and get_cutout [#3550]
+- The method ``get_spectrum`` accepts the new parameter ``linking_parameter`` to retrieve the spectra by source_id and
+  sourcepatch_id [#3543]
 
 vizier
 ^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,7 +27,8 @@ esa.euclid
   still accepting the original comma separated string format. [#3541]
 - update the output filename for downloads in the methods get_product and get_cutout [#3550]
 - The method ``get_spectrum`` accepts the new parameter ``linking_parameter`` to retrieve the spectra by source_id and
-  sourcepatch_id [#3543]
+  sourcepatch_id. [#3543]
+-  The ``source_id`` kwarg in the ``get_spectrum`` method has been renamed to ``ids``. [#3543]
 
 vizier
 ^^^^^^
--- a/astroquery/esa/euclid/core.py
+++ b/astroquery/esa/euclid/core.py
@@ -18,6 +18,7 @@ from astropy import units
 from astropy import units as u
 from astropy.coordinates import Angle
 from astropy.units import Quantity
+from astropy.utils import deprecated_renamed_argument
 from requests.exceptions import HTTPError
 
 from astroquery import log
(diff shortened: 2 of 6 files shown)


astroquery/esa/euclid/core.py
```diff
@@ -1453,7 +1452,8 @@ def get_cutout(self, *, file_path=None, instrument=None, id=None, coordinate, ra
 
         return files
 
-    def get_spectrum(self, *, source_id, schema='sedm', retrieval_type="ALL", output_file=None, verbose=False):
+    def get_spectrum(self, *, id, schema='sedm', retrieval_type="ALL", linking_parameter='SOURCE_ID',
```
bsipocz: Unfortunately this needs to be done through an argument rename deprecation

(see usage of `@deprecated_renamed_argument` in e.g. the simbad module)

astroquery/esa/euclid/core.py
```diff
@@ -1465,13 +1465,17 @@ def get_spectrum(self, *, source_id, schema='sedm', retrieval_type="ALL", output
 
         Parameters
         ----------
-        source_id : str, mandatory, default None
-            source id for the spectrum
+        id : str or int, mandatory
```
bsipocz: Any change that this method will be able to handle multiple IDs at the same time? I'm just asking as `get_datalinks` below can and as we already rename the argument we can make it a bit more self consistent and call this one `ids`, too.

bsipocz (changes requested):
Overall this looks good, but the argument rename will need to go through a deprecation. (it should be straightforward with the decorator though). 

This also needs a changelog entry.

astroquery/esa/euclid/core.py
cosmoJFH: Yes, you are right. We have included it

    @deprecated_renamed_argument('source_id', 'id',since='0.4.12')

astroquery/esa/euclid/core.py
cosmoJFH: Yes, we have a task to update this method to handle multiple IDs at the same time, but this will part  of our next sprint. So, would it be convenient to to change source_id by ids now. instead of of id? Or better to do it in the PR where the changes would be implemented?

bsipocz: I would say it's better to change it now, so we don't need to change it again when the functionality is extended.

cosmoJFH: We have changed the parameter source_id  to ids.

CHANGES.rst
```diff
@@ -21,13 +21,10 @@ esa.euclid
...
   ``cone_search`` is changed to "votable_gzip". [#3497]
-
-
 - Methods ``cone_search`` and ``cross_match_basic`` now define the  parameters ``table_name`` and ``ra_column_name`` and
   ``dec_column_name`` independently [#3496]
-
-- Method ``get_product`` now supports the input file_name as a Python list (e.g. ["file1.fits", "file2.fits"]) while 
-  still accepting the original comma separated string format. [#3541]
+- The method ``get_spectrum`` accepts the new parameter ``linking_parameter`` to retrieve the spectra by source_id and
+  sourcepatch_id [#3543]
```
```suggestion
- The method ``get_spectrum`` accepts the new parameter ``linking_parameter`` to retrieve the spectra by source_id and
  sourcepatch_id. [#3543]
-  The ``source_id`` kwarg in the ``get_spectrum`` method has been renamed to ``ids``. [#3543]
```

bsipocz (approved):
Thank you! Follow-up changes look good, one minor point about mentioning the kwarg rename in the changelog. I'll add that directly, rebase to fix the conflicts, and will merge this PR.

PR #3545, opened 2026-03-03
Add get_zeropoint and more flexible metadata querying to SVO FPS

keflavich:
There is metadata on SVO FPS that is inaccessible but should be.

https://github.com/astropy/astroquery/pull/3528 showed the right mechanism to retrieve metadata, but it didn't allow querying on PhotCalID - while `data_from_svo` allowed specification of this parameter, the mechanism we are using to parse the votable drops the relevant data.

This PR adds a new `get_zeropoint` mechanism to very explicitly and cleanly retrieve zeropoints and also generalizes `get_filter_metadata`.

Note that I have made some serious mistakes in recent publications because I failed to realize the default zeropoint was vega, so it's high importance for me to expose the `get_zeropoint` method with an explicit `mag_sys` keyword to users.

### DIFF
--- a/astroquery/svo_fps/core.py
+++ b/astroquery/svo_fps/core.py
@@ -23,6 +23,11 @@ for suffix in ("_min", "_max"):
 QUERY_PARAMETERS.update(("Instrument", "Facility", "PhotSystem", "ID", "PhotCalID",
                          "FORMAT", "VERB"))
 
+ALLOWED_QUERY_PARAMETERS = {
+    "VERB": {0, 1, 2},
+    "FORMAT": {"metadata", None}
+}
+
 
 class SvoFpsClass(BaseQuery):
     """
@@ -31,21 +36,88 @@ class SvoFpsClass(BaseQuery):
     SVO_MAIN_URL = conf.base_url
     TIMEOUT = conf.timeout
 
-    def data_from_svo(self, query, *, cache=True, timeout=None,
-                      error_msg='No data found for requested query'):
+    def data_from_svo(self,
+                      *,
+                      WavelengthRef_min=None,
+                      WavelengthRef_max=None,
+                      WavelengthMean_min=None,
+                      WavelengthMean_max=None,
+                      WavelengthEff_min=None,
+                      WavelengthEff_max=None,
+                      WavelengthMin_min=None,
+                      WavelengthMin_max=None,
+                      WavelengthMax_min=None,
+                      WavelengthMax_max=None,
+                      WidthEff_min=None,
+                      WidthEff_max=None,
+                      FWHM_min=None,
+                      FWHM_max=None,
+                      Instrument=None,
+                      Facility=None,
+                      PhotSystem=None,
+                      ID=None,
+                      PhotCalID=None,
+                      FORMAT=None,
+                      VERB=2,
+                      cache=True, timeout=None,
+                      error_msg='No data found for requested query',
+                      ):
         """Get data in response to the query send to SVO FPS.
         This method is not generally intended for users, but it can be helpful
         if you want something very specific from the SVO FPS service.
         If you don't know what you're doing, try `get_filter_index`,
         `get_filter_list`, and `get_transmission_data` instead.
 
+        Description of search parameters can be found at
+        https://svo2.cab.inta-csic.es/theory/fps/index.php?mode=voservice
+
+
         Parameters
         ----------
-        query : dict
-            Used to create a HTTP query string i.e. send to SVO FPS to get data.
-            In dictionary, specify keys as search parameters (str) and
-            values as required. Description of search parameters can be found at
-            https://svo2.cab.inta-csic.es/theory/fps/index.php?mode=voservice
+        WavelengthRef_min : float, optional
+            Min value for WavelengthRef parameter
+        WavelengthRef_max : float, optional
+            Max value for WavelengthRef parameter
+        WavelengthMean_min : float, optional
+            Min value for WavelengthMean parameter
+        WavelengthMean_max : float, optional
+            Max value for WavelengthMean parameter
+        WavelengthEff_min : float, optional
+            Min value for WavelengthEff parameter
+        WavelengthEff_max : float, optional
+            Max value for WavelengthEff parameter
+        WavelengthMin_min : float, optional
+            Min value for WavelengthMin parameter
+        WavelengthMin_max : float, optional
+            Max value for WavelengthMin parameter
+        WavelengthMax_min : float, optional
+            Min value for WavelengthMax parameter
+        WavelengthMax_max : float, optional
+            Max value for WavelengthMax parameter
+        WidthEff_min : float, optional
+            Min value for WidthEff parameter
+        WidthEff_max : float, optional
+            Max value for WidthEff parameter
+        FWHM_min : float, optional
+            Min value for FWHM parameter
+        FWHM_max : float, optional
+            Max value for FWHM parameter
+        Instrument : str, optional
+            Instrument for filters (default is None). Leave empty if there are no instruments for specified facility
+        Facility : str, optional
+            Facility for filters (default is None)
+        PhotSystem : str, optional
+            Photometric system for filters (default is None)
+        ID : str, optional
+            Filter ID (default is None)
+        PhotCalID : str, optional
+            Photometric calibration ID (default is None)
+        FORMAT : str, optional
+            Format of the output.  Default includes all data, ``metadata`` includes only metadata.
+        VERB : 0, 1, or 2
+            0: The resulting VOTable won't include the transmission curve or PARAM descriptions.
+            1: The resulting VOTable won't include the transmission curve but it will include PARAM descriptions.
+            2: The resulting VOTable will include the transmission curve and PARAM descriptions.
         error_msg : str, optional
             Error message to be shown in case no table element found in the
             responded VOTable. Use this to make error message verbose in context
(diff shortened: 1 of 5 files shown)


cgobat (approved):
Nice. I like it.

astroquery/svo_fps/core.py
```diff
@@ -108,7 +108,7 @@ def get_filter_index(self, wavelength_eff_min, wavelength_eff_max, **kwargs):
                 "succeed. Try increasing the timeout limit if a large range is needed."
             )
 
-    def get_filter_metadata(self, filter_id, *, cache=True, timeout=None):
+    def get_filter_metadata(self, filter_id, *, cache=True, timeout=None, **kwargs):
```
bsipocz, line 253: can we be more specific about the acceptable kwargs? Or is the list very long?

(I try to get rid of these wildcarded kwargs as they swallow a lot of problems as users can pass on practically anything even if it has no effect; besides it means that our documentation is practically useless.)

bsipocz (approved):
The changes look good, and tests are happy.

However, I'm less happy about introducing back the `**kwarg` approach. Could we maybe expand it out and list all the options explicitly? If nothing else, it would make the docstring more useful.

astroquery/svo_fps/core.py
bsipocz: I would still think we should spell these out; if for nothing else to show the users of what parameters are available.

keflavich: the kwargs are specifically query args that then go through validation
```
query.update(kwargs)
bad_params = [param for param in query if param not in QUERY_PARAMETERS]
```
we could refactor to put all these as explicit parameters:
```
# Valid query parameters taken from
# https://svo2.cab.inta-csic.es/theory/fps/index.php?mode=voservice
_params_with_range = {"WavelengthRef", "WavelengthMean", "WavelengthEff",
                      "WavelengthMin", "WavelengthMax", "WidthEff", "FWHM"}
QUERY_PARAMETERS = _params_with_range.copy()
for suffix in ("_min", "_max"):
    QUERY_PARAMETERS.update(param + suffix for param in _params_with_range)
QUERY_PARAMETERS.update(("Instrument", "Facility", "PhotSystem", "ID", "PhotCalID",
                         "FORMAT", "VERB"))
```
would you prefer that?   It's 21 parameters, but maybe it's more helpful to users to have them all explicitly specified?

bsipocz: I like to think that it's useful to have the list available in the signature and docstring rather then sending them to an outside link or the source code to figure it all out.

keflavich: I agree.  OK, I'll do that.

astroquery/svo_fps/core.py
```diff
@@ -31,21 +36,88 @@ class SvoFpsClass(BaseQuery):
...
+        Facility : str, optional
+            Facility for filters (default is None)
+        PhotSystem : str, optional
+            Photometric system for filters (default is None)
+        ID : str, optional
+            Filter ID (default is None)
+        PhotCalID : str, optional
+            Photometric calibration ID (default is None)
+        FORMAT : str, optional
+            Format of the output.  Default includes all data, `metadata` includes only metadata.
```
bsipocz:
```suggestion
            Format of the output.  Default includes all data, ``metadata`` includes only metadata.
```

astroquery/svo_fps/core.py
```diff
@@ -122,13 +222,26 @@ def get_filter_metadata(self, filter_id, *, cache=True, timeout=None):
             See :ref:`caching documentation <astroquery_cache>`.
         timeout : int
             Timeout in seconds. If not specified, defaults to ``conf.timeout``.
+        kwargs : dict
+            Appended to the ``query`` dictionary sent to SVO.
```
bsipocz:
```suggestion
            Appended to the ``query`` dictionary sent to SVO. See the API documentation of `data_from_svo` for the valid parameter names.
```

astroquery/svo_fps/core.py
```diff
@@ -143,6 +256,53 @@ def get_filter_metadata(self, filter_id, *, cache=True, timeout=None):
...
+        Parameters
+        ----------
+        filter_id : str
+            Filter ID in the format SVO specifies it: 'facilty/instrument.filter'.
+            This is returned by `get_filter_list` and `get_filter_index` as the
+            ``filterID`` column.
+        mag_system : str
+            The magnitude system for which to return the zero point.
+        kwargs : dict
+            Appended to the ``query`` dictionary sent to SVO.
```
bsipocz:
```suggestion
            Appended to the ``query`` dictionary sent to SVO. See the API documentation of `data_from_svo` for the valid parameter names.
```

bsipocz (approved):
Suggestion for some more docs, and the fix for the RTD failure is in the comments. Also, please rebase to get rid of the merge commit.

keflavich:
I think the remaining error here is a temporary service access one from rtd, so this should be good for final review.

astroquery/svo_fps/tests/test_svo_fps.py
```diff
@@ -84,10 +90,18 @@ def test_get_filter_list(patch_get):
...
+
 def test_invalid_query(patch_get):
-    msg = r"^parameter bad_param is invalid\. For a description of valid query "
-    with pytest.raises(InvalidQueryError, match=msg):
-        SvoFps.data_from_svo(query={"bad_param": 0, "FWHM": 20})
-    msg = r"^parameters invalid_param, bad_param are invalid\. For a description of "
-    with pytest.raises(InvalidQueryError, match=msg):
-        SvoFps.data_from_svo(query={"invalid_param": 0, 'bad_param': -1})
+    msg = r"^SvoFpsClass.data_from_svo\(\) got an unexpected keyword argument 'bad_param'"
+    with pytest.raises(TypeError, match=msg):
```
bsipocz: Why did you change the error type checks here? we should still raise InvalidQueryError for invalid parameters.

astroquery/svo_fps/core.py
```diff
@@ -122,13 +222,38 @@ def get_filter_metadata(self, filter_id, *, cache=True, timeout=None):
...
+        query.update(kwargs)
+
+        bad_params = [param for param in query if param not in QUERY_PARAMETERS]
+        if bad_params:
+            raise InvalidQueryError(
+                f"parameter{'s' if len(bad_params) > 1 else ''} "
+                f"{', '.join(bad_params)} {'are' if len(bad_params) > 1 else 'is'} "
+                f"invalid. For a description of valid query parameters see "
+                "https://svo2.cab.inta-csic.es/theory/fps/index.php?mode=voservice"
+            )
```
bsipocz: this is fine, but maybe you can in addition or instead reference the `data_from_svo` API as it lists all of them explicitly.

astroquery/svo_fps/core.py
```diff
@@ -143,6 +268,54 @@ def get_filter_metadata(self, filter_id, *, cache=True, timeout=None):
                 params[param.name] = param.value
         return params
 
+    def get_zeropoint(self, filter_id, mag_system='Vega', **kwargs):
```
bsipocz:
```suggestion
    def get_zeropoint(self, filter_id, *, mag_system='Vega', **kwargs):
```

astroquery/svo_fps/core.py
```diff
@@ -143,6 +268,54 @@ def get_filter_metadata(self, filter_id, *, cache=True, timeout=None):
...
+        ``Vega`` as the default system.
+
+        Parameters
+        ----------
+        filter_id : str
+            Filter ID in the format SVO specifies it: 'facilty/instrument.filter'.
+            This is returned by `get_filter_list` and `get_filter_index` as the
+            ``filterID`` column.
+        mag_system : str
+            The magnitude system for which to return the zero point.
```
bsipocz, line 317: so we have a list of valid options? it maybe useful, but also it's OK to refer to the server here.

bsipocz (changes requested):
The only actual issue is about testing for the correct exception type we raise, once fixed we should merge this.

astroquery/svo_fps/tests/test_svo_fps.py
keflavich: ahh.... the match was supposed to change, not the class.  Good catch.

astroquery/svo_fps/core.py
keflavich: I'll revisit this - I think we're not expecting `bad_params` to be possible any more...

astroquery/svo_fps/core.py
keflavich: no, this should be vega or ab. I'll specify it.

astroquery/svo_fps/tests/test_svo_fps.py
keflavich: right, there is a real reason: with the refactor, you can't get to `InvalidQueryError` any more - it's not a reachable state.  I'll rethink this.

keflavich: I revised this test and added a new one.

astroquery/svo_fps/core.py
```diff
@@ -31,21 +36,88 @@ class SvoFpsClass(BaseQuery):
...
+                      WidthEff_max=None,
+                      FWHM_min=None,
+                      FWHM_max=None,
+                      Instrument=None,
+                      Facility=None,
+                      PhotSystem=None,
+                      ID=None,
+                      PhotCalID=None,
+                      FORMAT=None,
+                      VERB=2,
```
bsipocz: Now that I'm doing a final go through, I noticed that these are cases all over the place. Could you change all of them to be lower case so the basic astroquery API design is not overriden? We already do a mapping, so it shouldn't really affect code complexity

bsipocz (changes requested):
Ouch, this is really close, but I think the the parameter cases should be fixed before merging it.

astroquery/svo_fps/core.py
keflavich: This matches the documentation on their website (https://svo2.cab.inta-csic.es/theory/fps/index.php?mode=voservice).  I'm OK with switching everything to lowercase everywhere if you prefer, but matching the URLs is easier and maybe more intuitive for users?

bsipocz: I like to think that users would read our documentation and docstrings foremost rather than directly want to match webforms, that's the whole point, right?

(also, I haven't checked, but does the server case sensitive to begin with?)

keflavich: it is not case-sensitive!  so I will change all to lowercase.  I think the underscores are ok but I'll review

keflavich: ahhh no I was wrong, it _is_ case sensitive, but it still returns _something_ if you get the case wrong.

bsipocz (approved):
OK, this looks good now. I'll go ahead and rebase: 1) smoke out the test failure, that is unrelated but I haven't seen elsewhere 2) get rid of the merge commit 3) squash out the style fixes, including a new commit that fixes the failure

bsipocz:
RTD failure is unrelated and looks to be temporary server side outage, so I go ahead with the merge here. Thanks @keflavich!

PR #3547, opened 2026-03-04
JPLSpec is even more down

keflavich:
This is a harder workaround for jplspec being down: it switches to using the wayback machine instead of jpl's server, which has gone totally unresponsive.

This raises bigger questions about data preservation/curation.  Can't answer those here, but here's a 🩹 .

### DIFF
--- a/astroquery/linelists/jplspec/core.py
+++ b/astroquery/linelists/jplspec/core.py
@@ -156,9 +156,19 @@ class JPLSpecClass(BaseQuery):
         by the query service.  This workaround is needed while JPLSpec's query
         tool is broken.
 
-        use_getmolecule is an option to force the query to use get_molecule.
-        It is needed if the JPL server is completely unresponsive.
+        use_getmolecule is an option to skip the query service entirely and
+        retrieve full molecule catalogs via `get_molecule`. It is needed when
+        the JPL query server is unresponsive. Frequency and strength limits
+        are not applied in this mode.
         """
+        if use_getmolecule:
+            if get_query_payload:
+                return [('Mol', tuple(self._resolve_molecules(molecule, flags=flags,
+                                                              parse_name_locally=parse_name_locally)))]
+            mols = self._resolve_molecules(molecule, flags=flags,
+                                           parse_name_locally=parse_name_locally)
+            return self._build_table_from_get_molecule(mols)
+
         response = self.query_lines_async(min_frequency=min_frequency,
                                           max_frequency=max_frequency,
                                           min_strength=min_strength,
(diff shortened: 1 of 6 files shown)


bsipocz:
This needs a changelog. And maybe separating out into two PRs, one for the backup url (I expect that to get merged quickly); and the other for the new kwarg (if it needs more work)?

keflavich:
The backup doesn't do anything without the new kwarg, so these should go together.  I will fix the failures tonight - this was an emergency patch with my student, since she needs access now.

astroquery/linelists/jplspec/core.py
```diff
@@ -153,6 +155,9 @@ def query_lines(self, min_frequency, max_frequency, *,
         governs whether `get_molecule` will be used when no results are returned
         by the query service.  This workaround is needed while JPLSpec's query
         tool is broken.
+
+        use_getmolecule is an option to force the query to use get_molecule.
```
bsipocz, line 159: outside of scope here, but these methods should have proper docstrings

bsipocz (approved):
jplspec seems to be back right now, do you still want to go ahead with the wayback machine addition?

keflavich:
are you sure it's back?  If you run a query, you probably get `Zero lines were found for your search criteria.` for all inputs

bsipocz:
> are you sure it's back? If you run a query, you probably get Zero lines were found for your search criteria. for all inputs

All tests pass locally on `main`, and  https://spec.jpl.nasa.gov/ftp/pub/catalog also works? (Testing from home network and not from within caltech)

keflavich:
Huh, you're right.  I'll dig a little more, maybe I can default the fallback to 'off'.

bsipocz:
@keflavich - can you look at the test failures?

keflavich:
Yep, I'll dig in - I clearly (re?)introduced a bug.

keflavich:
I just got an e-mail from the JPL folks saying they had fixed the server!  I'll try to verify that later.

astroquery/linelists/jplspec/core.py
```diff
@@ -170,6 +185,39 @@ def query_lines(self, min_frequency, max_frequency, *,
...
+            if len(mols) == 0:
+                raise InvalidQueryError('No matching species found.')
+            return mols
+        if isinstance(molecule, (list, tuple)):
+            return list(molecule)
+        return [molecule]
+
+    def _build_table_from_get_molecule(self, mols):
+        """Fetch full catalog tables for each molecule and combine them."""
+        self.lookup_ids = build_lookup()
```
bsipocz, line 211: couldn't/shouldn't you reuse `_resolve_molecules` here, or vice versa? I just feel some of the logic is repeated.

bsipocz (approved):
overall looks good. There is one remote test failure though with the documentation, if you could also fix that up before merging that would be awesome.

astroquery/linelists/jplspec/core.py
keflavich: seems likely.  will check

keflavich: There's no logic overlap here.  _resolve_molecules is getting molecule names either from local parsing or .... assuming the user knows what they're doing.   `_build_table_from_get_molecule` is the new fallback that runs a loop of molecule name retrieval.  I added to the docstring to clarify this

astroquery/linelists/jplspec/core.py
```diff
@@ -170,6 +185,45 @@ def query_lines(self, min_frequency, max_frequency, *,
...
+            return mols
+        if isinstance(molecule, (list, tuple)):
+            return list(molecule)
+        return [molecule]
+
+    def _build_table_from_get_molecule(self, mols):
+        """
+        Fetch full catalog tables for each molecule and combine them.
+
+        `mols` should be passed through `_resolve_molecules` before being
```
bsipocz: should not matter as we don't expose this docs, yet, let's just fix it.
```suggestion
        ``mols`` should be passed through ``_resolve_molecules`` before being
```

bsipocz (approved):
OK, let's merge this. I have one minor comment and pushed a doctest fix, but it should all be good to go now.

PR #3549, opened 2026-03-05
Added a method to HeasarcClass that returns the number of rows in a named catalog

DavidT3:
A small convenience method has been included in the HeasarcClass class which takes the name of a catalog as an argument and passes it to a simple ADQL query that uses COUNT(*) to fetch the number of rows it contains.

The integer number of rows is then extracted from that return, and passed back to the user.

This abstracts another use of ADQL away from the user.

A simple test was added, and an example in the Heasarc sub-module docs.

### DIFF
--- a/astroquery/heasarc/core.py
+++ b/astroquery/heasarc/core.py
@@ -976,7 +976,7 @@ class HeasarcClass(BaseVOQuery, BaseQuery):
             The total number of rows in the specified catalog.
         """
         # Send a query requesting a count of the number of rows in the
-        #  specified catalog.
+        #  specified catalog. Return is a one-row TAPResults object.
         count_return = self.query_tap(f"SELECT COUNT(*) FROM {catalog}")
 
         # Extract an integer number of rows from the returned table.
--- a/astroquery/heasarc/core.py
+++ b/astroquery/heasarc/core.py
@@ -961,7 +961,7 @@ class HeasarcClass(BaseVOQuery, BaseQuery):
     def count_rows(self, catalog):
         """
         Returns the number of rows in the HEASARC-hosted catalog specified by
-        the `catalog` argument.
+        the ``catalog`` argument.
 
         Parameters
         ----------
--- a/astroquery/heasarc/core.py
+++ b/astroquery/heasarc/core.py
@@ -958,5 +958,27 @@ class HeasarcClass(BaseVOQuery, BaseQuery):
             path = key.replace(f's3://{self.S3_BUCKET}/', '')
             _s3_tree_download(self.s3_client, self.S3_BUCKET, path, location)
 
+    def count_rows(self, catalog: str) -> int:
+        """
+        Returns the number of rows in the HEASARC-hosted catalog specified by
+        the `catalog` argument.
+
+        Parameters
+        ----------
+        catalog : `str`
+            The name of the catalog to query for a total number of rows. To list
+            the available catalogs, use
+            :meth:`~astroquery.heasarc.HeasarcClass.list_catalogs`.
+
+        Returns
+        -------
+        num_rows : `int`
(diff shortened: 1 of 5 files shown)


astroquery/heasarc/core.py
```diff
@@ -958,5 +958,30 @@ def _s3_tree_download(client, bucket_name, path, local):
             path = key.replace(f's3://{self.S3_BUCKET}/', '')
             _s3_tree_download(self.s3_client, self.S3_BUCKET, path, location)
 
+    def count_rows(self, catalog: str) -> np.int64:
```
bsipocz: please refrain from typing here, it's not being enforced and thus useless and is the cause of the docs build failure.

astroquery/heasarc/core.py
```diff
@@ -958,5 +958,30 @@ def _s3_tree_download(client, bucket_name, path, local):
...
+        Parameters
+        ----------
+        catalog : `str`
+            The name of the catalog to query for a total number of rows. To list
+            the available catalogs, use
+            :meth:`~astroquery.heasarc.HeasarcClass.list_catalogs`.
+
+        Returns
+        -------
+        num_rows : `np.int64`
```
bsipocz: same here, this may not even be the  correct intersphinx link (I suppose one needs to look into the intersphinx file, based on their deployed docs this may not be the one).

astroquery/heasarc/core.py
```diff
@@ -958,5 +958,30 @@ def _s3_tree_download(client, bucket_name, path, local):
             path = key.replace(f's3://{self.S3_BUCKET}/', '')
             _s3_tree_download(self.s3_client, self.S3_BUCKET, path, location)
 
+    def count_rows(self, catalog: str) -> np.int64:
+        """
+        Returns the number of rows in the HEASARC-hosted catalog specified by
+        the `catalog` argument.
```
bsipocz: Use double backtick, parameter names are not link resolvables and thus causing sphinx errors
```suggestion
        the ``catalog`` argument.
```

bsipocz (changes requested):
The code and test and changelog looks OK, but the failing docs build is a blocker for merging. I left some suggestions for fixing it up, if you insist on keeping the typehinting, that is OK, but the build still needs to pass.

astroquery/heasarc/core.py
DavidT3: Oops, sorry I don't even notice I'm doing it anymore - I'll remember in future.

DavidT3:
Thanks for looking at this Brigitta! Everything you pointed out should be fixed now, I'm definitely not insisting on type hinting, its just how I do it for my own projects and I didn't think to check here.

Also added a non-remote mock test, so coverage will be improved I hope? I'm assuming that codecov isn't set up to run the remote tests, as it was grumpy with me about missing lines that I had only covered with a remote test.

bsipocz:
> Also added a non-remote mock test, so coverage will be improved I hope? I'm assuming that codecov isn't set up to run the remote tests, as it was grumpy with me about missing lines that I had only covered with a remote test.

Yes, codecov is not running on remote as we don't run the remote tests for PRs. That also means that I only look at those numbers as a rough indication and merging happyly with grumpy codecov but reasonable test coverage in the PR via remote tests.

bsipocz (approved):
Changes look good, thanks!

Two more minor suggestions, neither related to the PR itself but are workflow relevant: 

 - working off of feature branches on your fork is recommended. you may end up in a situation when you work on multiple PRs at the same time; it would help you with managing it
 - grouping commits together and squashing out minor iteration would make the history leaner and less noisy. We don't squash merge in this repo, but instead trying to keep individual PR history as clean as possible (we don't always do well, but trying).

PR #3552, opened 2026-03-10
CI: bumping python version for OSX job

bsipocz:
Installing the dependencies takes forever, hopefully this change will make the job much faster than the current 13mins.

Currently a draft as it doesn't fixes the slowness of the job.


bsipocz:
Well, this doesn't do the trick, the job is still very slow :(

PR #3558, opened 2026-03-19
EUCLID: Fix test failures in Euclid documentation [ci skip]

cosmoJFH:
Hi Astroquery team,

in this pull request we would like to fix the issue  (#3553) related to the latest changes in the Euclid documentation.

We have included the label

   .. doctest-skip::

since most of the tests require logging in to the Euclid archive.

cc @esdc-esac-esa-int 
jira: EUCLIDPCR-2104

### DIFF
--- a/docs/esa/euclid/euclid.rst
+++ b/docs/esa/euclid/euclid.rst
@@ -143,7 +143,7 @@ the "mer_catalogue" and its outcome is restricted to 50 rows.
     >>> import astropy.units as u
     >>> coord  = SkyCoord("17h51m07.4s +65d31m50.8s", frame='icrs')
     >>> radius = u.Quantity(0.5, u.deg)
-    >>> job    = Euclid.cone_search(coordinate=coord, radius=radius, columns="*", async_job=True)
+    >>> job    = Euclid.cone_search(coordinate=coord, radius=radius, columns="*", async_job=True)  # doctest: +IGNORE_WARNINGS
     INFO: Query finished. [astroquery.utils.tap.core]
     >>> cone_results    = job.get_results()
     >>> print(f"Found {len(cone_results)} results")
--- a/docs/esa/euclid/euclid.rst
+++ b/docs/esa/euclid/euclid.rst
@@ -102,32 +102,33 @@ This is the access mode for non-registered (also known as "anonymous" in this do
 
 It is possible to access to the metadata (e.g., table names, sizes, descriptions, column names, etc.) of all the TAP_ tables stored in the Archive using the load_tables_ method. This feature allows to have a broad overview of the Archive content. To load only table names:
 
-.. doctest-skip::
-
     >>> tables = Euclid.load_tables(only_names=True, include_shared_tables=True)
+    INFO: Retrieving tables... [astroquery.utils.tap.core]
+    INFO: Parsing tables... [astroquery.utils.tap.core]
+    INFO: Done. [astroquery.utils.tap.core]
     >>> print(f'* Found {len(tables)} tables')
+    * Found 98 tables
     >>> print(*(table.name for table in tables), sep="\n")  # doctest: +IGNORE_OUTPUT
-    ivoa.obscore
-    public.dual
-    sedm.raw_detector
-    sedm.raw_frame
-    sedm.raw_quadrant
+    catalogue.mer_catalogue
+    catalogue.mer_cutouts
+    catalogue.mer_morphology
+    catalogue.phz_classification
     ...
 
 
 To load only one table and inspect its columns:
 
-  >>> raw_detector_table = Euclid.load_table('sedm.raw_detector')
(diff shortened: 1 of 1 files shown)


docs/esa/euclid/euclid.rst
```diff
@@ -104,15 +102,17 @@ This is the access mode for non-registered (also known as "anonymous" in this do
...
-  >>> tables = Euclid.load_tables(only_names=True, include_shared_tables=True)
-  >>> print(f'* Found {len(tables)} tables')
-  >>> print(*(table.name for table in tables), sep="\n")  # doctest: +IGNORE_OUTPUT
-  ivoa.obscore
-  public.dual
-  sedm.raw_detector
-  sedm.raw_frame
-  sedm.raw_quadrant
-  ...
+.. doctest-skip::
```
bsipocz: Maybe leave a comment why the testing is skipped, here and elsewhere below?
```suggestion
.. Skip testing as the example require authentication
.. doctest-skip::
```

docs/esa/euclid/euclid.rst
```diff
@@ -137,15 +137,18 @@ The cone_search_ method allows to easily retrieve data around a projected circul
...
-  basic_download_data_oid to_be_published      object_id       right_ascension   ...       gaia_id        gaia_match_quality           dist
-  ----------------------- --------------- ------------------- ------------------ ... ------------------- -------------------- ----------------------
+
+.. doctest-skip::
+
+    >>> from astropy.coordinates import SkyCoord
+    >>> import astropy.units as u
+    >>> coord  = SkyCoord("17h51m07.4s +65d31m50.8s", frame='icrs')
+    >>> radius = u.Quantity(0.5, u.deg)
+    >>> job    = Euclid.cone_search(coordinate=coord, radius=radius, columns="*", async_job=True)
```
bsipocz: which part here requires authentication? Maybe just a `# doctest: +IGNORE_OUTPUT` should be enough at the end of the print line? So really just hard skip the examples that indeed require the authentication?

docs/esa/euclid/euclid.rst
```diff
@@ -353,51 +374,60 @@ In the Archive the 1D Spectra data  is served via the the Datalink_ (a data acce
...
->>> fontsize = 18
->>> plt.figure(figsize = [20,5])
->>> plt.plot(spec['WAVELENGTH'], spec['SIGNAL'], linewidth = 3, label = f"Source: {inp_source}")
->>> plt.xlabel(f"WAVELENGTH [{spec['WAVELENGTH'].unit}]", fontsize = fontsize)
->>> plt.ylabel('SIGNAL', fontsize = fontsize)
->>> plt.xticks(fontsize = fontsize)
->>> plt.yticks(fontsize = fontsize)
->>> plt.legend(fontsize = fontsize)
->>> plt.show()
+.. doctest-skip::
```
bsipocz, line 465: I'm not sure that we need to include the plotting code if we just insert the png.

bsipocz (commented):
I think we should distinguish test that are failing as we need authentication vs the ones that are failing as we get slightly different outputs. E.g. I would not skip the latter just ignore the output comparison.

And for the authentication requiring ones I would add a comment before the skip, so we have a trace why the testing is skipped.

docs/esa/euclid/euclid.rst
cosmoJFH: We would like to leave that part, just to show how the figure was made.

docs/esa/euclid/euclid.rst
cosmoJFH: Hi @bsipocz , we have reviewed the documentation again, and we have checked that only a few tests require being logged in. We have fixed many errors in the documentation, but we think that now it is correct. (we did not get any error running the command pytest docs/esa/euclid/ -R)

docs/esa/euclid/euclid.rst
cosmoJFH: None of them. Therefore they will be executed by pytest.

bsipocz:
Unfortunately I still get test failures, but I'll look into a patch that gets around it 

(I think we should in fact hide these unit warnings from the user in the code itself as they cannot do much about it if the archive releases data with angstrom unit -- which frankly we will all keep doing on the archive levels).

bsipocz (approved):
Thank you! 

An unrelated suggestion that I picked up on while looking at the docs: I think the notes would look better in a box rather than boldfaced, but it's your call if you want that or not, so I'll open my suggestions in a separate PR and you can decide.

hectorcanovas:
Hi Brigitta,

Thanks a lot for your help and feedback; we will take into account you suggestions (e.g., using boxes instead of boldfaced) the next time that this documentation is updated (which probably will happen in less than a year given all the developments of this Archive).

Regards,
Héctor

PR #3559, opened 2026-03-20
EUCLIDSWRQ-285 Update get_cutout API and related tests

Enriquerzrz:
- Removed "instrument" and "id" arguments since they are deemed "useless"
- Updated the help content and comments to clarify the method goals and explain that it only retrieves MER mosaic (background substracted) image cutouts
- Update the "coordinate" argument information with the following: "coordinate: astropy.coordinate or Simbad/VizieR/NED name (str), mandatory. coordinates center point"
- Updated related tests

Jira:[EUCLIDSWRQ-285](https://s2e2.cosmos.esa.int/jira/browse/EUCLIDSWRQ-285)
CC: @esdc-esac-esa-int

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,7 +30,7 @@ esa.euclid
   sourcepatch_id. [#3543]
 -  The ``source_id`` kwarg in the ``get_spectrum`` method has been renamed to ``ids``. [#3543]
 - Method ``get_cutout`` was updated to remove the deprecated instrument and id parameters, and clarified that it retrieves only
-  MER (background‑subtracted) image cutouts. [#]
+  MER (background‑subtracted) image cutouts. [#3559]
 
 vizier
 ^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,8 @@ esa.euclid
 - The method ``get_spectrum`` accepts the new parameter ``linking_parameter`` to retrieve the spectra by source_id and
   sourcepatch_id. [#3543]
 -  The ``source_id`` kwarg in the ``get_spectrum`` method has been renamed to ``ids``. [#3543]
+- Method ``get_cutout`` was updated to remove the deprecated instrument and id parameters, and clarified that it retrieves only
+  MER (background‑subtracted) image cutouts. [#]
 
 vizier
 ^^^^^^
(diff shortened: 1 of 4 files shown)


astroquery/esa/euclid/core.py
```diff
@@ -1398,20 +1398,20 @@ def get_product(self, *, file_name=None, product_id=None, schema='sedm', output_
 
         return files
 
-    def get_cutout(self, *, file_path=None, instrument=None, id=None, coordinate, radius, output_file=None,
+    @deprecated_renamed_argument('instrument', 'id', since='0.4.12', arg_in_kwargs=True)
```
bsipocz: This is incorrect, now you renamed `instrument` to be `id` instead.
(the signature is `deprecated_renamed_argument(old_name, new_name, since, ...)`. Also, these were not `agr_in_kwargs`

To deprecate multiple parameters, make them a tupple and add `None` as the new name.
```suggestion
    @deprecated_renamed_argument(('instrument', 'id'), (None, None), since='0.4.12')
```

CHANGES.rst
```diff
@@ -29,6 +29,8 @@ esa.euclid
 - The method ``get_spectrum`` accepts the new parameter ``linking_parameter`` to retrieve the spectra by source_id and
   sourcepatch_id. [#3543]
 -  The ``source_id`` kwarg in the ``get_spectrum`` method has been renamed to ``ids``. [#3543]
+- Method ``get_cutout`` was updated to remove the deprecated instrument and id parameters, and clarified that it retrieves only
+  MER (background‑subtracted) image cutouts. [#3559]
```
bsipocz: Some rephrasing is needed here. 
```suggestion
- Method ``get_cutout`` has deprecated the 'instrument' and 'id' parameters, providing them has no effect any more. The method now only supports retrieval of MER (background‑subtracted) image cutouts. [#3559]
```

bsipocz (changes requested):
The usage of the decorator needs to be fixed, otherwise the PR is looking good.

(Normally we keep the old functionality around for a little while, but in this case I think this removal is OK).

astroquery/esa/euclid/core.py
Enriquerzrz: Sorry for the confusion. I used that form because the error traceback suggested it, but it was wrong since those arguments were already removed. Thanks for your comments, I understand now how the decorator should be used. Corrected in the last commit.

CHANGES.rst
Enriquerzrz: Corrected in the last commit.

bsipocz (approved):
Thank you this looks good now.

Minor comment that doesn't affect the merging for now: for resolving code conflicts with the main branch instead of merge commits, we prefer to have rebased branches as that leads to a cleaner history. Given this is a relatively straightfwd change, I won't rebase the PR, so this is just a workflow suggestion for the future.

Again, thanks for the PR.

PR #3561, opened 2026-03-20
ASB-30568: Adding read_product Function to load files from s3 to memory

AlexReedy:
Adding in ability to read FITS and ASDF data products to memory from s3:// using Observations.read_product() function


astroquery/mast/observations.py
```diff
@@ -1203,6 +1214,58 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+
+                try:
+                    fs = s3fs.S3FileSystem(anon=True)
+                    with fs.open(product_path, 'rb') as s3_file:
+                        af = asdf.open(s3_file, ignore_unrecognized_tag=ignore_unrecognized)
+                        return af
+                except Exception as e:
+                    log.exception(f"Failed to open ASD File: {product_path} {e}")
+        else:
+            print("Unsupported extension type")
```
bsipocz, line 1267: please don't print anything; either raise proper warning or error classes or put this behind a verbose option.

astroquery/mast/observations.py
```diff
@@ -44,6 +46,15 @@
     '`~astroquery.mast.ObservationsClass.enable_cloud_dataset` method.'
 )
 
+asdf_packages = ["asdf", "s3fs", "fsspec", "lz4", "gwcs"]
```
bsipocz, line 49: does all of these really need to be checked here even as they are not directly been used? If `asdf` requires the whole list then these checks should be dealt with upstream in asdf itself.

bsipocz (commented):
The commit history is a bit all over the place. Could you please clean it up and squash to logical chucks rather than quasi random back and forth? Thanks!

I suppose this will also need narrative documentation; but I'll leave the more detailed review to @snbianco

astroquery/mast/observations.py
AlexReedy: This has been an ongoing question with the implementation. Required would be asdf, s3fs, fsspec for the function to work. for asdf specifically, lz4 seems to be the primary compression algorithm being used for asdf and gwcs is for the general.

I know that gwcs is in their test environment, and they also make calls to lz4 but I don't believe it's included when it's installed. I agree it should probably be upstream in asdf.

astroquery/mast/observations.py
AlexReedy: ah thank you! I thought I got rid of all of those

snbianco:
Thanks again for this PR, Alex! Can you add a quick section about this function to the `Observations` docs (`docs/mast/mast_obsquery.rst`). You'll probably want to put it in the cloud data access section.

This PR will also need some tests. The non-remote-access tests may be slightly tricky with mocking. Let me look into it and see if I can point you in the right direction. The remote-access tests should be more straightforward.

astroquery/mast/tests/test_mast.py
```diff
@@ -1355,6 +1355,130 @@ def test_observations_disable_cloud_dataset(patch_boto3):
     assert Observations._cloud_enabled_explicitly is False
 
 
+@pytest.fixture
```
snbianco: The convention we have for fixtures in this file is to put them near the top, before any of the tests. For this one and the s3_asdf_path specifically, though, I'm not sure if we need a fixture, since they're only being used once. I'd rather see the value of `s3_fits_path` be parametrized.

astroquery/mast/tests/test_mast.py
```diff
@@ -1355,6 +1355,130 @@ def test_observations_disable_cloud_dataset(patch_boto3):
...
+    return "s3://stpubdata/hst/public/u9o4/u9o40504m/u9o40504m_c3m.fits"
+
+
+@pytest.fixture
+def mock_fits_open(mocker):
+    return mocker.patch("astropy.io.fits.open", return_value=MagicMock(name="HDUList"))
+
+
+def test_read_product_fits(s3_fits_path, mock_fits_open, mocker):
+    mocker.patch("astropy.__version__", "5.0.0")
```
snbianco: Why do we need this line?

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -1035,6 +1035,25 @@ def test_observations_get_cloud_uris_no_duplicates(self, msa_product_table, rese
...
+
+        product = Observations.read_product(product_path)
+
+        assert isinstance(product, fits.HDUList)
+
+    @pytest.mark.remote_data
+    def test_observations_read_product_asdf(self):
+        asdf = pytest.importorskip("asdf")
+
+        product_path = "s3://stpubdata/roman/nexus/soc_simulations/tutorial_data" \
```
snbianco: Are there any smaller files that we have available to us from the Nexus? This takes around 15 seconds to load in, and we try to keep this test suite as short in duration as we can.

astroquery/mast/observations.py
```diff
@@ -1203,6 +1210,51 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
 
+    # TODO: Need to inlcude way to parse if it is a MAST on prem URL and handle the streaming of that
+    def read_product(self, product_path, read_as="auto", ignore_unrecognized=True):
+        """
+        Read a product from Open S3 bucket to memory. Currently supports FITS and ASDF product types only.
+
+        Parameters
+        ----------
+        product_path: str
+            URI to the product in open bucket.
```
snbianco:
```suggestion
            URI to the product in the STScI S3 open data bucket.
```

astroquery/mast/observations.py
```diff
@@ -1203,6 +1210,51 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+        read_as: str, optional
+            How to read the file. Currently only .fits and .asdf is supported by "auto". Defaults to "auto".
```
snbianco:
```suggestion
            How to read the file. Currently only FITS and ASDF file types are supported by "auto". Default is "auto".
```

astroquery/mast/observations.py
```diff
@@ -1203,6 +1210,51 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+        ignore_unrecognized: bool
+            Tells asdf.open() to include or ignore warnings from unrecognized asdf tags. Defaults to True
+
+        Returns
+        -------
+        object
+            FITS or ASDF object.
```
snbianco:
```suggestion
            FITS or ASDF object for the given data product.
```

astroquery/mast/observations.py
```diff
@@ -1203,6 +1210,51 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+        """
+        path_lower = product_path.lower()
+
+        if read_as == "auto":
+            # Read logic for fits data products
+            if path_lower.endswith((".fits", ".fits.gz")):
+                try:
+                    log.info(f"Loaded: {product_path}")
```
snbianco: I don't think we should log this until after `fits.open` executes successfully.

astroquery/mast/observations.py
```diff
@@ -1203,6 +1210,51 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+                    return fits.open(product_path, fsspec_kwargs={"anon": True})
```
snbianco: I'm realizing that for this line, we will need fsspec[s3], which is not a listed dependency of astroquery. We should probably add this to `pyproject.toml`. See other comment about switching over ASDF to `fsspec` as well.

astroquery/mast/observations.py
```diff
@@ -1203,6 +1210,51 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+                except Exception as e:
```
snbianco: Can we be any more specific with this exception? If not, that's alright, but we should include a comment on why we catch exceptions so broadly. Same with line 1251.

astroquery/mast/observations.py
```diff
@@ -1203,6 +1210,51 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+                    log.exception(f"Failed to open FITS File: {product_path} {e}")
+
+            # Read logic for ASDF
+            elif path_lower.endswith(".asdf"):
```
snbianco: These should be some logic in this branch for checking that the user has the proper packages installed, and it should error if they do not.

astroquery/mast/observations.py
```diff
@@ -1203,6 +1210,51 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+                try:
+                    fs = s3fs.S3FileSystem(anon=True)
+                    with fs.open(product_path, 'rb') as s3_file:
```
snbianco: I actually was experimenting a bit on Friday with a version of this function for the `MastMissions` class, and I think we should actually use `fsspec` here over `s3fs`. `fsspec` is a generic wrapper for `s3fs`, but it also supports other types of file systems (like HTTP, which we will want later).
```suggestion
                    f = fsspec.open(product_path, "rb", anon=True).open()
                    af = asdf.open(fs, ignore_uncrecognized_tag=ignore_unrecognized_tag)
```

astroquery/mast/observations.py
```diff
@@ -1203,6 +1210,51 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+        path_lower = product_path.lower()
+
+        if read_as == "auto":
+            # Read logic for fits data products
+            if path_lower.endswith((".fits", ".fits.gz")):
+                try:
+                    log.info(f"Loaded: {product_path}")
+                    return fits.open(product_path, fsspec_kwargs={"anon": True})
+                except Exception as e:
+                    log.exception(f"Failed to open FITS File: {product_path} {e}")
```
snbianco: Instead of logging an exception here, we should actually raise one. Same for line 1252 and 1255.

docs/mast/mast_obsquery.rst
```diff
@@ -629,3 +629,14 @@ remain fully cloud-based.
    COMPLETE
    COMPLETE
    COMPLETE
+   
+Streaming Data Products from S3 to memory
+-----------------------------------------
+If instead of downloading you would like to load an S3 URI directly to memory you can use `~astroquery.mast.ObservationsClass.read_product`. 
+This function supports FITS and ASDF data products and will automatically parse the file for the suffix and load it to memory using `~astropy.io.fits.open` or ``~asdf.open``.
```
snbianco:
```suggestion
This function supports FITS and ASDF data products and will automatically parse the file for the suffix and load it to memory using `~astropy.io.fits.open` or `~asdf.open`.
```

docs/mast/mast_obsquery.rst
```diff
@@ -629,3 +629,14 @@ remain fully cloud-based.
    COMPLETE
    COMPLETE
    COMPLETE
+   
+Streaming Data Products from S3 to memory
+-----------------------------------------
+If instead of downloading you would like to load an S3 URI directly to memory you can use `~astroquery.mast.ObservationsClass.read_product`.
```
snbianco:
```suggestion
If instead of downloading you would like to load an S3 URI directly to memory, you can use the `~astroquery.mast.ObservationsClass.read_product` method. 
```

docs/mast/mast_obsquery.rst
```diff
@@ -629,3 +629,14 @@ remain fully cloud-based.
    COMPLETE
    COMPLETE
    COMPLETE
+   
+Streaming Data Products from S3 to memory
+-----------------------------------------
+If instead of downloading you would like to load an S3 URI directly to memory you can use `~astroquery.mast.ObservationsClass.read_product`. 
+This function supports FITS and ASDF data products and will automatically parse the file for the suffix and load it to memory using `~astropy.io.fits.open` or ``~asdf.open``.
+For ASDF data products additional packages may be required (e.g lz4 and roman-datamodels for ROMAN data).
```
snbianco:
```suggestion
For ASDF data products, additional packages may be required (e.g `~lz4` and `~roman-datamodels` for data from the Roman Space Telescope). 
```

CHANGES.rst
```diff
@@ -147,6 +147,8 @@ mast
 - The cloud dataset in ``Observations`` is now enabled by default if the ``boto3`` and ``botocore`` packages are installed. This
   default can be overridden by setting the ``enable_cloud_dataset`` configuration option to False. [#3534]
 
+- Adding in ability to read FITS and ASDF dataproducts to memory from s3:// using ``Observations.read_product()`` function. [#3561]
```
snbianco:
```suggestion
- Adding in ability to read FITS and ASDF data products to memory from STScI's S3 open data bucket using ``Observations.read_product()`` function. [#3561]
```

pyproject.toml
snbianco:
```suggestion
   "fsspec[http,s3]",
```

snbianco (commented):
Another round of review! This is looking good. A lot of my comments were questions or notes on the docs/docstrings. Let me know if you have questions or want to chat more about the switch to `fsspec`!

astroquery/mast/observations.py
snbianco: Also! I forgot in my initial review, but ASDF file objects are a little finicky with when and how you close the files. Data arrays that are unloaded when you close the file handler stay unloaded and will throw an error if you try to access them later. For that reason, we can't use context managers here and need to leave both file handlers open. I would also add a line in the remote tests to make sure that you can access unloaded arrays in ASDF objects.

astroquery/mast/tests/test_mast_remote.py
AlexReedy: I can check again but looking at the stuff that's on s3 the tutorial data is the smallest I saw at 4.0KB

PR #3550, opened 2026-03-06
EUCLID: update output filename for downloads in the methods get_product  and get_cutout

cosmoJFH:
Dear Astroquery team,

The Astroquery Euclid get_product method (https://astroquery.readthedocs.io/en/latest/api/astroquery.esa.euclid.EuclidClass.html#astroquery.esa.euclid.EuclidClass.get_product) allows one to retrieve one or multiple Euclid products (FITS files) provided that the user knows the file_name or product_id of the product to be retrieved. The method is certainly useful, but its implementation could be improved to make the life of the users easier:

out_path = Euclid.get_product(file_name = 'EUC_VIS_SWL-DET-002704-01-1-0000000__20241017T042805.678587Z.fits')

By default, the code writes the downloaded file into a file named "temp_YYYYMMDD_HHMMDD/temp", where "temp_YYYYMMDD_HHMMDD" is automatically created by the method. The name of the file is at the very least misleading (especially since it does not even include the .fits extension). We suggest changing "temp" to the actual name of the file downloaded. That is, in the example above, the downloaded file should be stored as: "temp_YYYYMMDD_HHMMDD/EUC_VIS_SWL-DET-002704-01-1-0000000__20241017T042805.678587Z.fits"

The same scenario is present in the method get_cutout.

cc @esdc-esac-esa-int 
Jira: EUCLIDSWRQ-283

PR #3541, opened 2026-02-27
Updated get_product to accept file_name as a python list

Enriquerzrz:
Method get_product now supports providing file_name as a Python list (e.g. ["file1.fits", "file2.fits"]) while still accepting the original comma‑separated string format. Also added new one test to check this new functionality.

CC: @esdc-esac-esa-int 
Jira: EUCLIDSWRQ-284

PR #3564, opened 2026-03-23
EUCLID: update output filename for downloads in the method get_product for multiple values in  `file_name` or `product_id`

cosmoJFH:
Dear Astroquery team,

in the PR #3550 the method `get_product` was updated so that the output file name is directly related to the input `file_name` or `product_id`. We have realized that that implementation does not work as we would expect in the case multiple values for `file_name`or `product_id` are used.

In this PR we have implemented the following changes:

1. The parameter `product_id` now accepts a list/tuple (similar to the changes in PR #3541)

2. For multiple values, in case the parameter `output_file` is not defined, the downloaded zip file is automatically created with the name get_product_output.zip and it is not uncompressed. The present implementation downloads the zip file (with wrong name and extension) and uncompresses it. So the user always have both files, the compressed and the uncompressed. Since the downloaded fits files can be pretty large (tens of GB), we think that this change is pretty convenient.
3. A similar behaviour is obtained in case  multiple values are used and the parameter `output_file `is defined by the user.

cc @esdc-esac-esa-int 
jira: EUCLIDSWRQ-283

### DIFF
--- a/astroquery/esa/euclid/core.py
+++ b/astroquery/esa/euclid/core.py
@@ -792,23 +792,39 @@ class EuclidClass(TapPlus):
             now = datetime.now()
             output_dir = os.path.join(os.getcwd(), "temp_" + now.strftime("%Y%m%d_%H%M%S"))
             output_file_full_path = os.path.join(output_dir, str(observation_id))
+
+            try:
+                if output_dir:
+                    os.makedirs(output_dir, exist_ok=True)
+            except OSError as err:
+                raise OSError("Creation of the directory %s failed: %s"
+                              % (output_dir, err.strerror))
         else:
-            output_file_full_path = output_file
-            output_dir = os.path.dirname(output_file_full_path)
-        try:
-            if output_dir:
-                os.makedirs(output_dir, exist_ok=True)
-        except OSError as err:
-            raise OSError("Creation of the directory %s failed: %s"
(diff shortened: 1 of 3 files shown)


astroquery/esa/euclid/core.py
```diff
@@ -1318,25 +1336,28 @@ def get_product(self, *, file_name=None, product_id=None, schema='sedm', output_
 
         Parameters
         ----------
-        file_name : str, optional, default None
+        file_name : str or list of str, mandatory, default None
```
bsipocz:
```suggestion
        file_name : str or list of str, default None
```

bsipocz (commented):
overall looks good, but it needs to be rebased to pick up recent conflicting changes.

astroquery/esa/euclid/core.py
cosmoJFH: Hi @bsipocz, we have rebased, resolved the conflicts and pushed (-force) the changes. Please double-check our changes, just in case we did something wrong.

bsipocz: changes look good. However, I would still apply this suggestion above as later the docstring (and code) says that providing either file_name or product_id should be sufficient, thus it's not mandatory.

bsipocz (approved):
Thanks!

PR #3567, opened 2026-03-24
MAST: Deprecate objectname

snbianco:
The `objectname` parameter in `Catalogs`, `Observations`, `Tesscut`, and `utils` is deprecated in favor of `object_name`.

I also used this PR to fix an unintentional backwards compatibility issue that was introduced in #3540. Previously, users could input a single coordinate string with a comma separating the RA and declination (`"20, -30"`). The code introduced in that PR splits the string on commas, assuming that coordinates are only separated by a space, which breaks this workflow. I implemented a corner case that detects this historical single-coordinate input and works around it to still return results.

### DIFF
--- a/astroquery/mast/missions.py
+++ b/astroquery/mast/missions.py
@@ -290,7 +290,8 @@ class MastMissionsClass(MastQueryWithLogin):
         list of str
             A list of target strings in "ra dec" format for the API.
         """
-        def _normalize_inputs(values):
+        def _as_list(values):
+            """Normalize the input values into a list of strings or coordinate objects."""
             if values is None:
                 items = []
             elif isinstance(values, str):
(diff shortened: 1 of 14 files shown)


astroquery/mast/missions.py
```diff
@@ -338,7 +357,7 @@ def _normalize_inputs(values):
         return targets
 
     @class_or_instance
-    @deprecated_renamed_argument('objectname', 'object_names', since='0.4.11')
+    @deprecated_renamed_argument('objectname', 'object_names', since='0.4.12')
```
bsipocz, line 360: Thanks for spotting this and fixing it!

bsipocz (approved):
This all looks good, thanks!

Unrelated issue that came up while I was running the remote tests for this PR:
- the name resolution sometimes running into timeouts, so I would consider defaulting into a coordinate input for the remote tests rather than using `object_name`. I mean we should test object_name, but maybe just once and all the rest should use a coordinate rather than doing the additional name resolution with the remote service?

Either case, that is something for a follow-up.

PR #3506, opened 2026-01-21
EUCLID: new method get_sia  to access the Simple Image Access Protocol (SIAP)  v2.0

cosmoJFH:
Dear astroquery team,

we would like to implement a new method for the  Simple Image Access Protocol (SIAP)  v2.0 exposed by the Euclid project.

The Euclid Archive applies the following IVOA standard to serve:

1. Tables (catalogues), via the IVOA Table Access Protocol [TAP](https://www.ivoa.net/documents/TAP/)
2. Spectra, via the IVOA [DataLink](https://www.ivoa.net/documents/DataLink/) prococol.
3. Images (and cutouts), via the Simple Image Access ([SIA](https://www.ivoa.net/documents/SIA/20151223/REC-SIA-2.0-20151223.html))

The new get_sia method will allow users to obtain the “file_name” and “file_path” of the FITS files that contain the observations of the input coordinates. 

cc @esdc-esac-esa-int 
jira: EUCLIDSWRQ-250

PR #3569, opened 2026-03-25
EUCLID: new method get_sia to access the Simple Image Access Protocol (SIAP) v2.0

cosmoJFH:
Dear astroquery team,

we would like to implement a new method for the  Simple Image Access Protocol (SIAP)  v2.0 exposed by the Euclid project.

The Euclid Archive applies the following IVOA standard to serve:

1. Tables (catalogues), via the IVOA Table Access Protocol [TAP](https://www.ivoa.net/documents/TAP/)
2. Spectra, via the IVOA [DataLink](https://www.ivoa.net/documents/DataLink/) prococol.
3. Images (and cutouts), via the Simple Image Access ([SIA](https://www.ivoa.net/documents/SIA/20151223/REC-SIA-2.0-20151223.html))

The new get_sia method will allow users to obtain the “file_name” and “file_path” of the FITS files that contain the observations of the input coordinates. 

cc @esdc-esac-esa-int 
jira: EUCLIDSWRQ-250 

Note that this PR clones #3506, that was closed due to an error when updating the remote branch.


cosmoJFH:
Hi @bsipocz, we’re planning to include this feature in the next release scheduled for the end of May. Would the team have time to review this pull request?

cosmoJFH:
Hi @bsipocz,

We'd like to merge this new method into the develop branch because we need it for the DataLabs service. Would you have time to review the changes when you get a chance?

Thanks very much in advance!

cosmoJFH:
Hi @bsipocz,

we’re planning to include this new feature in the next release scheduled for the end of July. Would the team have time to review this pull request?

astroquery/esa/euclid/core.py
```diff
@@ -1482,6 +1514,110 @@ def __is_multiple(self, value):
...
 
+    def get_sia(self, *, search_type='CIRCLE', ra, dec, radius, calibration=2, instrument='ALL', band=None,
+                collection='sedm', dsr_part1=None, dsr_part2=None, dsr_part3=None, output_file=None, verbose=False):
+        """
+        Access the Euclid Observation Images by VO SIAP v2.0. This service will return public images from Calibrated
+        and Stacked NISP and VIS images, MER Mosaics from VIS and NISP and Level 1 (RAW) images for NISP and VIS
+
+        Parameters
+        ----------
+        search_type : str, mandatory, default None
```
keflavich: default  is `'CIRCLE'` right now

astroquery/esa/euclid/core.py
```diff
@@ -1482,6 +1514,110 @@ def __is_multiple(self, value):
 
         return not isinstance(value, int) and ((isinstance(value, (list, tuple)) and len(value) > 1) or ',' in value)
 
+    def get_sia(self, *, search_type='CIRCLE', ra, dec, radius, calibration=2, instrument='ALL', band=None,
```
keflavich: there are non-keyword arguments after a keyword argument. I thought this was syntactically disallowed?

We also prefer using `coordinate` (a skycoord) rather than manually specified ra/dec

keflavich: the order should be required positional arguments (e.g., `coordinate, radius, search_type`), then `*`, then keyword arguments.

astroquery/esa/euclid/core.py
```diff
@@ -1482,6 +1514,110 @@ def __is_multiple(self, value):
...
+
+        if instrument == 'VIS' and band is not None and band not in valid_band_vis:
+            raise ValueError(f"Invalid band {band} for instrument {instrument}")
+
+        if instrument == 'NISP' and band is not None and band not in valid_band_nisp:
+            raise ValueError(f"Invalid band {band} for instrument {instrument}")
+
+        ra_deg = self.coordinates_degrees(ra)
+        dec_deg = self.coordinates_degrees(dec)
+        radius_deg = self.coordinates_degrees(radius)
```
keflavich: this should be handled by standard astroquery coordinate handling mechanisms

astroquery/utils/tap/core.py
```diff
@@ -863,6 +863,8 @@ def load_data(self, *, params_dict=None, output_file=None, verbose=False):
         output_file : string, optional, default None
             file where the results are saved.
             If it is not provided, the http response contents are returned.
+        http_method: string, compulsory, default POST
```
keflavich: avoid mixing 'compulsory' and 'mandatory'.  `required` is the standard term anyway.

I don't think you need to say `required` when there is a default value, though - just say the options

astroquery/utils/tap/core.py
```diff
@@ -879,12 +881,20 @@ def load_data(self, *, params_dict=None, output_file=None, verbose=False):
         data = urlencode(params_dict)
         if verbose:
             print(f"Data request: {data}")
-        response = connHandler.execute_datapost(data=data, verbose=verbose)
+
+        if http_method == 'POST':
+            response = connHandler.execute_datapost(data=data, verbose=verbose)
+        else:
```
keflavich: maybe `elif http_method == 'get'`?

keflavich (dismissed):

astroquery/esa/euclid/core.py
cosmoJFH: We fully agree with your comment. the description of the parameter `search_type`  is now

```
        search_type : str, optional, default CIRCLE
            Shape of the search region. Supported values are "CIRCLE" and "BOX".
```

astroquery/esa/euclid/core.py
cosmoJFH: The parameter `coordinates` accepts a `str` or `SkyCoord`, since it is parsed by the method `commons.parse_coordinates`.

We would like to define the parameters as keyword-only arguments, because `get_sia` has many optional arguments, and requiring keywords makes calls much more self-documenting. We do not set a default value for `coordinates`, so callers must provide the coordinates argument.

astroquery/esa/euclid/core.py
cosmoJFH: We fully agree with you. We make use of the methods  `commons.parse_coordinates(coordinates=coordinates)` and 
 `esautils.get_degree_radius(radius)`

astroquery/utils/tap/core.py
cosmoJFH: Updated.

astroquery/utils/tap/core.py
cosmoJFH: Updated.

cosmoJFH:
Converted to a draft till is fully validated by our team.

cosmoJFH:
The ESAC team has reviewed and tested the changes. We think that the PR is ready to be reviewed again by the Astroquery team. Thank you very much in advance.

cosmoJFH:
Hi @bsipocz,

we’re planning to include this new feature in the next release of the DATALABS scheduled for the end of September. Would the team have time to review this pull request? Thank you very much in advance.

astroquery/esa/euclid/core.py
```diff
@@ -1482,6 +1515,119 @@ def __is_multiple(self, value):
 
         return not isinstance(value, int) and ((isinstance(value, (list, tuple)) and len(value) > 1) or ',' in value)
 
+    def get_sia(self, *, coordinates, radius=0.1, search_type='CIRCLE', calibration=2, instrument='ALL', band=None,
```
bsipocz: I would say we need to rename this method to be `query_sia`, so it will be more consistent with the the API in the other modules.

bsipocz: (The naming of the kwargs won't be consistent, but OTOH, I feel this one makes more sense, so I would propose to migrate the other modules to be something like this instead)

bsipocz (commented):
Overall looks good, however we have `query_sia` as method name in the other modules, so I think we should stick to it here, too.

cosmoJFH:
> Overall looks good, however we have `query_sia` as method name in the other modules, so I think we should stick to it here, too.

The method name was updated: get_sia --> query_sia

astroquery/esa/euclid/core.py
cosmoJFH: I think that it was a mistake in the signature of the method. It has been updated to:

    def query_sia(self, coordinates, *,  radius=0.1, search_type='CIRCLE', calibration=2, instrument='ALL', band=None,
                  collection='sedm', dsr_part1=None, dsr_part2=None, dsr_part3=None, output_file=None, verbose=False):
                  
So _coordinates_ is a required argument, and the rest must be supplied as a keyword argument.

bsipocz: Sounds good. I would not have minded to have coordinates as a mandatory kwarg either, but this looks good, too.

(my comment was directed to the other `query_sia` methods kept the SIA vocabulary and use `pos`, etc, which is far from being end user friendly, and I think we should change all of that to what you have here instead, using `coordinates`, etc.)

docs/esa/euclid/euclid.rst
```diff
@@ -356,6 +356,142 @@ and their sky coverage (in its "fov" field) is queried using ADQL_. Please note:
...
+The service returns a VOTable containing all the information required by the cutout function, including the path to the FITS files (file_name), the instrument, and the observation identifier (obs_id).
+
+The VOTable can also be saved for later use.
+
+  >>> from astroquery.esa.euclid import Euclid
+  >>> from astropy.coordinates import SkyCoord
+  >>> import astropy.units as u
+  >>> coords = SkyCoord(267.78, 65.53, frame='icrs', unit="deg") # NGC 6505
+  >>> radius = u.Quantity(0.01, u.deg)
+  >>> table = Euclid.get_sia(coordinates=coords, radius=radius, calibration=3)
```
bsipocz:
```suggestion
  >>> table = Euclid.query_sia(coordinates=coords, radius=radius, calibration=3)
```

astroquery/esa/euclid/tests/test_euclid_remote.py
```diff
@@ -77,3 +77,24 @@ def test_get_tables():
 
     table = euclid.load_table("catalogue.mer_catalogue")
     assert len(table.columns) == 471
+
+
+@pytest.mark.remote_data
+def test_get_sia():
```
bsipocz:
```suggestion
def test_query_sia():
```

astroquery/esa/euclid/tests/test_euclidtap.py
```diff
@@ -1986,6 +1999,104 @@ def test_load_async_job(mock_querier_async):
     assert job.jobid == jobid
 
 
+@pytest.mark.parametrize("verbose", [False, True])
+def test_get_sia(monkeypatch, verbose):
```
bsipocz:
```suggestion
def test_query_sia(monkeypatch, verbose):
```

astroquery/esa/euclid/tests/test_euclidtap.py
```diff
@@ -1986,6 +1999,104 @@ def test_load_async_job(mock_querier_async):
...
+    fn = 'file_name'
+    assert table[fn][0] == 'EUC_MER_BGSUB-MOSAIC-VIS_TILE101007315-D84386_20230826T000856.482420Z_00.00.fits.gz'
+
+    table = euclid.query_sia(coordinates=coords, radius=u.Quantity(0.01, u.deg), dsr_part1='CALBLOCK',
+                             dsr_part2='PV-023', dsr_part3=1, verbose=verbose)
+    assert isinstance(table, Table)
+    assert table[fn][0] == 'EUC_MER_BGSUB-MOSAIC-VIS_TILE101007315-D84386_20230826T000856.482420Z_00.00.fits.gz'
+
+
+def test_get_sia_exceptions(monkeypatch):
```
bsipocz:
```suggestion
def test_query_sia_exceptions(monkeypatch):
```

bsipocz (approved):
These need to be renamed, too. Otherwise looks good.

bsipocz:
@cosmoJFH - you can certainly keep the additional commits, it in facts help with the re-review as I could then just look at the new diffs.
Anyways, one more update to the docs and this should be good to go.

astroquery/esa/euclid/tests/test_euclidtap.py
cosmoJFH: Changes committed to the branch.

astroquery/esa/euclid/tests/test_euclidtap.py
cosmoJFH: Changes committed to the branch.

astroquery/esa/euclid/tests/test_euclid_remote.py
cosmoJFH: Changes committed to the branch.

astroquery/esa/euclid/core.py
bsipocz: > here are non-keyword arguments after a keyword argument. I thought this was syntactically disallowed?

just to wrap up this discussion: it is allowed to have kwarg only arguments, but also, the signature has changes in more recent commits.

docs/esa/euclid/euclid.rst
cosmoJFH: Changes committed to the branch.

bsipocz:
Thanks @cosmoJFH!

PR #3511, opened 2026-01-28
ESA: abstract EsaTap using PyVO, HST and ISLA refactored, EMDS module and EinsteinProbe module

jespinosaar:
Dear Astroquery team,

We come from PR 3501: https://github.com/astropy/astroquery/pull/3501

As part of the migration to PyVO, we have generated a more general class called EsaTap, extending PyVO capabilities with custom code (authentication, parameters in the request, methods to query tables with specific filters...). We will be using this class to extend ESA modules, so the common code will remain under EsaTap and the mission-specific methods will be defined in their specific modules. As a reference, we have already applied this approach to eHST and ISLA. 

In addition to this, we have generated a new module called ESDC Multi-Mission Data Services (EMDS). Then, inside it, we have generated the module for Einstein Probe module, that depends on the EMDS one.

Please let us know if this is ok with you.

Happy to receive your feedback and implement any change you require!

Kind regards,
@jespinosaar and @pdmElisa

cc @esdc-esac-esa-int

PR #3572, opened 2026-03-31
Migrated ESASky module to use EsaTap/PyVO instead of TAPPlus

imbasimba:
Following the merge of https://github.com/astropy/astroquery/pull/3511, here is the migration for the ESASky module

### DIFF
--- a/astroquery/esasky/__init__.py
+++ b/astroquery/esasky/__init__.py
@@ -17,7 +17,8 @@ class Conf(_config.ConfigNamespace):
     Configuration parameters for `astroquery.esasky`.
     """
 
-    ESASKY_DOMAIN_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER, "ESASky TAP Common Server", aliases='urlBase')
+    ESASKY_DOMAIN_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER, "ESASky TAP Common Server",
+                                              aliases=['astroquery.esasky.urlBase'])
     ESASKY_TAP_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER + ESASKY_TAP_COMMON, "ESASky TAP Server")
     ESASKY_DATA_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER + 'data?', "ESASky Data Server")
     ESASKY_TABLES_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER + ESASKY_TAP_COMMON + "/tables",
@@ -27,9 +28,9 @@ class Conf(_config.ConfigNamespace):
     ESASKY_LOGIN_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER + 'login', "ESASky Login Server")
     ESASKY_LOGOUT_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER + 'logout', "ESASky Logout Server")
     ESASKY_CONNECTION_TIMEOUT = _config.ConfigItem(1000, 'Time limit for connecting to a data product server.',
-                                                   aliases='timeout')
+                                                   aliases=['astroquery.esasky.timeout'])
     ESASKY_ROW_LIMIT = _config.ConfigItem(10000, 'Maximum number of rows returned (set to -1 for unlimited).',
-                                          aliases='row_limit')
+                                          aliases=['astroquery.esasky.row_limit'])
 
     @property
     def urlBase(self):
--- a/astroquery/esasky/__init__.py
+++ b/astroquery/esasky/__init__.py
@@ -17,7 +17,7 @@ class Conf(_config.ConfigNamespace):
     Configuration parameters for `astroquery.esasky`.
     """
 
-    ESASKY_DOMAIN_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER, "ESASky TAP Common Server", alias='urlBase')
+    ESASKY_DOMAIN_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER, "ESASky TAP Common Server", aliases='urlBase')
     ESASKY_TAP_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER + ESASKY_TAP_COMMON, "ESASky TAP Server")
     ESASKY_DATA_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER + 'data?', "ESASky Data Server")
     ESASKY_TABLES_SERVER = _config.ConfigItem(ESASKY_COMMON_SERVER + ESASKY_TAP_COMMON + "/tables",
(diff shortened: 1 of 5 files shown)


astroquery/esasky/__init__.py
```diff
@@ -1,22 +1,33 @@
...
-        'https://sky.esa.int/esasky-tap',
-        'ESASky base URL')
-
-    timeout = _config.ConfigItem(
-        1000,
-        'Time limit for connecting to template_module server.')
 
-    row_limit = _config.ConfigItem(
-        10000,
-        'Maximum number of rows returned (set to -1 for unlimited).')
```
bsipocz, line 19: These are public attributes, so I don't think we can remove them without doing a deprecation first.

bsipocz: (you can make them an alias to the new variables along with a warning)

astroquery/esasky/core.py
```diff
@@ -79,21 +81,22 @@ class ESASkyClass(BaseQuery):
     SSO_TYPES = ['ALL', 'ASTEROID', 'COMET', 'SATELLITE', 'PLANET',
                  'DWARF_PLANET', 'SPACECRAFT', 'SPACEJUNK', 'EXOPLANET', 'STAR']
 
-    def __init__(self, tap_handler=None):
-        super().__init__()
+    def __init__(self, *, show_messages=False, auth_session=None, tap_url=None):
+        super().__init__(auth_session=auth_session, tap_url=tap_url)
```
bsipocz, line 89: OK, so this PR will add some backwards incompatible api changes. If we clearly communicate that in the changelog, I think we can be OK with it even if we decide of not doing a deprecation, but we definitely need to communicate that.

cc @keflavich

bsipocz (commented):
Big picture looks OK, but we will need to think a bit more about what deprecations will need to be done; or if we want to do a clear break and call this a major enough refactor where we don't provide non-breaking API changes.

bsipocz (changes requested):
I see a lot of remote test failures, changing the review status to make sure those will get fixed up before we merge this.

imbasimba:
Now, I have improved the backward compatibility. I deprecated the old properties and used aliases, restored get_columns, restored and deprecated get_tap, and deprecated parameter tap_handler. Hopefully, that is all the API breaks. 
Let me know if I got the property deprecation/aliases right. I’ve never used them myself. 

Regarding the failing remote tests, they seem unrelated to the ESASky module and appear to be a cache and hooks issue in astroquery.query. See https://github.com/imbasimba/astroquery/pull/2/changes/62661283fc6b35c16cca6923958b68c76ee03451
Let me know your thoughts on this workaround, and if acceptable, I can merge it into this PR.

imbasimba:
@bsipocz

bsipocz:
@imbasimba - could you separately PR the caching fix?

astroquery/esasky/__init__.py
```diff
@@ -24,8 +26,50 @@ class Conf(_config.ConfigNamespace):
...
+    ESASKY_CONNECTION_TIMEOUT = _config.ConfigItem(1000, 'Time limit for connecting to a data product server.', alias='timeout')
+    ESASKY_ROW_LIMIT = _config.ConfigItem(10000, 'Maximum number of rows returned (set to -1 for unlimited).', alias='row_limit')
+
+    @property
+    def urlBase(self):
+        return self.ESASKY_DOMAIN_SERVER
+
+    @urlBase.setter
+    def urlBase(self, value):
+        warnings.warn(
```
bsipocz: I'm sorry not picking up on this in the first round, but I wonder if we could/should use the deprecated_attribute functionality from astropy.utils instead?

usage is shown e.g. here: https://github.com/astropy/astropy/pull/19593#discussion_r3170229767

bsipocz: although in these cases the alternative is not another property, so you may want to use the `message` rather than the `alternative` argument, and may need to instead follow the example in its documentation

astroquery/esasky/core.py
```diff
@@ -81,8 +83,15 @@ class ESASkyClass(EsaTap):
     SSO_TYPES = ['ALL', 'ASTEROID', 'COMET', 'SATELLITE', 'PLANET',
                  'DWARF_PLANET', 'SPACECRAFT', 'SPACEJUNK', 'EXOPLANET', 'STAR']
 
-    def __init__(self, *, show_messages=False, auth_session=None, tap_url=None):
+    def __init__(self, *, tap_handler=None, show_messages=False, auth_session=None, tap_url=None):
```
bsipocz, line 88: please use the @deprecated_renamed_argument decorator from astropy.utils instead; plenty of examples for it in the other modules here

emellega:
> @imbasimba - could you separately PR the caching fix?

Here's a separate PR for the cache fix: https://github.com/astropy/astroquery/pull/3586

bsipocz (commented):
Test failures seem to be related and coming from pyvo, I don't think it has anything to do with the caching bug. Could you please have a look at it? Also, a few comments, but I stopped looking into into the details once noticed the test issues.

emellega:
You are correct, there were some unrelated errors from our backend. We updated our backend recently (in between me writing the code for the "caching fix" and actually posting it as a PR), which broke some of the tests. I have now deployed a fix to our backend. I have also pushed another commit addressing some of the doc tests (mostly just removing validation of the output that changes a lot depending on updates we do in our database almost monthly).

When I run the ESASky remote tests now on this branch I get five errors. All of them show an error that looks something like this:

```
pyvo.io.vosi.exceptions.W19: ?:?:?: W19: TAP Capabilities must not have an ivo-id other than ivo://ivoa.net/std/TAP
```
If I apply the fix from [the other PR](https://github.com/astropy/astroquery/pull/3586) all the ESASky remote tests pass.

I have not yet addressed the other comments you had about how to best deprecate things.

emellega:
Hello again!

I have now addressed the other comments you had, deprecating the attributes and adding an entry in the changelog.

And if I apply the fix from the [other PR](https://github.com/astropy/astroquery/pull/3586) all the ESASky remote tests pass.

bsipocz:
Ohh, I see that the pyvo warning is still hit, I'll try to look more into it maybe on the train and next week and then wrap this up and merge.

emellega:
Hello again @bsipocz!

Do you think you could have a look at this PR again? The remote tests are passing again now, I had to change one test slightly. I'm not really sure what the two failing checks are complaining about, but let me know if you think it's something I should do something about.

astroquery/esasky/tests/test_esasky_remote.py
```diff
@@ -218,35 +211,24 @@ def test_esasky_get_spectra_from_table(self, tmp_path):
...
+        assert len(result) > 0
 
     def test_get_tables(self):
         table_names = ESASky.get_tables(only_names=True)
         assert len(table_names) > 70
         tables = ESASky.get_tables(only_names=False)
-        assert isinstance(tables[0], TapTableMeta)
         assert len(table_names) == len(tables)
 
-    def test_get_columns(self):
```
bsipocz, line 242: why did you remove  this?

astroquery/esasky/core.py
```diff
@@ -1349,6 +1312,27 @@ def get_spectra_from_table(self, query_table_list, missions=__ALL_STRING,
...
+            esautils.execute_servlet_request(
+                url=conf.ESASKY_TAP_SERVER + "/" + conf.ESASKY_MESSAGES,
+                tap=self.tap,
+                query_params={},
+                parser_method=self.parse_messages_response
+            )
+        except OSError:
+            print("Status messages could not be retrieved")
+
+    def parse_messages_response(self, response):
```
bsipocz, line 1328: maybe this should be a private method?

bsipocz (commented):
Super minor comments, overall this looks good.

PR #3573, opened 2026-03-31
ESA Plato Astroquery module

jespinosaar:
Dear Astroquery Team,

Many thanks for merging PR #3511. With this, we can now proceed with the new module we already discussed.

We have been preparing a module for the PLATO Archive, to be available for the Guest Observer call for proposals, providing relevant information that may be helpful for astronomers.  For this, if possible, we would like to iterate with you and merge this branch as soon as possible, so we have a pre-released version right after this is merged. You can find more information here: 

https://www.cosmos.esa.int/web/plato/getting-ready

With this, we have prepared a module based on the EsaTap class we have already discussed, together with specific methods to retrieve data from these catalogues. As you will see, there is also a method to plot data, but we are detecting there if matplotlib is imported and, if not, it raises an exception. This method is also relevant, as it mimics the behavior of an application that is currently used, providing the same features in python script to our users.

With respect to tests, you will see that we have the offline tests and the documentation tests are set to be skipped. This is due to the URL of the TAP, currently not available for the public audience, but it will be opened next week. Right after the Archive is public, we will create another pull request to remove the SKIP clauses and add the remote tests, that we already have.

Many thanks for your support! Please let us know any comment you have.

Kind regards,
@jespinosaar 

cc @esdc-esac-esa-int


jespinosaar:
After some iterations, I managed to pass all the checks. @bsipocz , please let me know if you have any comment

jespinosaar:
Just added the latest changes in the manual.

astroquery/esa/plato/tests/setup_package.py
bsipocz, line 1: Remove this file as it is not required/used with the refactored infrastructure.
(All the `tests/data` files are included by default in the new)

astroquery/esa/plato/tests/test_plato_remote.py
bsipocz, line 1: normally I would say don't add this file until it's functional, but as it's already been included, we could keep it.

astroquery/esa/plato/core.py
```diff
@@ -0,0 +1,286 @@
...
+
+    ESA_ARCHIVE_NAME = "PLATO"
+    TAP_URL = conf.PLATO_TAP_SERVER
+    LOGIN_URL = conf.PLATO_LOGIN_SERVER
+    LOGOUT_URL = conf.PLATO_LOGOUT_SERVER
+
+    def search_catalogue(self, table_name, *, target_name=None, coordinates=None, radius=1*u.arcmin, columns=None,
+                         get_metadata=False, output_file=None, **filters):
+        """
+        Execute a search in one of the catalogues and associated tbles available in PLATO TAP.
```
bsipocz:
```suggestion
        Execute a search in one of the catalogues and associated tables available in PLATO TAP.
```

astroquery/esa/plato/core.py
```diff
@@ -0,0 +1,286 @@
...
+        -------
+        An astropy.table object containing the results of the filters in the PIC target catalogue
+        """
+        cone_search_filter = None
+        if radius is not None:
+            radius = esautils.get_degree_radius(radius)
+
+        if target_name and coordinates:
+            raise TypeError("Please use only target or coordinates as "
+                            "parameter.")
```
bsipocz:
```suggestion
            raise TypeError("Please use only 'target_name' or 'coordinates' as "
                            "parameter, not both.")
```

astroquery/esa/plato/core.py
```diff
@@ -0,0 +1,286 @@
...
+            coord = commons.parse_coordinates(coordinates=coordinates)
+            ra = coord.ra.degree
+            dec = coord.dec.degree
+            cone_search_filter = self.create_cone_search_query(ra, dec, "RAdeg", "DEdeg", radius)
+
+        return self.query_table(table_name=table_name, columns=columns,
+                                custom_filters=cone_search_filter, get_metadata=get_metadata, async_job=True,
+                                output_file=output_file, **filters)
+
+    def search_pic_target_go(self, *, target_name=None, coordinates=None, radius=1*u.arcmin, columns=None,
```
bsipocz, line 94: what does `go` mean in the name?

docs/esa/plato/plato.rst
```diff
@@ -0,0 +1,280 @@
...
+
+  >>> pic_target_table = plato.get_table(table='pic_go.pic_target_go') # doctest: +SKIP
+  >>> pic_target_table.columns # doctest: +SKIP
+
+----------------------------
+3. ADQL Queries to PLATO TAP
+----------------------------
+
+The Query TAP functionality facilitates the execution of custom Table Access Protocol (TAP)
+queries within the PLATO ArXiv. Results can be exported to a specified
```
bsipocz, line 72: Will you spell the plato archive as Arxiv?

astroquery/esa/plato/core.py
```diff
@@ -0,0 +1,286 @@
...
+                             get_metadata=False, output_file=None, **filters):
+        """
+        Execute a search in PIC Target Catalogue.
```
bsipocz: Long term this could use another sentence or two of description.

bsipocz (approved):
Overall this looks good for the archive launch. There are a few minor typos, and extra files, but none are blockers.

I would also think that the search_pic_* methods could be merged into one and have an extra kwarg that would select between target and contaminant to keep the API a bit simpler of practically duplicated functionalities. But this is not a strong preference.

astroquery/esa/plato/tests/setup_package.py
jespinosaar: I just removed this file

astroquery/esa/plato/tests/test_plato_remote.py
jespinosaar: I initially added this as a mistake, but I thought I had remove it. Now it is not included, I just removed it.

astroquery/esa/plato/core.py
jespinosaar: Thanks for spotting this, I already fixed this

astroquery/esa/plato/core.py
jespinosaar: Just rewrited this

astroquery/esa/plato/core.py
jespinosaar: GO means Guest Observer, as these methods are developed for the Guest Observer Call for Proposals

docs/esa/plato/plato.rst
jespinosaar: Yes, the acronym chosen for PLATO Archive is PAX (Plato ArXiv). I know this is different from others, but this is how it was decided.

astroquery/esa/plato/core.py
jespinosaar: I added some words with more info.

jespinosaar:
I added your comments, but it seems that when removing the setup_package file this is not working. Let me check

bsipocz:
@jespinosaar - to me it looks like the issue from the changing of the error message. 

I'm happy to fix all of these up and merge it in the next hour or so, no need to spend your time with it during Easter.

jespinosaar:
> @jespinosaar - to me it looks like the issue from the changing of the error message.
> 
> I'm happy to fix all of these up and merge it in the next hour or so, no need to spend your time with it during Easter.

Yes, I realized about that after checking, just added a new commit to fix this.

jespinosaar:
I agree, maybe it is a good idea to add an extra kwarg for the search_pic methods. But, for now, we were requested to include two separate methods. So, if it is ok with you, let's keep it like this for now.

On the other hand, it seems that now the pipeline will finish successfully.

bsipocz (approved):
OK, let's have this in for now and we can iterate as needed once the service is up and running.

astroquery/esa/plato/core.py
```diff
@@ -0,0 +1,286 @@
...
+        Returns
+        -------
+        An astropy.table object containing the results of the filters in the PIC target catalogue
+        """
+
+        return self.search_catalogue(table_name='pic_go.pic_contaminant_go', target_name=target_name,
+                                     coordinates=coordinates, radius=radius, columns=columns,
+                                     get_metadata=get_metadata, output_file=output_file, **filters)
+
+    def plot_plato_results(self, x, y, x_label, y_label, plot_title, *, z=None, z_label=None, error_x=None,
```
bsipocz, line 180: I'm not certain that this plotting belongs to astroquery, but we can clean this up later.

jespinosaar:
Many many thanks!! And sure, let's iterate on this later. Just let me know when the pre-release version is available, and so the new documentation

bsipocz:
Dev release is on pypi, `pip install --pre -U astroquery` should be able to pick it up.

jespinosaar:
Yes, everything is working as expected, many thanks!!!

PR #3577, opened 2026-04-07
PLATOWRQ-402: remote and doc tests added

jespinosaar:
Dear Astroquery team:

Now that ESA PLATO Archive has become operational, metadata is available and we want to include some remote tests and also doc testing to the module. Together with this, we are removing the message informing users that the Archive will become operationally on April 7th.

Many thanks for your support with previous PRs and, of course, with this one. We are not in a hurry for this, just wanted to consolidate and complete this module.

Kind regards,
@jespinosaar 

cc @esdc-esac-esa-int 

follow-up after #3573

### DIFF
--- a/astroquery/esa/plato/tests/test_plato_remote.py
+++ b/astroquery/esa/plato/tests/test_plato_remote.py
@@ -1,0 +1,102 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+===============
+PLATO TAP tests
+===============
+
+European Space Astronomy Centre (ESAC)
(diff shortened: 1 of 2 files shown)


jespinosaar:
As we are only adding the remote tests and updating the documentation, I am assuming we should use the no-changelog-entry-needed label. If I am wrong, please remove it.

jespinosaar:
All checks are passed now.

bsipocz (approved):
Thanks!

jespinosaar:
Many thanks!!

PR #3578, opened 2026-04-14
ESA-HST: Improve get_datalabs_path for eHST

jespinosaar:
Dear Astroquery team,

As the number of collections available in ESA HST Module has grown, we had to extend the get_datalabs_path method to consider different paths depending on the type of the product we want to retrieve.

As always, we are open to your feedback.

Many thanks!
@jespinosaar 

cc @esdc-esac-esa-int

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -118,6 +118,8 @@ esa.hubble
 
 - Update ``get_datalabs_path`` method so an alternative path is checked if the
   file is not in Datalabs yet [#3437]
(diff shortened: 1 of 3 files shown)


jespinosaar:
Checks completed successfully, please review them at your earliest convenience. Many thanks in advance for your feedback!

bsipocz (approved):
All looks good, thank you. I may merge changelog entries in a way that makes them most useful for the users, but I'll do that before release.

jespinosaar:
Many thanks @bsipocz !!!

PR #3579, opened 2026-04-15
Disable cache expiration if timeout is -1 instead of None

lpsinger:
There does not appear to be a way to set the value of `cache_timeout` to `None` in a .ini file because there is no value that will be decoded to None and will also validate as an integer.


bsipocz (approved):
Thank you @lpsinger, this makes all sense. It also makes me wonder if we should have some actual testing of the config cfg file rather than setting the values directly in the tests.

lpsinger:
> Thank you @lpsinger, this makes all sense. It also makes me wonder if we should have some actual testing of the config cfg file rather than setting the values directly in the tests.

That's a good idea. I suggest that should be a separate PR.

bsipocz:
> That's a good idea. I suggest that should be a separate PR.

Certainly! This still needs a changelog entry before merge, but I can do it now.

PR #3584, opened 2026-04-29
Fix for `tox -e build_docs`

snbianco:
The `tox -e build_docs` command has been failing for me, and I think it's because of recent updates to `setuptools`. These changes to `pyproject.toml` fixed the build for me.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,8 +6,7 @@ readme = { file = "README.rst", content-type = "text/x-rst" }
 authors = [
   {name = "The Astroquery Developers"},
 ]
-license = "BSD-3-Clause"
-license-files = ["LICENSE.rst"]
+license = { text = "BSD-3-Clause" }
 classifiers = [
   "Intended Audience :: Science/Research",
   "Operating System :: OS Independent",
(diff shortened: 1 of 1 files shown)


bsipocz:
This is weird, I don't have any issues with it neither on my default 3.12 nor with python 3.14. Can you show some tracebacks (either here or on slack).

bsipocz (commented):
We had some dive in here and concluded not to have this PR.

- I couldn't reproduce and tried with 3.11, 3.12, and 3.14; with brand new and older setuptools versions, but I wasn't systematic enough, really just let tox run

- Sam only experienced it on 3.11 but it works in here newer versions. [edit: deleting the `.tox` directory seems to do the trick] 

- there is the note on packaging documentation which I think is what we run into (https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files):

> Note
If you get a build error that license should be a dict/table, your build backend doesn’t yet support the new format. See the [above section](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files) for more context. The now deprecated format is [described in PEP 621](https://peps.python.org/pep-0621/#license).

bsipocz:
OK, closing it, but thanks for the rabbit hole dive, we clearly don't cover all the build cases and versions, but based on the python packaging documentation, what we have in the toml file should be correct.

I have patched up some of the inconsistencies we run into, but haven't dived into adding further recommendations into the narrative docs.

PR #3585, opened 2026-04-30
MAINT: some TLC for old versions

bsipocz:
We run into a couple of easy to fix issues while investigating #3584 

I still have issues directly running `sphinx-build` in some old python env, but all those python versions do work for me™️ with `tox -e build_docs`, so I stopped investigating further. 

(typical flavour of the failure I see is automodapi related with something like this:

```Extension error (sphinx_automodapi.automodsumm):
Handler <function process_automodsumm_generation at 0x123f96f80> for event 'builder-inited' threw an exception (exception: No module named 'astroquery.esa.integral')
```

I would say it probably wouldn't hurt to find and declare minimum sphinx and sphinx extension versions, too.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -62,22 +62,12 @@ dependencies required to build the documentation or run the tests):
     $ python -m pip install -U --pre astroquery[all]
 
 
-To install the 'bleeding edge' version:
+To install the 'bleeding edge' version from GitHub:
 
 .. code-block:: bash
 
    $ python -m pip install git+https://github.com/astropy/astroquery.git
 
-or cloned and installed from source:
-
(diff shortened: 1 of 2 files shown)


bsipocz:
Just for cross references, we may need to look into https://github.com/astropy/sphinx-automodapi/issues/119 and the workaround used by https://github.com/spacetelescope/roman_datamodels/pull/156?

bsipocz:
I'm not sure why we never merged this, I go ahead and do it now.

I would say we may want to actually drop 3.10 support or even 3.11, but that can be done in a follow-up PR.

PR #3586, opened 2026-05-04
Fix bug caching requests with pyvo request_hooks

emellega:
When we migrated the ESASky module to use EsaTap/PyVO we noticed some weird errors for some of our requests that we didn't have before. When downloading images through the ESASky module we got Pyvo errors saying that we didn't conform to the TAP spec. It turns out that when we do a deep copy on the response (which contains the request hooks) during caching, it triggers something within Pyvo to do that validation.

I guess this is more of a workaround around something that is broken within pyvo rather than an actual bug fix. Let me know if you find this acceptable, or if you would rather see a proper fix within pyvo.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -201,6 +201,8 @@ Infrastructure, Utility and Other Changes and Additions
 
 - Fix no expiration case for ``cache_timeout`` config option. [#3579]
 
+- Workaround upstream bug when caching a response using pyvo. [#3586]
+
 utils.tap
 ^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


bsipocz (approved):
Thank you @emellega, this looks good to me. Even if it being fixed in pyvo right now, we will need this patch until the pyvo fix is present even in the oldest supported version.

However, I think it would be better to:

1) open an upstream pyvo issue.
2) add some brief comments in the code and refer to that issue, so we could potentially clean up this logic when everything is fixed upstream.

bsipocz:
CI issues are unrelated, so let's merge this now. Thank you for the PR!

PR #3587, opened 2026-05-04
API: cleanup get_query_payload and specify coordinates

keflavich:
These are a few small steps toward correcting the API inconsistencies noted in #2803.

I think these two changes are safe.  Implementing the change of `position` or `coordinate` to `coordinates` is next.


bsipocz (approved):
Thanks!

PR #3588, opened 2026-05-07
MAST: MastMissions improvements

snbianco:
This PR includes some improvements for our `MastMissions` class:

- Results returned from `MastMissions` metadata query functions now include search parameters in the metadata of the `astropy.table.Table` object and column descriptions in the column metadata.
- Added `pass_id` as an alias for the Roman mission's `pass` column in query functions to avoid conflicts with 
  the reserved Python keyword.
- Refactored `download_products` so that when a user inputs a JSON file from the subscription service, the value of `mission` is taken from the data. This will allow users to download products from multiple missions in a single call.
- Fix in our `test_mast.py` file where a directory was being created outside of the test environment.

### DIFF
--- a/astroquery/mast/missions.py
+++ b/astroquery/mast/missions.py
@@ -15,19 +15,23 @@ from pathlib import Path
 from urllib.parse import quote
 
 import astropy.units as u
-from astropy.coordinates import SkyCoord, BaseCoordinateFrame, Angle
 import numpy as np
-from astropy.table import Table, Row, Column, vstack
+from astropy.coordinates import Angle, BaseCoordinateFrame, SkyCoord
+from astropy.table import Column, Row, Table, vstack
 from astropy.utils.decorators import deprecated_renamed_argument
 from requests import HTTPError, RequestException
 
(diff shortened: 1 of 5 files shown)


snbianco:
For your awareness: @sedonaprice @bmorris3 @ttdu

astroquery/mast/missions.py
```diff
@@ -15,19 +15,23 @@
...
+from astropy.table import Column, Row, Table, vstack
 from astropy.utils.decorators import deprecated_renamed_argument
 from requests import HTTPError, RequestException
 
 from astroquery import log
-from astroquery.utils import commons, async_to_sync
-from astroquery.utils.class_or_instance import class_or_instance
-from astroquery.exceptions import InputWarning, InvalidQueryError, MaxResultsWarning, NoResultsWarning
-
+from astroquery.exceptions import (
```
bsipocz, line 25: fyi: this was fine, our linter is set to allow longer lines

bsipocz (approved):
This all looks good, thank you!

PR #3589, opened 2026-05-11
Factor out the HITRAN reader so it can be used on pre-downloaded files

keflavich:
As in the title: the reader built into the hitran module can read downloadable .par files.  This PR exposes it as a public function.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -115,6 +115,10 @@ heasarc
 - Add automatic guessing for the data host in ``download_data``. [#3403]
 - Include method to count the number of rows in a specified table. [#3549]
 
+hitran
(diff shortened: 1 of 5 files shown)


bsipocz (approved):
This needs a changelog, but otherwise is looking good. I see a failing remote test and a failing doctest for hitran, but both are unrelated so fixing them doesn't have to be part of this PR.

keflavich:
i'll check on the remote test; hitran might have moved to login-only downloads

bsipocz:
> i'll check on the remote test; hitran might have moved to login-only downloads

They just look to be a mismatched output with now having a few more returned lines

PR #3592, opened 2026-05-14
ESA JWST Migration to PyVO

jespinosaar:
Dear Astroquery team,

This PR is an upgrade of the ESA JWST module, replacing the previous TAP implementation with the newest one based on PyVO, already applied to other Archives.

With this, code has been cleaned and classes are now making use of the EsaTap class. Together with this, now we have added the possibility of uploading tables.

Thanks to @lmasselo for doing this! Please add her to the list of contributors as well (we currently have a lot of people working on Astroquery).

Please let us know if you have any comment.

Kind regards,
jespinosa


jespinosaar:
We will work next week on increasing the coverage. In the meantime, we can start having a look at this PR, many thanks!

jespinosaar:
It seems all the checks are passing now, except for the oldest version for all dependencies. This seems to be an unrelated error.

Please let me know if you have any comment. Thanks in advance!

PR #3593, opened 2026-05-18
MAST: Read product function for MastMissions

snbianco:
A function in the `MastMissions` class that can read products into memory.

For FITS files, we use `astropy.io.fits.open`. The product files are downloaded and cached by Astropy, and then read into a `HDUList` object.

For ASDF files, we use `fsspec` and `asdf`. Our downloads for Roman data are redirected to S3, so these files are not downloaded and are streamed directly into `asdf.AsdfFile` objects.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,9 +42,9 @@ test = [
    # Temp workaround for https://github.com/RKrahl/pytest-dependency/issues/91
    "pytest-dependency; platform_system != 'Windows'",
    "pytest-rerunfailures",
-   "fsspec[http]",
+   "fsspec[http,s3]",
    "moto[s3]",
-   "s3fs",
+   "asdf"
 ]
 docs = [
    "sphinx",
@@ -62,6 +62,11 @@ all = [
    "boto3",
    "botocore",
    "regions>=0.5",
+   "asdf",
+   "lz4",
+   "gwcs",
+   "roman_datamodels",
+   "fsspec[http,s3]"
 ]
 
 [build-system]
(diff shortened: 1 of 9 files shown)


scfleming:
Until we have a more mature version ready, can we please ensure documentation / tutorials call out that for FITS files `read_product()` still downloads the file locally, so people working on Roman work who may want to use this to download HST or JWST FITS files aren't surprised if `read_product()` is using up credits writing to local cloud storage on them?

I presume this can be avoided once there's a "get_s3_uris()" endpoint in MissionMAST for all the supported missions beyond Roman?

pyproject.toml
```diff
@@ -54,13 +54,19 @@ docs = [
    "sphinx-astropy>=1.5",
    "scipy",
    "fsspec[http]",
+   "asdf"
```
bsipocz: where is it used in the docs?

pyproject.toml
```diff
@@ -42,9 +42,9 @@ test = [
    # Temp workaround for https://github.com/RKrahl/pytest-dependency/issues/91
    "pytest-dependency; platform_system != 'Windows'",
    "pytest-rerunfailures",
-   "fsspec[http]",
+   "fsspec[http,s3]",
    "moto[s3]",
-   "s3fs",
+   "asdf"
```
bsipocz: I don't think that optional dependencies should be installed for all test runs. The fact that we have fsspec and s3fs listed here is a problem of its own and should be cleaned up. So, I would not think we should add asdf.

bsipocz (changes requested):
code changes look good, but I have problem with the proposed changes in pyproject.toml, IMO we should avoid workaround for `[all]` and should not include optional dependencies to be installed for all the test jobs.

pyproject.toml
snbianco: I think I thought I had to include it for the code blocks. Removed!

PR #3596, opened 2026-05-19
MAINT: Pin `pytest-rerunfailures` in oldest deps

snbianco:
The "oldest version for all dependencies" CI job has been failing since May 13th. I tracked this down to a release of `pytest-rerunfailures` that requires a newer version of pytest. Pinning the version to 16.1 for `oldestdeps` makes the job pass again.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -48,6 +48,7 @@ deps =
     oldestdeps: keyring==23.0
     oldestdeps: pytest==7.4
     oldestdeps: beautifulsoup4==4.10
+    # Newer versions would bump the pytest version, revise the pin when we revise pytest
     oldestdeps: pytest-rerunfailures==16.1
     oldestdeps-alldeps: mocpy==0.12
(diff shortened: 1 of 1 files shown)


bsipocz:
Ahh, good catch. Fixing that job was on my list.

tox.ini
```diff
@@ -48,6 +48,7 @@ deps =
     oldestdeps: keyring==23.0
     oldestdeps: pytest==7.4
     oldestdeps: beautifulsoup4==4.10
+    oldestdeps: pytest-rerunfailures==16.1
```
bsipocz, line 52:
```suggestion
    # Newer versions would bump the pytest version, revise the pin when we revise pytest
    oldestdeps: pytest-rerunfailures==16.1
```

bsipocz (approved):
Thanks! I'll merge once the oldestdeps job is passing, anything else is irrelevant here

PR #3598, opened 2026-05-22
MAST: Fix for deprecation of objectname in Observations

snbianco:
Previously, the `objectname` parameter was not properly deprecated in `Observations.query_criteria` and `Observations.query_criteria_count`.

### DIFF
--- a/astroquery/mast/observations.py
+++ b/astroquery/mast/observations.py
@@ -6,33 +6,38 @@ MAST Observations
 This module contains various methods for querying MAST observations.
 """
 
-from pathlib import Path
-import warnings
-import time
 import os
+import time
+import warnings
+from pathlib import Path
 from urllib.parse import quote
 
-import numpy as np
-import astropy.units as u
 import astropy.coordinates as coord
-from requests import HTTPError
-from astropy.table import Table, Row, vstack
+import astropy.units as u
+import numpy as np
+from astropy.table import Row, Table, vstack
 from astropy.utils.decorators import deprecated_renamed_argument
+from requests import HTTPError
 
 from astroquery import log
 from astroquery.mast.cloud import CloudAccess
 from astroquery.utils import commons
 
+from ..exceptions import (
+    CloudAccessWarning,
+    InputWarning,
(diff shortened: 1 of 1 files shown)


sedonaprice:
@bsipocz as a heads up, this fix is necessary for some MAST notebooks (selected for being pulled into Fornax) to run in a unified MAST environment.  We're hoping to have these ready and deployed in time for the AAS workshop.

astroquery/mast/observations.py
```diff
@@ -332,13 +337,23 @@ def query_object_async(self, object_name, *, radius=0.2*u.deg, pagesize=None, pa
         return self.query_region_async(coordinates, radius=radius, pagesize=pagesize, page=page)
 
     @class_or_instance
-    def query_criteria_async(self, *, pagesize=None, page=None, resolver=None, **criteria):
+    @deprecated_renamed_argument('objectname', 'object_name', since='0.4.12')
+    def query_criteria_async(self, *, coordinates=None, object_name=None, radius=0.2*u.deg,
+                             pagesize=None, page=None, resolver=None, **criteria):
         """
```
bsipocz, line 343: The signature is changed significantly, can I assume that all the new parameters were just in fact hidden previous in ``criteria``?

(for the record I am very much in favour of spelling out everything rather than hide in `**kwarg`

bsipocz (approved):
All looks good, thank you.

I'm on the fence if we need to mention this PR in the chaneglog or not, I'll decide when I do a changelog cleanup before tagging the release.

bsipocz:
This should fix the issue we run into with TESS queries in the fornax notebooks; xref https://github.com/nasa-fornax/fornax-demo-notebooks/issues/682

astroquery/mast/observations.py
snbianco: Yes, all the parameters that are now explicitly named were already expected by the function.

bsipocz: are there anything else left in `**criteria`? If not, then maybe we can also remove it so it won't keep swallowing unused argument names.

PR #3601, opened 2026-06-03
Euclid: refactor method get_scientific_product_list so that parameter dsr_part3 accepts the value 'latest'

cosmoJFH:
Dear astroquery team,

We have opened this PR to update and refactor the get_scientific_product_list method.

The main changes are:

1. The method now queries the sedm.level_3 table instead of sedm.basic_download_data.
2. The dsr_part3 parameter now supports the additional value _latest_.
3. A new method, get_valid_le3_configuration_values, has been introduced to retrieve the valid values for the category, group, and product_type parameters dynamically. It returns an astropy Table.
4. Since get_valid_le3_configuration_values provides up-to-date values for these parameters, the VALID_LE3_PRODUCT_TYPES_CATEGORIES_GROUPS dictionary is removed.
5. For the same reason, the appendix describing the valid Level 3 configuration values has been removed from the documentation.

We would appreciate any feedback or suggestions you may have.

cc @esdc-esac-esa-int 
jira: EUCLIDPCR-2131


astroquery/esa/euclid/core.py
```diff
@@ -1307,31 +1309,40 @@ def get_product_list(self, *, observation_id=None, tile_index=None, product_type
...
+            dsr_1_value: Optional value for the first dataset release field.
+            dsr_2_value: Optional value for the second dataset release field.
+            dsr_3_value: Optional value for the third dataset release field.
+                         If set to "latest", the clause filters on latest = true.
+            alias: Optional table alias prepended to field names.
+
+        Returns:
+            A string containing SQL conditions joined with AND, or None if no
+            filter values are provided.
+        """
```
bsipocz, line 1326: we don't render this into any docs pages, so it doesn't matter, but in general please use numpydoc style for docstrings

bsipocz (approved):
Thank you, this looks all good.

PR #3602, opened 2026-06-03
RTD: Fetch all tags for accurate version reporting

pllim:
Fix https://github.com/astropy/astroquery/issues/3565

No need for CI, just need RTD.

Before: https://astroquery.readthedocs.io/en/latest/ (`astroquery v0.1.dev329+g1ad1025d5`)

After: https://astroquery--3602.org.readthedocs.build/en/3602/ (`astroquery v0.4.12.dev532+g998ead2a5`)

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -6,6 +6,9 @@ build:
     python: "3.11"
   apt_packages:
     - graphviz
+  jobs:
+    post_checkout:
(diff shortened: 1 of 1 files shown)


bsipocz (approved):
Thanks. I can't recall why we haven't opted into this at the same time we did for pyvo.

PR #3606, opened 2026-06-04
update the NED to use the new API

cwang2016:
Updates on NED web service calls including the parts of the core functions,  testing and documentation based on NED N36.1 release
The changes are made on astroquery/ipac/ned/core.py, astroquery/ipac/ned/tests/, and docs/ipac/ned/.


cwang2016:
Regarding the ReadTheDocs build failure: no warnings or errors specific to docs/ipac/ned/ned.rst were found locally (make html SPHINXOPTS="-W" returned clean). Could this be related to a pre-existing issue rather than changes in this PR? Please advise.

bsipocz:
I get that warning locally, too and I've double checked that we don't really expose the exceptions into the public API.

docs/ipac/ned/ned.rst
```diff
@@ -223,45 +240,71 @@ Similarly the list of URLs for spectra of a particular object may be fetched:
...
+    ...            ...             ... ...          ...       ...                 ...      ...
+    186	12h29m06.7000s	+02d03m08.700s ...           --        -- 2014ApJ...789..135W	    --
+    187	12h29m06.7000s	+02d03m08.600s ...           --        -- 2014MNRAS.439..690H	    --
+    188	12h29m06.7018s	+02d03m08.604s ...      0.50000	        0 2014MNRAS.439.2701H       --
+    189	12h29m06.7000s	+02d03m09.000s ...           --        -- 2014MNRAS.441.3375X	    --
+    190	12h29m06.8017s	+02d03m08.247s ...      1.17525	        0 2012GASC..C...0000S	    --
+    Length = 190 rows
+
+All above queries return results in a `~astropy.table.Table` or raise a Exception error (i.e.
+`~astroquery.exceptions.RemoteServiceError`) if the service returns a query error.
```
bsipocz: we don't expose this in the public API, you can either keep it within double backticks or remove the parenthesis altogether.

docs/ipac/ned/ned.rst
```diff
@@ -223,45 +240,71 @@ Similarly the list of URLs for spectra of a particular object may be fetched:
...
+
+    >>> from astroquery.ipac.ned import Ned
+    >>> try:
+    >>>     result_table = Ned.query_object("mm")
+    >>>     print(result_table) 
+    >>> except Exception as e:
+    >>>     print(e)
+    The remote service returned the following message.
+    ERROR: GeneralFault:
+    Service could not complete request; Failed to resolve input object name (6)
```
bsipocz: I think we can/should drop this section; we don't need to teach users basic error/warning handling. Though, if you think it's necessary for NED users, then I would recommend to show an example where you except on specific Exceptions rather than caption it too generically (which we never recommend as a good coding practice)

bsipocz (commented):
I still need to do actual code review, and will also suggest to squash out a few commits (the ones that added and then removed the bigger test files); but here is just the fix for the docs builld issue

docs/ipac/ned/ned.rst
cwang2016: yes, this part is removed.

docs/ipac/ned/ned.rst
cwang2016: fixed per the comment.

astroquery/ipac/ned/tests/test_ned.py
```diff
@@ -65,219 +68,423 @@ def get_readable_fileobj_mockreturn(filename, cache=True, encoding=None,
...
-                        'extend': 'no',
-                        'img_stamp': 'NO',
-                        'list_limit': 0,
-                        'of': 'xml_main'
-                        }
-    response = ned.core.Ned.query_refcode_async('1997A&A...323...31K')
+    response = ned.Ned.query_refcode_async('1997A&A...323...31K',
+                                           get_query_payload=True)
+    s_type = ned.Ned.SEARCH_TYPE
+    # assert response == {'search_type': 'Search',
```
bsipocz: you can remove this rather than just comment out

astroquery/ipac/ned/core.py
```diff
@@ -73,278 +102,387 @@ def query_object(self, object_name, *, get_query_payload=False,
...
+        self.SEARCH_TYPE = "object"
 
-    def query_region(self, coordinates, *, radius=1 * u.arcmin, equinox='J2000.0',
-                     get_query_payload=False, verbose=False):
+        return self.handle_response(get_query_payload,
+                                    self.DBR_BASE_URL + self.CONESEARCH_TARGET,
+                                    request_payload)
+
+    def query_region(self, coordinates, *, radius=1 * u.arcmin,
+                     get_query_payload=False, verbose=False, **kwargs):
```
bsipocz: what are the possible values for **kwargs? only those 4 options listed in the docstring? If yes, I would recommend to add those directly, so we don't just swallow invalid options but error out on them early.

astroquery/ipac/ned/core.py
```diff
@@ -73,278 +102,387 @@ def query_object(self, object_name, *, get_query_payload=False,
...
-            return request_payload
-        response = self._request("GET", url=Ned.OBJ_SEARCH_URL,
-                                 params=request_payload, timeout=Ned.TIMEOUT)
-        response.raise_for_status()
-        return response
+        request_payload = self._request_payload_init(target_name=object_name)
+        request_payload[self.DBR_RADIUS] = 0
+        self.SEARCH_TYPE = "object"
 
-    def query_region(self, coordinates, *, radius=1 * u.arcmin, equinox='J2000.0',
```
bsipocz, line 104: It maybe worth deprecating equinox instead of removing it straight away

astroquery/ipac/ned/core.py
```diff
@@ -73,278 +102,387 @@ def query_object(self, object_name, *, get_query_payload=False,
...
+                ``Unavailable``, ``Larger Than``, ``Less Than`` or ``Between``.
+            z_value1, z_value2: float, optional.
+                z_value1 is the redshift bound for the constraint
+                ``Larger Than``, ``Less Than``, or the minimum value within
+                a range for ``Between``.
+                z_value2 is the maximum value within a range for ``Between``.
+            z_unit: str, optional
+                redshift value unit may be ``z`` or ``km/s``.
+                Defaults to ``z``.
+            max_rec: int, optional
```
bsipocz: I think almost everywhere we use `maxrec` rather than `max_rec`. Other option could be `row_limit`. (maxrec is  feel a very ivoa specific jargon)

astroquery/ipac/ned/core.py
```diff
@@ -73,278 +102,387 @@ def query_object(self, object_name, *, get_query_payload=False,
 
     def query_object_async(self, object_name, *, get_query_payload=False):
```
bsipocz, line 103: We have all these *_async methods, most of them are not actually async. In a future PR / cleanup operation, we may want to revisit to see if things could be cleaned up.

astroquery/ipac/ned/core.py
```diff
@@ -530,210 +686,417 @@ def _extract_image_urls(self, html_in, *, file_format='fits'):
...
+            max_rec=self._get_maxrec(**kwargs))
+        if table == 'photometry' and "is_line" in kwargs and kwargs['is_line']:
+            request_payload[self.DBR_ISLINE] = 'yes'
+        self.SEARCH_TYPE = _SEARCH_TYPE[table]
+
+        return self.handle_response(get_query_payload,
+                                    self.DBR_BASE_URL + _SEARCH_TYPE[table],
+                                    request_payload)
+
+    def handle_response(self, get_query_payload, service_url, request_payload):
```
bsipocz: I suppose this could be private

astroquery/ipac/ned/core.py
```diff
@@ -530,210 +686,417 @@ def _extract_image_urls(self, html_in, *, file_format='fits'):
...
+            tf = BytesIO(response.content)
+            p_table = votable.parse(tf, verify='warn')
+            first_table = p_table.get_first_table()
+            table = first_table.to_table(use_names_over_ids=True)
+            return table
+        except (ValueError, TypeError, AttributeError,
+                IndexError, ExpatError) as ex:
+            _raise_service_or_parse_error(ex)
+        except Exception as ex:
+            _raise_service_or_parse_error(ex)
```
bsipocz: why do we need these two cases? Also, it's not that many lines, maybe it could directly go here rather having as a separate method

astroquery/ipac/ned/core.py
```diff
@@ -530,210 +686,417 @@ def _extract_image_urls(self, html_in, *, file_format='fits'):
...
 
 
 Ned = NedClass()
 
 
+def get_value_from_paths(obj, attr_paths):
```
bsipocz: Does this have to be a public function?

astroquery/ipac/ned/core.py
```diff
@@ -530,210 +686,417 @@ def _extract_image_urls(self, html_in, *, file_format='fits'):
...
+        try:
+            for attr in attr_path.split('.'):
+                obj = getattr(obj, attr)
+            return obj
+        except AttributeError:
+            continue
+    return None
+
+
+def get_coord_for_ned(c):
```
bsipocz: Same, does it needs to be public?

bsipocz (commented):
Some minor comments, once rebased I think this should pass the RTD build

astroquery/ipac/ned/core.py
cwang2016: Is there specific measure to do 'deprecating'?

bsipocz: We usually do it with the `deprecated_renamed_argument` decorator, example usage is all over astroquery, for example this one: https://github.com/astropy/astroquery/blob/main/astroquery/ipac/irsa/core.py#L211

astroquery/ipac/ned/core.py
cwang2016: changed

astroquery/ipac/ned/core.py
cwang2016: changed

astroquery/ipac/ned/core.py
cwang2016: updated

astroquery/ipac/ned/core.py
cwang2016: yes

astroquery/ipac/ned/core.py
cwang2016: yes, maxrec

astroquery/ipac/ned/core.py
cwang2016: yes. thanks

cwang2016: this optional keyword is kept.

astroquery/ipac/ned/tests/test_ned.py
cwang2016: yes

astroquery/ipac/ned/core.py
```diff
@@ -521,219 +719,471 @@ def _extract_image_urls(self, html_in, *, file_format='fits'):
...
+        try:
+            for attr in attr_path.split('.'):
+                obj = getattr(obj, attr)
+            return obj
+        except AttributeError:
+            continue
+    return None
+
+
+def _get_coord_for_ned(c, *, equinox=None, isStrCoord=False):
```
bsipocz: Super minor, but please don't mix and match variable styling
```suggestion
def _get_coord_for_ned(c, *, equinox=None, is_str_coord=False):
```

astroquery/ipac/ned/core.py
```diff
@@ -73,278 +102,430 @@ def query_object(self, object_name, *, get_query_payload=False,
...
         else:
             try:
                 c = commons.parse_coordinates(coordinates)
-                if c.frame.name == 'galactic':
-                    request_payload['in_csys'] = 'Galactic'
-                    request_payload['lon'] = c.l.degree
-                    request_payload['lat'] = c.b.degree
-                # for any other, convert to ICRS and send
+                isStrCoord = isinstance(coordinates, str)
+                ra, dec, equ, frame = _get_coord_for_ned(c, equinox=equinox, isStrCoord=isStrCoord)
```
bsipocz:
```suggestion
                is_str_coord = isinstance(coordinates, str)
                ra, dec, equ, frame = _get_coord_for_ned(c, equinox=equinox, is_str_coord=is_str_coord)
```

astroquery/ipac/ned/core.py
```diff
@@ -521,219 +719,471 @@ def _extract_image_urls(self, html_in, *, file_format='fits'):
...
+        # For string-form coordinates outside FK4/FK5, parse the requested equinox,
+        # transform to FK5 for RA/Dec extraction, and keep equinox in payload.
+        # For non-string inputs or mismatched frame/equinox pairs, normalize to
+        # FK5/J2000 before building the request payload.
+        if frame_name in ('fk4', 'fk5'):
+            def_equ = NED_COORD_EQUINOX['b'] \
+                if frame_name == 'fk4' else NED_COORD_EQUINOX['j']
+            equ = _get_equinox(c)
+        else:
+            if isStrCoord:
```
bsipocz:
```suggestion
            if is_str_coord:
```

bsipocz (approved):
Two of the remote tests are running into timeouts, and there are a few doctest issues in the docs, too. But I don't think those should hold up the merging. 

I have one minor comment that I'll push in a new commit and will also go ahead and rebase the PR to squash out the data file changes; and then go ahead with the merge.

Thank you @cwang2016 for working on this, it's really nice to have NED on board now and to get rid of yet another reverse engineered module.

PR #3608, opened 2026-06-05
Update Zcut request parameter

snbianco:
One of the request parameters in our Zcut service was recently changed, making Astroquery return unexpected results. This is a quick fix.


bsipocz (approved):
Looks good, thanks!

PR #3609, opened 2026-06-10
MAINT: standardize spatial query parameter name to 'coordinates'

keflavich:
(this is an experiment @bsipocz - take with a heavy grain of salt, I tried out new claude fable this morning.  My first-pass review says this is reasonable: it's actually pretty limited in scope, just changing coordinate->coordinates for the most part.)

Part of the API consolidation tracked in #2803: the first argument of the query_region/query_object/cone_search family is now uniformly named 'coordinates'. The old names ('coordinate' in alma, esa.euclid, esa.hsa, esa.jwst, gaia, ipac.irsa.ibe, noirlab; 'position' in esasky and heasarc; 'coord' in ogle) keep working via
deprecated_renamed_argument shims (since 0.4.12) and now emit AstropyDeprecationWarning.

In heasarc, the renamed parameter shadowed the 'from astropy import coordinates' module import, so that import was narrowed to the names actually used.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -64,8 +64,6 @@ esa.euclid
    product_type parameters dynamically. [#3601]
 - In the method, ``get_scientific_product_list``, the ``dsr_part3`` parameter now supports the additional value
    ``latest``. [#3601]
-- The ``coordinate`` keyword of ``query_object``, ``cone_search``, and ``get_cutout`` is deprecated in
-  favor of ``coordinates`` for consistency with the rest of astroquery. [#3609]
 
 
 vizier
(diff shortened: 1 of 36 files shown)


bsipocz:
Hmm, that is quite a few prominent modules, I would say if we do this it should have a very long deprecation period.

OTOH, we should also make sure if we change the parameter to plural then all these modules indeed work with a vector SkyCoord object.

Anyways; I'm milestoning it as "API clenaup" -- I would think that could be a release where we also break the versioning scheme and also do other API cleanups, so we break things only once.

keflavich:
Good idea.  I can mark it all deprecated and check whether they take vectors.

Should we formalize the API such that any services that _don't_ take a vector use `coordinate` while those that take one or more use `coordinates`? I think that makes sense.

bsipocz:
> Should we formalize the API such that any services that don't take a vector use coordinate while those that take one or more use coordinates? I think that makes sense.

I think the astroquery API should always take the vectors and if there is no server side support we should just loop it in the module.

keflavich:
OK, I'm happy enough to do that.  I think we've avoided it in some cases to avoid spamming the services.  Should we impose a default (local) rate limit?

bsipocz:
> OK, I'm happy enough to do that. I think we've avoided it in some cases to avoid spamming the services. Should we impose a default (local) rate limit?

In those cases we could/should reach out and ask about their preferences. The spamming was prblematic when the service were supporting vector queries yet we did it in a loop of signletons

keflavich:
Yeah.... we can reach out, but for first pass, if we add a looped-singleton implementation where before there was none, I'd like to impose a default maximum number of parallel requests or a minimum time between requests. Are you ok with me doing that arbitrarily?  It would be nice to clean up the API in a uniform sweep and handle archive detailed preferences later.

PR #3610, opened 2026-06-12
MAST: Adding read_product function to mast.observations

AlexReedy:
A function of the Mast.Observations class to read product from S3 to memory.

Observations.read_product() takes both FITS and ASDF file URIs and stores them using astropy.io.fits.open and asdf.open respectively. 

Connection to S3 is done using fsspec

Changes made
1. Mast/observations - added read_product() function and import checks for asdf and fsspec
2. Mast/test - testing suite for read_product that mocks S3 connection and data and checks functionality (test_mast.py) as well as tests live connection to S3 and confirms product type return.


astroquery/mast/tests/test_mast.py
```diff
@@ -1372,6 +1372,72 @@ def test_observations_disable_cloud_dataset(patch_boto3):
...
 
 
+@pytest.fixture
+def mock_fits_open(mocker):
+    """Mock fits.open to return a valid HDUList without network access."""
+    return mocker.patch("astropy.io.fits.open", return_value=fits.HDUList([fits.PrimaryHDU()]))
+
+
+@pytest.fixture
+def mock_asdf_open(mocker):
```
snbianco, line 1389: If you add:
```
pytest.importorskip("asdf")
pytest.importorskip("fsspec")
```
tests that use this fixture will automatically skip if `asdf` or `fsspec` are not installed. I think this would be preferable to mocking `mock_fsspec_open`.

astroquery/mast/tests/test_mast.py
```diff
@@ -1372,6 +1372,72 @@ def test_observations_disable_cloud_dataset(patch_boto3):
...
+
+def test_observations_read_product_fsspec_missing(monkeypatch):
+    # Forces fsspec to be None
+    monkeypatch.setitem(Observations.read_product.__globals__, "fsspec", None)
+
+    with pytest.raises(ImportError, match="fsspec"):
+        Observations.read_product("file.fits")
+
+
+def test_observations_read_product_asdf_missing(monkeypatch):
```
snbianco: Could you combine this test with the one before it using parametrization?

astroquery/mast/observations.py
```diff
@@ -1235,6 +1246,72 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+            FITS or ASDF object for the given data product.
+        """
+        # Checks if a path is empty or None.
+        if not product_path or not str(product_path).strip():
+            raise ValueError("No product path provided")
+
+        # Forces the path to be lowercase for the extension checks. This is only used for the checks
+        path = str(product_path).lower()
+
+        # Checks users enviornment for fsspec, required for both fits and asdf
```
snbianco:
```suggestion
        # Checks users environment for fsspec, required for both fits and asdf
```

astroquery/mast/observations.py
```diff
@@ -1235,6 +1246,72 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+            raise ImportError('The "fsspec" package is required to read products directly from a URI. '
+                              'Please install it with `pip install fsspec`.')
+
+        # Logic for reading FITS files
+        if path.endswith((".fits", ".fits.gz")):
+            try:
+                data_product = fits.open(product_path, fsspec_kwargs={"anon": True}, **kwargs)
+                log.info(f"Loaded: {product_path}")
+                return data_product
+            except Exception as e:
```
snbianco, line 1290: Can you be more specific about the exception here? Or if you want to catch generally, add a comment explaining why.

astroquery/mast/observations.py
```diff
@@ -1235,6 +1246,72 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+                raise RuntimeError(f"Failed to open FITS File: {product_path} {e}")
+
+        # Logic for reading ASDF files
+        elif path.endswith(".asdf"):
+            if fsspec is None:
```
snbianco: Since this package is needed for both FITS and ASDF, can you check that it exists in the environment outside of this block so that it's not there twice?

astroquery/mast/observations.py
```diff
@@ -1235,6 +1246,72 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+                raise ImportError('The "fsspec" package is required to read products directly from a URI. '
+                                  'Please install it with `pip install fsspec` or install astroquery with '
+                                  'optional dependencies using `pip install astroquery[all]`.')
+            if asdf is None:
+                raise ImportError('The "asdf" package is required to read ASDF files. Please install it with '
+                                  '`pip install asdf` or install astroquery with optional dependencies using '
+                                  '`pip install astroquery[all]`.')
+
```
snbianco, line 1299: I'd also like to raise warnings if the optional packages are not found. Something like this:

```
for package in ["gwcs", "lz4", "roman_datamodels"]:
                    if importlib.util.find_spec(package) is None:
                        warnings.warn(f'The "{package}" package is encouraged when reading ASDF files from the '
                                      'Roman Space Telescope mission. Please install it with `pip install {package}` '
                                      'or install astroquery with optional dependencies using '
                                      '`pip install astroquery[all]`.', ImportWarning)
```

astroquery/mast/observations.py
```diff
@@ -1235,6 +1246,72 @@ def get_unique_product_list(self, observations, *, batch_size=500):
...
+            # Attempts to open the asdf files
+            try:
+                f = fsspec.open(product_path, "rb", anon=True).open()
+                data_product = asdf.open(f, ignore_unrecognized_tag=ignore_unrecognized)
+                log.info(f"Loaded: {product_path}")
+                return data_product
+            except Exception as e:
```
snbianco, line 1313: Use a more specific exception here if you can

docs/mast/mast_obsquery.rst
```diff
@@ -629,3 +629,15 @@ remain fully cloud-based.
...
+Streaming Data Products from S3 to memory
+-----------------------------------------
+If instead of downloading you would like to load an S3 URI directly to memory, you can use the `~astroquery.mast.ObservationsClass.read_product` method. 
+This function supports FITS and ASDF data products and will automatically parse the file for the suffix and load it to memory using `~astropy.io.fits.open` or `~asdf.open`.
+For ASDF data products, additional packages may be required such as ``roman_datamodels``, ``gwcs`` and ``lz4``.  To install astroquery with all these, and all other optional dependencies, use ``pip install astroquery[all]``.
+
+.. doctest-remote-data::
+
+   >>> from astroquery.mast import Observations
+   >>> fits_product = Observations.read_product(product_path="s3://stpubdata/hst/public/u9o4/u9o40504m/u9o40504m_c3m.fits")
```
snbianco, line 644: Can you add a `fits_product.info()` and `asdf_product.info()` just to demonstrate that actual objects with data are being returned?

snbianco (commented):
I had a few comments, but this looks good!

tox.ini
```diff
@@ -53,6 +53,9 @@ deps =
     oldestdeps-alldeps: mocpy==0.12
     oldestdeps-alldeps: regions==0.5
     oldestdeps-alldeps: astropy-healpix==0.7
+    oldestdeps-alldeps: roman_datamodels==0.11.0
+    oldestdeps-alldeps: gwcs==0.18
```
bsipocz, line 57: gwcs is not listed as a dependency, so including it here makes little sense to me?

But also, if we add all the 3 packages as optional dependencies we will need to make sure they don't introduce some indirect limitations on supported versions.

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -1048,6 +1053,23 @@ def test_observations_get_cloud_uris_no_duplicates(self, msa_product_table, rese
...
+    @pytest.mark.parametrize(
+        "product_path, expected_type",
+        [
+            (
+                "s3://stpubdata/hst/public/u24r/u24r0102t/u24r0102t_c1f.fits",
+                fits.HDUList,
+            )
+        ],
+    )
+    def test_observations_read_product(self, product_path, expected_type):
```
bsipocz, line 1066: we allow long lines in astroquery, please reformat it back to something less verbose and whitespace-y

astroquery/mast/tests/test_mast_remote.py
```diff
@@ -18,6 +18,11 @@
 from ...exceptions import (InputWarning, InvalidQueryError, MaxResultsWarning, NoResultsWarning)
 from ..utils import ResolverError
 
+try:
+    import asdf
+except ImportError:
+    asdf = None
+
```
bsipocz, line 25: You already do skipifs, so is this really necessary? or maybe you can use @pytest.mark.skipif instead of the importorskip, so it's more consistent

bsipocz (commented):
please rebase, and once I have a passing review from Sam I'll come back and do a final review.

astroquery/mast/tests/test_mast_remote.py
snbianco: Brigitta is right, we actually don't need this block for the tests to pass.

tox.ini
snbianco: The reason we do this is that the oldestdeps test suite fails if these two packages are not pinned. I think it was some issue with Numpy or the versions of other packages defined. `gwcs` is a dependency of `roman_datamodels`, which is why it's installed.

astroquery/mast/observations.py
snbianco: I think we talked about not wrapping this in a try-except at some point. I still think we should do that rather than re-raise.

astroquery/mast/observations.py
snbianco: See other comment.

PR #3611, opened 2026-06-14
EUCLID: new parameter schema in the methods get_product_list and get_scientific_product_list

cosmoJFH:
Dear astroquery team,

We have opened this PR to update the get_product_list and get_scientific_product_list methods, so they accept the new parameter _schema_.

The main changes are:

1. In the original implementation, the _sedm_ schema is used as the default schema for the tables defined in the previous two functions. In the future, the Euclid archive will distinguish products based on the schema name.
2. Fixed a bug in the __get_tile_catalogue_list method that prevented it from accepting the parameters dsr_part1, dsr_part2, and dsr_part3.

We would appreciate any feedback or suggestions you may have.

cc @esdc-esac-esa-int 
jira: EUCLIDSWRQ-292

### DIFF
--- a/astroquery/esa/euclid/__init__.py
+++ b/astroquery/esa/euclid/__init__.py
@@ -71,8 +71,6 @@ class Conf(_config.ConfigNamespace):
 
     PRODUCT_TYPES = ['observation', 'mosaic']
 
-    SCHEMAS = ['sedm']
-
     VALID_DATALINK_RETRIEVAL_TYPES = ['SPECTRA_BGS', 'SPECTRA_RGS']
 
     VALID_LINKING_PARAMETERS = {'SOURCE_ID', 'SOURCEPATCH_ID'}
--- a/astroquery/esa/euclid/tests/test_euclidtap.py
+++ b/astroquery/esa/euclid/tests/test_euclidtap.py
@@ -726,8 +726,6 @@ def test_not_overlaping_values_in_product_groups():
         "MOSAIC_PRODUCTS": conf.MOSAIC_PRODUCTS
     }
 
-    assert len(product_groups) == 9
-
     seen = {}
 
     for group_name, products in product_groups.items():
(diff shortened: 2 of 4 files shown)


astroquery/esa/euclid/tests/test_euclidtap.py
```diff
@@ -705,6 +705,41 @@ def test_get_product_list():
...
+        "MER_SEGMENTATION_MAP_PRODUCTS": conf.MER_SEGMENTATION_MAP_PRODUCTS,
+        "CALIBRATED_FRAME_PRODUCTS": conf.CALIBRATED_FRAME_PRODUCTS,
+        "FRAME_CATALOG_PRODUCTS": conf.FRAME_CATALOG_PRODUCTS,
+        "COMBINED_SPECTRA_PRODUCTS": conf.COMBINED_SPECTRA_PRODUCTS,
+        "SIR_SCIENCE_FRAME_PRODUCTS": conf.SIR_SCIENCE_FRAME_PRODUCTS,
+        "MOSAIC_PRODUCTS": conf.MOSAIC_PRODUCTS
+    }
+
+    assert len(product_groups) == 9
+
```
bsipocz: I don't think this is needed

```suggestion
```

astroquery/esa/euclid/__init__.py
```diff
@@ -71,8 +71,6 @@ class Conf(_config.ConfigNamespace):
 
     PRODUCT_TYPES = ['observation', 'mosaic']
 
-    SCHEMAS = ['sedm']
```
bsipocz, line 74: Is there any scenario the user could have interacted with this config item? If yes, the bare minimum is to mention the removal from the config into the changelog.

bsipocz (approved):
Overall looks good, I just have one minor API change/changlog related question about the removed config item

astroquery/esa/euclid/tests/test_euclidtap.py
cosmoJFH: I agree with you. I have removed the line.

astroquery/esa/euclid/__init__.py
cosmoJFH: The variable SCHEMAS was never used. I think it was left over from the original implementation.

bsipocz:
Thanks @cosmoJFH!

Issue #3612, opened 2026-06-15
BUG: new pytest warnings fails CI


bsipocz:
This should have been picked up by CI earlier, so fix the issues, but also check into why this slipped through.

```
ERROR ../../.tox/py314-test/lib/python3.14/site-packages/astroquery/skyview/tests/test_skyview_remote.py::TestSkyviewRemote - pytest.PytestRemovedIn10Warning: Passing a non-Collection iterable to parametrize is deprecated.
```

And also this:

```
Please convert to a list or tuple.
See https://docs.pytest.org/en/stable/deprecations.html#parametrize-iterators
ERROR ../../.tox/py314-test/lib/python3.14/site-packages/astroquery/linelists/jplspec/tests/test_jplspec_remote.py::TestRegressionAllMolecules - pytest.PytestRemovedIn10Warning: Passing a non-Collection iterable to parametrize is deprecated.
Test: .tox/py314-test/lib/python3.14/site-packages/astroquery/linelists/jplspec/tests/test_jplspec_remote.py::TestRegressionAllMolecules::test_regression_all_molecules, argvalues type: Table
Please convert to a list or tuple.
See https://docs.pytest.org/en/stable/deprecations.html#parametrize-iterators
ERROR ../../.tox/py314-test/lib/python3.14/site-packages/astroquery/skyview/tests/test_skyview_remote.py::TestSkyviewRemote - pytest.PytestRemovedIn10Warning: Passing a non-Collection iterable to parametrize is deprecated.
Test: .tox/py314-test/lib/python3.14/site-packages/astroquery/skyview/tests/test_skyview_remote.py::TestSkyviewRemote::test_survey, argvalues type: zip
Please convert to a list or tuple.
See https://docs.pytest.org/en/stable/deprecations.html#parametrize-iterators
!!!!!!!!!!!!!!!!!!! Interrupted: 5 errors during collection !!!!!!!!!!!!!!!!!!!!

```

PR #3613, opened 2026-06-15
CI: testing with pytest dev version

bsipocz:
to smoke out issues sooner rather than later to avoid issues like #3612 
 
 I would also think that we should stop dev testing on the PRs, and should just do that on merge and cron as there is only a very small chance of a regressing PR. And if it happens we can ask for a follow-up PR. And to be sure I can add a label trigger point, too.


keflavich:
That's fine by me.  Not testing dev on each PR is just to get them to run faster / avoid burning extra CPU cycles?

bsipocz:
> Not testing dev on each PR is just to get them to run faster / avoid burning extra CPU cycles?

No, I don't think this is significant, but if we have a regression somewhere due to some coming incompatibilities it wouldn't show up in unrelated PRs. Like this recent pytest failure.

PR #3615, opened 2026-06-16
MAST: Resolve objects more performantly

snbianco:
Small fix to resolve named objects with our MAST resolver service more performantly.

### DIFF
--- a/astroquery/mast/tests/test_mast_remote.py
+++ b/astroquery/mast/tests/test_mast_remote.py
@@ -1,24 +1,36 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-import logging
-from pathlib import Path
-import numpy as np
-import os
-import pytest
 import json
+import logging
+import os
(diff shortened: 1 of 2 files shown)


astroquery/mast/tests/test_mast_remote.py
```diff
@@ -1,24 +1,36 @@
...
+    Hapcut,
+    Mast,
+    MastMissions,
+    Observations,
+    Tesscut,
+    Zcut,
+    utils,
+)
+
+from ...exceptions import (
```
bsipocz: A bit of bikeshedding, but I wonder if you could fine tune your linter's setting for astroquery and not do these reformats for shorter lines.

bsipocz (approved):
This all looks good, thanks!

Temporarily commenting out NED sounds totally fine, but maybe you want to open up an issue as a reminder so it's more likely to get a revisit in the future than a TODO left in the test file itself.

PR #3619, opened 2026-06-20
hips2fits: make Pillow an optional dependency with a clear ImportError

kushasahu7:
## Summary

Fixes the remaining open item in #3556.

`PIL` was imported unconditionally at module level in `hips2fits/core.py`, causing a hard `ModuleNotFoundError` on `import astroquery.hips2fits` when Pillow is not installed — even though it is only needed when decoding jpg/png responses.

## Changes

- Moved `from PIL import Image` inside `_parse_result()` behind a `try/except ImportError` with a helpful install message
- Added `Pillow` to the `all` optional extras in `pyproject.toml`
- Added two unit tests covering jpg and png formats when Pillow is absent

## Related

- Closes #3556
- Follows the same pattern used for `boto3`/`botocore` in #3534


kushasahu7:
@bsipocz  When you have some time, can you review this PR ?

ManonMarchand (approved):
LGTM, thank you

astroquery/hips2fits/tests/test_hips2fits.py
```diff
@@ -7,6 +11,27 @@
...
 
 
+class TestHips2fits:
+
+    def test_jpg_format_requires_pillow(self):
+        mock_response = MagicMock()
+        mock_response.status_code = 200
+        mock_response.content = b'fake_image_bytes'
+
+        with patch.dict(sys.modules, {'PIL': None, 'PIL.Image': None}):
```
bsipocz: I don't think this is the correct approach, use pytest markers instead to check for the missing pillow install.

astroquery/hips2fits/tests/test_hips2fits.py
```diff
@@ -7,6 +11,27 @@
...
+            with pytest.raises(ImportError, match="Pillow is required"):
+                hips2fits._parse_result(mock_response, verbose=False, format='jpg')
+
+    def test_png_format_requires_pillow(self):
```
bsipocz: These two tests are practically identical, please use pytest parametrize

CHANGES.rst
```diff
@@ -103,6 +103,13 @@ svo_fps
 - Add ``get_filter_metadata`` to allow retrieval of filter metadata. [#3528]
 - Add ``get_zeropoint`` to allow retrieval of filter zeropoints and allow kwarg passing to ``get_filter_metadata``. [#3545]
 
+hips2fits
+^^^^^^^^^
+
+- ``Pillow`` is now an optional dependency: importing ``astroquery.hips2fits`` no
+  longer crashes when Pillow is not installed. A clear ``ImportError`` is raised
+  only when a jpg/png response is actually decoded. [#3619]
```
bsipocz: This needs a rewrite, pillow has been an optional dependency already.
```suggestion
- Fix the import error for not available optional dependency Pillow. A clear ``ImportError`` is raised
  only when a jpg/png response is actually decoded. [#3619]
```

bsipocz (commented):
this needs some cleanup, but the big picture looks good.

thanks

kushasahu7:
Thanks for the review @bsipocz! Addressed all three:
- Dropped the `sys.modules` patching in favour of a `HAS_PILLOW` + `@pytest.mark.skipif` marker.
- Merged the jpg/png tests into a single `@pytest.mark.parametrize`.
- Reworded the changelog per your suggestion.

PR #3622, opened 2026-06-26
MAST: Add Gaia DR3 and make it default

pllim:
Someone really wanted Gaia DR3 access on MAST. Without this, the astroquery API is inconsistent with web interface.


pllim:
`E       KeyError: 'Mast.Catalogs.GaiaDR3.Cone'`

I guess that is incorrect key but what is the correct one?

tomdonaldson:
> `E KeyError: 'Mast.Catalogs.GaiaDR3.Cone'`
> 
> I guess that is incorrect key but what is the correct one?

That is the correct service name.  Just a quick look and I'm wondering if this just needs a new entry in the map that sets up the mock results for that test here: https://github.com/astropy/astroquery/blob/59105119c2c0abe67744ea3198f958abc225adda/astroquery/mast/tests/test_mast.py#L52

pllim:
RTD failure is unrelated.

@tomdonaldson , thanks! I would have never figured that out.

pllim:
The person who needed this confirmed that it works. This is ready for final review. Thanks!

keflavich (approved):
Docs build failure is unrelated - looks like a configuration issue.

pllim:
Merged. Thanks, all!

PR #3625, opened 2026-07-02
Use the global session in heasarc queries

### LINKED ISSUE #3624: Session not set correctly in the heasarc module
heasarc/core.py overrides the global `_session` from `query.py` to use pyVO's session. It leads to 'astroquery' not showing up in the request's 'User-Agent'. Having 'astroquery` there is useful to tracking.

zoghbi-a:
Fixes #3624

### DIFF
--- a/astroquery/heasarc/core.py
+++ b/astroquery/heasarc/core.py
@@ -38,7 +38,6 @@ class HeasarcClass(BaseVOQuery, BaseQuery):
         self._tap = None
         self._datalink = None
         self._meta_info = None
-        self._session = None
 
     @property
     def tap(self):


zoghbi-a:
This is a trivial change, and changelog is added. I am not sure why it is failing.

CHANGES.rst
```diff
@@ -29,6 +29,11 @@ esa.emds.einsteinprobe
 API changes
 -----------
 
+heasarc
+^^^^^^^^^^
+
+- Ensure 'astroquery' is in the request headers. [#3439]
```
bsipocz: it's always the PR number that goes into the changelog and not references to issues.

That being said, I don't think this change needs a changelog, as `_session` is private and doesn't really affect the end user.

bsipocz (approved):
I'll go back and remove the changelog commits to make the CI happy and the merge. Thanks!

PR #3630, opened 2026-07-15
Fix two bugs in calling heasarc.query_region

zoghbi-a:
This is fixes two issues:

- `Heasarc.query_region(position, catalog=None)`: This used to fail but the recent addition of `_query_execute` has pushed the failure down. This fix make the failure happen earlier.

- Passing `add_offset=True` to `Heasarc.query_region` fails when `columns=None`. This was also caused by the recent addition of `_query_execute`

Tests were added for both cases.

Other minor changes:
- remove the `test_tap__maxrec` test. Recent changes in pyvo make it irrelevant.
- Style fixes using flake8

### DIFF
--- a/astroquery/heasarc/core.py
+++ b/astroquery/heasarc/core.py
@@ -38,6 +38,11 @@ class HeasarcClass(BaseVOQuery, BaseQuery):
         self._tap = None
         self._datalink = None
         self._meta_info = None
+        self._catalog_msg = (
+            "catalog name is required! "
+            "Use 'xray' to search the master X-ray catalog. Or list catalogs by"
+            "calling :meth:`~astroquery.heasarc.HeasarcClass.list_catalogs`"
+        )
 
     @property
     def tap(self):
@@ -326,8 +331,7 @@ class HeasarcClass(BaseVOQuery, BaseQuery):
             commons.suppress_vo_warnings()
 
         if catalog is None:
-            raise InvalidQueryError("catalog name is required! Use 'xray' "
-                                    "to search the master X-ray catalog")
+            raise InvalidQueryError(self._catalog_msg)
 
         if where is None:
             where = ''
@@ -501,6 +505,9 @@ class HeasarcClass(BaseVOQuery, BaseQuery):
         # if we have column_filters and no position, assume all-sky search
         if position is None and column_filters is not None:
             spatial = 'all-sky'
+        
+        if catalog is None:
+            raise InvalidQueryError(self._catalog_msg)
 
         if spatial.lower() == 'all-sky':
             where = ''
--- a/astroquery/heasarc/core.py
+++ b/astroquery/heasarc/core.py
@@ -296,6 +296,7 @@ class HeasarcClass(BaseVOQuery, BaseQuery):
             query, language='ADQL', maxrec=maxrec, uploads=uploads)
 
     def _query_execute(self, catalog=None, where=None, *,
+                       offset_column=None,
                        get_query_payload=False, columns=None,
                        verbose=False, maxrec=None):
         """Queries some catalog using the HEASARC TAP server based on the
@@ -309,6 +310,9 @@ class HeasarcClass(BaseVOQuery, BaseQuery):
         where : str
             The WHERE condition to be used in the query. It must
             include the 'WHERE' keyword or be empty.
+        offset_column: str or None
+            If add_offset is True in query_regoni, this contains the
+            the string that addes that to the query.
         get_query_payload : bool, optional
             If `True` then returns the generated ADQL query as str.
             Defaults to `False`.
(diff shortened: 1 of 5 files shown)


keflavich (changes requested):
Could you please filter out the unrelated whitespace-only changes?  The diff is full of non-functional changes that makes it difficult to review.

zoghbi-a:
@keflavich, the style fixes were typing to make my linter happy but that clearly complicated the review. I removed them and kept only those that CI complained about.

astroquery/heasarc/core.py
```diff
@@ -304,6 +310,9 @@ def _query_execute(self, catalog=None, where=None, *,
         where : str
             The WHERE condition to be used in the query. It must
             include the 'WHERE' keyword or be empty.
+        offset_column: str or None
+            If add_offset is True in query_regoni, this contains the
```
keflavich:
```suggestion
            If add_offset is True in query_region, this contains the
```

keflavich (approved):
just a minor doc typo to fix, otherwise lgtm

astroquery/heasarc/tests/test_heasarc_remote.py
```diff
@@ -280,6 +268,18 @@ def test_row_count(self):
         cat = "suzamaster"
         assert Heasarc.count_rows(cat) == 3055
 
+    def test_query_region_offset_with_no_column(self):
+        # use columns='*' to avoid remote call to obtain the default columns
+        _ = Heasarc.query_region(
```
bsipocz: can we have an assert here? The comment somewhat contradicts the call below as the default is being used for columns, but then no checks are done to see if additional columns are present in the result.

(as a rule of thumb, assertless tests  are fine in rare occasion, but we really should aim for doing some actual content testing as we already did the expensive part by running the query)

astroquery/heasarc/tests/test_heasarc_remote.py
```diff
@@ -156,13 +151,6 @@ def test_list_catalogs__keywords(self):
         assert "rosmaster" in catalogs
         assert "rassmaster" in catalogs
 
-    def test_tap__maxrec(self):
```
bsipocz, line 159: why has been an test removed? The comment above indicates that we can revise this once we have pyvo 1.7 as minimum; which is not yet the case and we do see see warnings still.

astroquery/heasarc/core.py
```diff
@@ -502,6 +513,17 @@ def query_region(self, position=None, catalog=None, radius=None, *,
...
 
+        # check for a valid catalog name
+        if catalog is None:
+            raise InvalidQueryError(self._catalog_msg)
+
+        # add_offset is valid only with cone searches
+        if spatial != 'cone' and add_offset:
+            raise InvalidQueryError("add_offset is valid only spatial=='cone'")
+
+        # to hold the offset columns, if needed
```
bsipocz:
```suggestion
        # to hold the offset column, if needed
```

astroquery/heasarc/core.py
```diff
@@ -502,6 +513,17 @@ def query_region(self, position=None, catalog=None, radius=None, *,
         if position is None and column_filters is not None:
             spatial = 'all-sky'
 
+        # check for a valid catalog name
+        if catalog is None:
+            raise InvalidQueryError(self._catalog_msg)
+
+        # add_offset is valid only with cone searches
+        if spatial != 'cone' and add_offset:
```
bsipocz: terrible nitpick, but I would switch up the code logic to follow the intention/comments.
```suggestion
        if add_offset and spatial != 'cone':
```

bsipocz (approved):
Overall looks good but I have a few, mostly nitpicky comments. And some squashing would be nice so we have less back and forth with changelog and style commits.

Thanks.

bsipocz:
(fwiw, I liked your cleanup commit, but agree with Adam that it's better suited for a separate PR rather than mixed in with a bug fixes)

astroquery/heasarc/tests/test_heasarc_remote.py
zoghbi-a: I removed it because the tests were failing when I have 1.7. What is the recommended way to deal with this?

zoghbi-a: I added it back with mark to check for pyvo version.

zoghbi-a:
Thanks @bsipocz. I think those are addressed. I squashed some of the changes and rebased. Feel free to do a full squash when you merge.

bsipocz (approved):
I see some remote tests failing, as well as some of the doctest are failing, but I see those failured on `main`, too. A simple commit should fix the tests, so I'll just go ahead and push it directly.

The rest of the diff looks good; Thanks!

Issue #3626, opened 2026-07-05
Vizier incorrectly treats some errors as empty results


orionlee:
For Vizier, there are cases that the Vizier server returns error, but 
1. `astroquery` returns a misleading empty result (empty `TableList`). 
Expected behavior: it should raise an exception, informing the caller the vizier server error.

2. the issue is further exasperated by that the results of failed queries are not purged from cache #2252 . So even after the server is available , one would still get the misleading empty results, and has to manually delete the cache (`.pickle` file in astroquery cache dir) to rectify it.

---

**Example:***
```python
import astroquery.vizier
from astropy.coordinates import SkyCoord, Angle

table_list = astroquery.vizier.Vizier().query_region(SkyCoord(295.91256250430, 63.11031720565, frame="icrs", unit="deg"), catalog="I/355/paramp", radius=Angle(15, "arcsec"))
print(table_list)
```

1. Normally it returns 1 table with 2 rows:
```
TableList with 1 tables:
	'0:I/355/paramp' with 24 column(s) and 2 row(s) 
```

2. when the Vizier server is unstable with database connection error, the code would return a misleading empty `TableList`. 
Here is an example actual votable response returned from Vizier (the response has an HTTP 200 status code): 

```xml
<ns0:VOTABLE xmlns:ns0="http://www.ivoa.net/xml/VOTable/v1.3" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.4" xsi:schemaLocation="http://www.ivoa.net/xml/VOTable/v1.3 http://www.ivoa.net/xml/VOTable/v1.3">
  <ns0:DESCRIPTION>
   VizieR Astronomical Server vizier.cds.unistra.fr
    Date: 2026-07-05T21:24:43 [V7.5.6]
   Explanations and Statistics of UCDs:			See LINK below
   In case of problem, please report to:	cds-question@unistra.fr
 </ns0:DESCRIPTION>
  <ns0:INFO ID="Error" name="Error" value="The database is not currently reachable." />
  <ns0:INFO name="Error" value=" " />
  <ns0:INFO name="Error" value=" -- no connection" />
  <ns0:INFO name="Error" value=" Report from: stdb_pg,     Status: STDB_ERR " />
  <ns0:INFO name="Error" value="Postgres connect error" />
  <ns0:INFO name="Error" value=" " />
  <ns0:INFO name="Error" value=" -- no connection" />
[...]

PR #3632, opened 2026-07-17
[VizieR] Raise error in case of server-side issues

ManonMarchand:
Hello astroquery maintainers :wave: 

This PR intends to solve #3626 

It raises an error when a VizieR response contains an ``info`` element ``QUERY_STATUS`` with value ``ERROR`` and removes the corresponding entry from the cache.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -215,6 +215,13 @@ simbad
 - Add the possibility to declare more information in the HTTP User-Agent header
   in ``SimbadClass`` [#3529]
 
+vizier
+~~~~~~
+
+- When server-side errors occur (ex: VizieR is overloaded), the response is not kept in
+  astroquery's cache anymore, and a Python error is raised instead of returning a
+  misleading empty ``TableList`` [#3632]
+
 xmatch
 ^^^^^^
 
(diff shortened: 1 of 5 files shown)


ManonMarchand:
Aaarg ``get_infos_by_name`` was introduced in astropy 5.3. I'll fix this on monday.

keflavich:
I'm not sure we should be supporting versions older than that - it was added 3 years ago.  @bsipocz what would you say to bumping min-astropy-version?

bsipocz:
> I'm not sure we should be supporting versions older than that - it was added 3 years ago. @bsipocz what would you say to bumping min-astropy-version?

Let me see if how quickly we can ship the next release and if it would make sense to keep the versions as is for that release.

 But in general, I agree that we should bump versions.

ManonMarchand:
No need to bump versions for this PR, I'll loop through the ``infos`` and write it in a separate commit so that this workaround will be easy to remove when the support for versions older than astropy 5.3 will be dropped.

bsipocz (approved):
Thanks and thanks for being flexible about the astropy version! I'm sure we'll bump requirements soon, but it's also good to be not blocked by it.

PR #3633, opened 2026-07-17
Added explicit Fornax datalinks and download support to the Heasarc submodule

DavidT3:
# Changes

- A new constructed column, "fornax", is now included in the datalinks table returned by Heasarc.locate_data. This is almost identical in function to the "sciserver" column currently included in the return.
- Updated the Heasarc.download_data method, as well as several internal methods called by it, to support 'fornax' as an input to the `host` argument. This means that data can be copied from the Fornax mount of the HEASARC S3 bucket.
- Updated the _guess_host method to be able to detect when the module is running on Fornax.
- Altered the tests to account for the new 'fornax' host.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -110,6 +110,7 @@ heasarc
 - Add support for uploading tables when using TAP directly through ``query_tap``. [#3403]
 - Add automatic guessing for the data host in ``download_data``. [#3403]
 - Include method to count the number of rows in a specified table. [#3549]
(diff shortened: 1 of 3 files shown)


DavidT3:
Actually, we might not continue with the mounted S3 buckets on Fornax - converted to a draft while I wait to see what we might do instead.

DavidT3:
Nevermind, feature no longer needed, closing the PR.

PR #3636, opened 2026-07-21
Remove unquote method from completed UWS job XML

### LINKED ISSUE #3637: CASDA stage_data providing incomplete URL for artefacts
The CASDA SODA API is returning a URL that is incompatible with the current astroquery.casda library. A recent release to CASDA caused the URLs to not be completely escaped, which was previously handled by an unquote method called after the data had been retrieved. The newest version of CASDA altered this and now provides users with an URL, that when used in conjunction with urllib, fails to parse and causes an error.

Below are examples of the old vs new behavior when links are returned through CASDA,

old 
`http%3A%2F%2Fcasda.csiro.au%2Fdownload%2Fweb%2F111-000-111-000%2Faskap_img.fits`
new
`http://casda.csiro.au/download/web/111-000-111-000/askap_img.fits`

Issue arises when a AWS pre-signed URL is returned from the service which will contain illegal characters.

Below is an example of replication, provided by a user

```
import pyvo as vo
from astroquery.casda import Casda
import urllib

session=vo.tap.TAPService('https://casda.csiro.au/casda_vo_tools/tap')

sbid='85679'
query=f"""SELECT * FROM ivoa.obscore
    WHERE ( obs_collection = 'EMU' OR obs_collection = 'WALLABY' )
    AND (obs_id = 'ASKAP-{sbid}')
    AND ( quality_level = 'GOOD' OR quality_level = 'UNCERTAIN' )
    AND filename LIKE '%components.xml'  """
result=session.search(query).to_table()
print('CASDA Obscode returns:')
print(result)
print()

casda_session=Casda()
casda_session.login(username="[OPAL account email address redacted]")



url_list=casda_session.stage_data(result,verbose=True)
print('CASDA stage_data returns:')
print(url_list)
print()

for url in url_list:
    if url.endswith('checksum'):
        continue

#    url=url.replace('"','').replace(' ','') #Strip out illegal URL characters
    with urllib.request.urlopen(url) as r: #Crashes here, malformed URL or Forbidden
        filename = r.info().get_filename()
        filepath = f"{output}/{filename}"
        http_size = int(r.info()['Content-Length'])

```

Removing the `unquote` method from the resulting links produces a workable URL.

mitchcs:
Hi Astroquery,

Recent changes to the underlying CASDA uws XML results in this unquote method not longer requiring being called. Tested locally against CASDA and resolves issues users are seeing

resolves #3637

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,7 +25,6 @@ esa.emds.einsteinprobe
 - New module to access the ESA Einstein Probe Science Archive. [#3511]
 
 
-
 API changes
 -----------
 
@@ -146,6 +145,13 @@ ipac.irsa
   were typos; the API expects ``naifid``. Old ``obj_nafid`` keyword and
   ``"nafid_input"`` ``input_mode`` still work but emit
   ``AstropyDeprecationWarning``. [#3607]
+
+casda
+^^^^^
+
+- Preserve the percent-encoding of staged file URLs so that pre-signed S3
+  download URLs remain valid. Previously the URLs were unquoted, which corrupted
+  pre-signed URLs and could raise errors when parsed by ``urllib``. [#3636]
 
 
 mast
--- a/astroquery/casda/core.py
+++ b/astroquery/casda/core.py
@@ -280,7 +280,7 @@ class CasdaClass(QueryWithLogin):
         job_details = self._get_job_details_xml(job_url)
         fileurls = []
         for result in job_details.find("uws:results", self._uws_ns).findall("uws:result", self._uws_ns):
-            file_location = unquote(result.get("{http://www.w3.org/1999/xlink}href"))
+            file_location = result.get("{http://www.w3.org/1999/xlink}href")
             fileurls.append(file_location)
 
         return fileurls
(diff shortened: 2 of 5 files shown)


bsipocz:
@mitchcs - why did you close this? Could you open an issue with the problems users are seeing?

mitchcs:
@bsipocz I wasn't sure at the time if this was the completely correct fix for the issue. I have updated the branch and will reopen with the additional changes.

The issue is that CASDA altered the resulting URL from these requests and was historically using unquote to unescape, as shown in the example. The newest version of CASDA instead provides the functional formatted URL.

CHANGES.rst
```diff
@@ -147,6 +146,13 @@ ipac.irsa
   ``"nafid_input"`` ``input_mode`` still work but emit
   ``AstropyDeprecationWarning``. [#3607]
 
+casda
+^^^^^
+
+- Preserve the percent-encoding of staged file URLs so that pre-signed S3
+  download URLs remain valid. Previously the URLs were unquoted, which corrupted
+  pre-signed URLs and could raise errors when parsed by ``urllib``. [#3637]
```
```suggestion
  pre-signed URLs and could raise errors when parsed by ``urllib``. [#3636]
```

bsipocz (approved):
Looks good with the exception of the changelog -- we reference the PR and not the issue.
Thanks!

bsipocz:
(I went ahead and rebased, too, so we have clear branch history without unnecessary merge commits)

PR #3643, opened 2026-08-06
Add SOFIA and SPHEREx to MOST docs

jonesmg:
I added SOFIA and SPHEREx to the MOST docs to bring it up to date with https://irsa.ipac.caltech.edu/applications/MOST/instructions.html

### DIFF
--- a/astroquery/ipac/irsa/most.py
+++ b/astroquery/ipac/irsa/most.py
@@ -286,12 +286,12 @@ class MostClass(BaseQuery):
             Epoch in MJD.
             Required when input mode is ``"manual_input"``.
         eccentricity : float or None
-            Eccentricity (0-1).
+            Eccentricity (>=0).
             Required when input mode is ``"manual_input"``.
         inclination : float or None
             Inclination (0-180 degrees).
             Required when input mode is ``"manual_input"``.
-        arg_perihelion : str or None
+        arg_perihelion : float or None
             Argument of perihelion (0-360 degrees).
             Required when input mode is ``"manual_input"``.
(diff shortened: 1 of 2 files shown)


astroquery/ipac/irsa/most.py
bsipocz, line 313: idea for future improvement: this could/should also be an astropy.time.Time

docs/ipac/irsa/most.rst
```diff
@@ -27,7 +27,7 @@ JPL's Horizon database whenever possible. This lookup can fail when name
 ambiguities exist. For example, entering "Neptune" is unclear because ephemeris
 for both Neptune and Neptune system barycenter exist.
 
-Most will deliver a messsage if there is an ambiguity. The message may contain
+Most will deliver a message if there is an ambiguity. The message may contain
```
bsipocz, line 30: 😄

bsipocz (approved):
The diff looks great, thanks!

The base of this branch is rather outdated and that is the reason for the RTD failure. (It's actually not clear why RTD not treat this as a pull request but just builds the top of the branch, maybe @pllim has an idea?)

As for the GHA failures, that's all looking unrelated, there is red github status for actions atm.

pllim:
> It's actually not clear why RTD not treat this as a pull request but just builds the top of the branch

I think that's just how RTD configured the backend; probably cheaper too since it is not doing one extra step of rebasing. I cannot find any info to change that behavior on our end but feel free to ask at https://github.com/readthedocs/readthedocs.org/issues

PR #3645, opened 2026-08-14
MAST: `count_only` parameter for `MastMissions` queries

snbianco:
Quick PR to add a `count_only` parameter to `MastMissions` query methods. If this parameter is set, the method will return a count of results rather than the full result table. Also updated docs and tests.


bsipocz (approved):
This looks all good, thanks!

Also, heads up, there are quite a lot of remote test failures for the module, but they seem to be unrelated to this PR so I go ahead with the merge.

PR #3659, opened 2026-09-10
Las Cumbres Observatory (LCO) Archive Access

jnation3406:
This is an updated module for [Las Cumbres Observatory](https://lco.global) (LCO)'s [Archive](https://archive.lco.global). I tried to follow the template and existing modules examples for everything, but please let me know if there are any API changes that need to be made.

One question I have is about the use of `async`. I decorated the class with `@async_to_sync` as shown in the template example, but after reading #2598, I am wondering if I should be leaving that out completely? The LCO Archive API is entirely synchronous in its request/response cycle - it does not support asynchronous jobs since its just querying and returning data.

There are unit tests, remote tests, and documentation for the module - please let me know if anything else is needed.


docs/lco/lco.rst
```diff
@@ -0,0 +1,311 @@
...
+    lsc0m409-kb98-20240302-0017-b00     lsc         0m4a               BIAS
+    lsc0m409-kb98-20240302-0016-b00     lsc         0m4a               BIAS
+    lsc0m409-kb98-20240302-0015-b00     lsc         0m4a               BIAS
+
+The same criteria can be passed to ``query_region`` and ``query_object``, which
+narrow a positional or name search rather than replacing it.
+
+``start`` and ``end`` bound the observation date and accept a string, an
+`~astropy.time.Time`, or a `~datetime.datetime`. A query should **always** be
+narrowed by ``start`` and ``end`` to reduce load on the database.
```
keflavich: Because of this concern, I'd recommend making these mandatory instead of keyword arguments - we don't want astroquery accidentally taking down the db

keflavich: or alternatively, warn users if they're performing scary big queries

keflavich: keep in mind that many 'users' may end up being AIs, so a stern warning may be enough to stop them if they're well-aligned

docs/lco/lco.rst
```diff
@@ -0,0 +1,311 @@
...
+
+``reduction_level`` is ``0`` for raw data and ``91`` for reduced.
```
keflavich: are these the only two values?  Why 91?  (are there docs to link to about this?)

docs/lco/lco.rst
```diff
@@ -0,0 +1,311 @@
...
+holds more than were returned, a ``MaxResultsWarning`` is printed.
+
+.. doctest-remote-data::
+
+    >>> frames = LcoArchive.query_criteria(site_id="lsc", row_limit=3,
+    ...                                    start="2024-03-02", end="2024-03-03")   # doctest: +SHOW_WARNINGS
+    MaxResultsWarning: Results truncated to 3 frames. Pass row_limit=-1 to retrieve every matching frame.
+
+Pass ``row_limit=-1`` to retrieve every match, paging through the archive as
+needed.
```
keflavich, line 185: what's the return if the results are paged?  Will it be 500 at a time, say?  How does the user page through - or will the query tool handle it?

docs/lco/lco.rst
```diff
@@ -0,0 +1,311 @@
...
+
+    >>> from astroquery.lco import LcoArchive
+    >>> LcoArchive.login(token="your-api-token")
+    >>> LcoArchive.authenticated()
+    True
+
+You can log in with your observing portal username instead, which exchanges
+your credentials for a token. The password is read from your system keyring if
+it is stored there, and prompted for otherwise; pass ``store_password=True`` to
+save it for later sessions:
```
keflavich: tell the user how it is stored - is the python `keyring` package used?

docs/lco/lco.rst
```diff
@@ -0,0 +1,311 @@
...
+
+
+Configuration
+=============
+
+The module reads four configuration items, which can be changed at runtime through
+``astroquery.lco.conf``:
+
+``archive_url``
+    The archive to query. Defaults to ``https://archive-api.lco.global``.
```
keflavich: I'm not sure what the formatting is going to look like here - we should inspect the rendered docs and make sure it looks OK

keflavich (commented):
This looks like a great addition, thanks!  I asked some questions in the docs.

docs/lco/lco.rst
jnation3406: The full results are returned to the user, up to the `row_limit` defined (or the default of 100). Internally, if the number of results is greater than the page size of 100 for unauthenticated or 1000 for authenticated, the module will make multiple queries until the full results are retrieved and return those.

docs/lco/lco.rst
jnation3406: I'll change the default values that are links to be links in the docs rather than a code block / string.

docs/lco/lco.rst
jnation3406: So maybe I over-emphasized things there. It's not a requirement, and it won't bring our server's down on its own. I was just trying to emphasize that it should be a first line of narrowing down broad queries. It's equally good to narrow down queries in other ways, such as with a specific request id or proposal id. I just wanted to try to encourage users (and Bots!) to narrow their date ranges when possible, to get results more quickly. I guess I'll re-write this to try to better capture that intent.

docs/lco/lco.rst
jnation3406: I'll link to the docs that describe the valid values

docs/lco/lco.rst
jnation3406: Thinking about it some more, I'm going to add some warnings and progress statements logged while retrieving data from a large query. I don't want a user to accidentally request multiple GB of data and then run out of memory without warning.

jnation3406:
I've added a `show_progress` param to queries defaulting to True, which displays a progress bar or spinner with the count of rows retrieved from the Archive during a query when run in a tty. This is to help inform users to the progress of paging through the archive to get their results, and also let them know how much longer it is expected to take on a bound query. 

For unbounded queries (`row_limit=-1`), a warning will be printed when the rows retrieved exceeds 10,000: this is to alert the user that their query may not be as constrained as they think, so they have a chance to kill it early.

I also think I've addressed your initial review comments. I see some failing checks in the PR -  I don't think the failing CI-devtest is related to my module but should I be updating the Changelog as part of this PR?. Please let me know if there is anything else I need to do to get the module accepted.

astroquery/lco/core.py
```diff
@@ -0,0 +1,790 @@
...
+        whose footprint overlaps the requested region.
+
+        Parameters
+        ----------
+        coordinates : str or `~astropy.coordinates.SkyCoord`
+            The single position to search around.
+        radius : str or `~astropy.units.Quantity`, optional
+            Radius of a cone search. The circle is sent to the archive as a
+            polygon with 32 vertices, so the match is approximate at the
+            sub-arcsecond level.
```
keflavich: if you specify a big enough circle, it's approximate at worse than sub-arcsecond, right?

astroquery/lco/core.py
```diff
@@ -0,0 +1,790 @@
...
+            :meth:`~astroquery.lco.LcoArchiveQuery.list_criteria` returns every
+            accepted name, and the ``list_*`` methods give the accepted values
+            for the code-valued ones. The criteria below are the ones that
+            take a value you cannot guess from the name:
+
+            ``start``, ``end``
+                Bounds on the observation date. Accepts a string, an
+                `~astropy.time.Time`, or a `~datetime.datetime`.
+            ``covers``, ``intersects``
+                A WKT region the frame footprint must contain, or overlap.
```
keflavich: what's WKT?

keflavich (approved):
@bsipocz this looks good to me.  I skimmed through focusing mostly on documentation, but all the pieces are in place and this is one of the most thorough matches to our template documents we've gotten. 

I'd recommend we do one more pass for obvious things, but then merge and allow for small corrections in subsequent PRs.

@jnation3406 thanks for an excellent new module!

astroquery/lco/core.py
jnation3406: [Well Known Text](https://libgeos.org/specifications/wkt/) is a specification for describing geometry in plain text strings. Like "Point (RA DEC)" for a point or "Polygon ((RA1 DEC1), (RA2 DEC2), etc...)". I'll add a link to that description in the doc.

keflavich: thanks!  TIL

keflavich: probably link to that description the first time it comes up in the docs

astroquery/lco/core.py
jnation3406: Correct, I'll fix the comment to more accurately describe the error in approximating a circle as a polygon (the error is <=0.5% of the radius).

astroquery/lco/__init__.py
```diff
@@ -0,0 +1,42 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# Generative AI was used in the making of this module
+
+"""
+Las Cumbres Observatory (LCO)
+-----------------------------
+
+:author: Jon Nation (jnation@lco.global)
```
bsipocz, line 8: We don't consistently cleaned stuff up, but we don't track authorship in code files, the git history should suffice.

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+from astropy.utils.console import ProgressBarOrSpinner, Spinner
+
+from astroquery import cache_conf, log
+
+from ..exceptions import (LargeQueryWarning, LoginError, MaxResultsWarning,
+                          NoResultsWarning, RemoteServiceError)
+from ..query import QueryWithLogin
+from ..utils import async_to_sync, commons
+from . import conf
+
```
bsipocz, line 30: do absolute imports. Again we haven't cleaned up everything, but ultimately we should aim for this.

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+
+__all__ = ['LcoArchive', 'LcoArchiveQuery']
```
bsipocz, line 32: This naming doesn't really follow the conventions we have, any reasons for the diversion?

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+
+
+#: Filters accepted by the ``/frames/`` endpoint. Anything not in here is
+#: rejected client-side so that typos do not silently return the whole archive.
+FRAME_FILTERS = (
```
bsipocz, line 37: Is there a way to programmatically get this from the serverside? How can we ensure that this stays up-to-date?

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+                          NoResultsWarning, RemoteServiceError)
+from ..query import QueryWithLogin
+from ..utils import async_to_sync, commons
+from . import conf
+
+
+__all__ = ['LcoArchive', 'LcoArchiveQuery']
+
+
+#: Filters accepted by the ``/frames/`` endpoint. Anything not in here is
```
bsipocz, line 35: super minor, but could you remove the `:`? The convention is to use a single hashmark for comments followed by one space

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+#: Thumbnail sizes the archive publishes.
+THUMBNAIL_SIZES = ('small', 'large')
+
+#: LCO Archive presigned URLs are only valid for this long, so only
+#: cache responses for this long.
+CACHE_TIMEOUT = 48 * 60 * 60
+
+#: How many frames an unbounded query may collect before it warns that it is
+#: still going.
+LARGE_RESULT_WARNING = 10000
```
bsipocz, line 102: I wonder if these could/should be in the config instead to have some exposure to the users, or only being hardwired in the code?

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+    proprietary period has expired. Logging in with an API token additionally
+    exposes the proprietary frames belonging to your proposals.
+    """
+
+    ARCHIVE_URL = conf.archive_url
+    TOKEN_URL = conf.token_url
+    TIMEOUT = conf.timeout
+    ROW_LIMIT = conf.row_limit
+
+    _thumbnail_size = 'small'
```
bsipocz, line 120: could/should this be configurable?

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+#: LCO Archive presigned URLs are only valid for this long, so only
+#: cache responses for this long.
+CACHE_TIMEOUT = 48 * 60 * 60
+
+#: How many frames an unbounded query may collect before it warns that it is
+#: still going.
+LARGE_RESULT_WARNING = 10000
+
+
+@async_to_sync
```
bsipocz, line 105: I would say we don't need this any more, we're moving away from having duplicated `method` and `method_async` approach. We can do proper async when the server supports it, but then we tend to switch to that via a kwarg instead of duplicating methods.

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+        log.info(f"Authenticated to the LCO archive as "
+                 f"{profile.json()['username']}.")
+        return True
+
+    def logout(self):
+        """Drop the API token and return the session to anonymous access."""
+        self._session.headers.pop('Authorization', None)
+        self._authenticated = False
+
+    def query_region_async(self, coordinates, *, radius=None, width=None,
```
bsipocz, line 214: have only `query_region` etc, drop the async part and have Tables as the return objects.

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+                           height=None, show_progress=True,
+                           get_query_payload=False, cache=None, **criteria):
```
bsipocz, line 216: What can be included in ``**criteria``? If that is a finite and smallish number of options, then I would very strongly support of adding them directly rather than as a kwarg that practically swallows everything.

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+            Any other frame filter, as accepted by
+            :meth:`~astroquery.lco.LcoArchiveQuery.query_criteria`.
+
+        Returns
+        -------
+        frames : list of dict
+            The frame records the archive returned, one dict per frame.
+        """
+        name_filter = 'target_name_exact' if exact else 'target_name'
+        return self.query_criteria_async(**{name_filter: object_name},
```
bsipocz, line 298: why do it this way with `**`?

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+    def query_criteria_async(self, *, get_query_payload=False, cache=None,
+                             row_limit=None, include_thumbnails=False,
+                             thumbnail_size='small', show_progress=True,
+                             **criteria):
+        """
+        Query the archive on any combination of frame filters.
+
+        Parameters
+        ----------
+        **criteria
```
bsipocz, line 313: I wonder if we infact want to expose all of the possible criteria as kwargs? Are they expected to change? 
That being said, there are 25 of them right now (returned by list_criteria) so it's really borderline which approach is the more user friendly

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+        row_limit = self.ROW_LIMIT if row_limit is None else row_limit
+        if row_limit != -1 and row_limit < 1:
+            # A page of no rows tells a cursor query nothing, so there is no
+            # sensible request to make for it.
+            raise ValueError("'row_limit' must be 1 or more, or -1 to "
+                             f"retrieve every matching frame; got {row_limit}.")
+
+        # Stashed for _parse_result, because async_to_sync only returns the
+        # response and 'verbose'.
+        self._thumbnail_size = thumbnail_size
```
bsipocz, line 386: I think this can be removed from the class once we remove the `async_to_sync` deadend

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+        cache : bool, optional
+            Cache the response. Defaults to `True` for anonymous queries and
+            `False` for authenticated ones.
+
+        Returns
+        -------
+        frames : list of dict
+            The frame records the archive returned, one dict per frame.
+        """
+        if thumbnail_size not in THUMBNAIL_SIZES:
```
bsipocz, line 374: if you have `include_thumbnails` on False, then why does this matter?

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+        payload = self._args_to_payload(**criteria)
+        if include_thumbnails:
+            payload['include_thumbnails'] = 'true'
+        if get_query_payload:
+            return payload
+
+        return self._fetch_pages(payload, row_limit=row_limit, cache=cache,
+                                 show_progress=show_progress)
+
+    def _region_to_criteria(self, coordinates, *, radius=None, width=None,
```
bsipocz, line 397: Why does this have to be factored out? I would prefer it more if it was part of the method given it's really a small number of LOC

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+                                if t.get('size') == self._thumbnail_size), None)
+                          for row in rows]
+            table['thumbnail_url'] = [t['url'] if t else '' for t in thumbnails]
+            table['thumbnail_filename'] = [
+                f"{t['basename']}{t['extension']}" if t else ''
+                for t in thumbnails]
+
+        return table
+
+    def get_frame(self, frame_id, *, cache=None):
```
bsipocz, line 575: I suppose we will prefer to call this `get_image` or `get_images` as we have that for the other modules, we don't call it "frames" anywhere else.

astroquery/lco/core.py
```diff
@@ -0,0 +1,792 @@
...
+    def list_configuration_types(self):
+        """List the ``configuration_type`` values present in the archive."""
+        return self._aggregate('obstypes')
+
+    def list_proposals(self):
+        """List the proposal ids present in the archive."""
+        return self._aggregate('proposals')
+
+
+class _ProgressBarOrCountingSpinner(ProgressBarOrSpinner):
```
bsipocz, line 739: we do we need this subclass?

bsipocz (commented):
I have a bunch of comments, mostly about API consistency with the rest of astroquery. 

Also, as this is a new module, I would recommend of not doing any of the async_to_sync machinery. And that being said, I suppose we need to cleanup the docs, too, in a separate PR.

astroquery/lco/__init__.py
jnation3406: happy to remove it, just FYI it's in your template module too.

astroquery/lco/core.py
jnation3406: No good reason, I'll switch it to `LcoArchiveClass`

astroquery/lco/core.py
jnation3406: Unfortunately we don't currently have an endpoint that returns the full set of filter options for frame queries. These change very rarely - basically never being removed, only occasionally adding new filter options. I will try to keep this module updated as we do add new options, but the functionality defined in the module should continue to work regardless.

astroquery/lco/core.py
jnation3406: This subclass overrides the behavior of the default Spinner provided by astropy. The default only spins indefinitely for a query with limit=-1 (unbounded). This adds a count next to the spinner of the amount of images currently retrieved (since we batch up retrievals per call of 100 or 1000). That way the user can see there is progress being made on their query, and approximately what rate data is coming back at.

astroquery/lco/core.py
jnation3406: So this method just retrieves metadata, including a fresh download url, for a given id from our archive database. I don't see any precedence for what it's doing in other modules, and it's only really meant to be an internal method (called by `download_files` in the case that the frames table provided has expired image urls, i.e. a few days passed since the query was made before they tried to download the files). If I rename it to `get_metadata` would that satisfy your concerns?

jnation3406: or I could name it `_get_metadata` so users are less likely to try to use it themselves and get confused?

jnation3406:
@bsipocz Thank you for your review comments. I think I've addressed them all in my last commit, but please let me know if there is anything else you would like me to change.

